PR #1, opened 2025-09-30
Add v0.1.0

cpubot:
#### Problem
See https://github.com/anza-xyz/agave/pull/8242

After writing https://github.com/anza-xyz/agave/pull/8141, it's clear that writing manual serialization to overcome performance limitations of `serde` and various solana sdk types is a frequent problem. In particular, Safe Rust offers limited support for placement initialization (direct writes without intermediate staging buffers), especially for heap memory. The effect of this is a lot of unnecessary copying, which is unacceptable in performance‑critical code; made significantly worse by the fact we frequently serialize `Vec`s of fixed sized arrays (`Signature`, `Address`, etc). `serde` inherits these limitations since it neither attempts to initialize in‑place nor exposes APIs to do so (especially in heap memory --again). Worse still, many of these types are serialized with `solana_short_vec`, which does per-element visitation on byte arrays and byte vectors.

The result of this is a proliferation of manual serialization / deserialization implementations across various types to circumvent these limitations. These manual implementations require a lot of time and effort to get right.

#### Summary of Changes

This builds upon those manual implementations, and the patterns observed while writing https://github.com/anza-xyz/agave/pull/8141, and packages that functionality into a its own crate -- `wincode`.

`wincode` is a bincode‑compatible serializer/deserializer focused on in‑place initialization and direct memory writes. The goal is to make it easy to make performant and efficient serializer / deserializers that avoid all the aforementioned pitfalls.
In short, `wincode` operates over traits that facilitate direct writes of memory into final destinations (including heap-allocated buffers) without intermediate staging buffers. It also supports different length encodings, to support `solana-short-vec`, for example.

The simple case looks like the following:
```rs
struct MyStruct {
    data: Vec<u64>,
    win: bool,
}

compound! {
    MyStruct {
        data: Vec<u64>,
        win: bool,
    }
}

let val = MyStruct { data: vec![1,2,3], win: true };
assert_eq!(wincode::serialize(&val).unwrap(), bincode::serialize(&val).unwrap());
```

And here is case more typical of agave:
```rs
struct MyStruct {
    data: Vec<u8>,
    addresses: Vec<[u8; 32]>,
}

compound! {
    MyStruct {
        data: containers::Vec<Pod<u8>>,
        addresses: containers::Vec<Pod<[u8; 32]>>,
    }
}

let val = MyStruct { data: vec![1,2,3], addresses: vec![[0; 32], [1; 32]] };
assert_eq!(wincode::serialize(&val).unwrap(), bincode::serialize(&val).unwrap());
```

In this example,
- All initialization is done in-place, including heap-allocated memory (true of all supported heap-allocated structures in wincode).
- Byte fields are written in a single pass by leveraging the `Pod` `container`.
- No intermediate staging buffers: bytes are copied directly into the final destination allocation during deserialization and written directly into the output buffer during serialization.

APIs like `wincode::serialize` and `wincode::deserialize` should feel nearly identical to what users expect of `bincode::serialize` and `bincode::deserialize`, though there is richer functionality available for more exotic cases -- code is fully documented.

Here is what the `Entry` implementation looks like (replaces https://github.com/anza-xyz/agave/pull/8141)
```rs
compound! {
    struct MessageHeader => solana_message::MessageHeader {
        num_required_signatures: u8,
        num_readonly_signed_accounts: u8,
        num_readonly_unsigned_accounts: u8,
    }
}

compound! {
    struct CompiledInstruction => solana_transaction::CompiledInstruction {
        program_id_index: u8,
        accounts: containers::Vec<Pod<u8>, ShortU16Len>,
        data: containers::Vec<Pod<u8>, ShortU16Len>,
    }
}

compound! {
    struct LegacyMessage => legacy::Message {
        header: MessageHeader,
        account_keys: containers::Vec<Pod<Address>, ShortU16Len>,
        recent_blockhash: Pod<Hash>,
        instructions: containers::Vec<Elem<CompiledInstruction>, ShortU16Len>,
    }
}

compound! {
    struct MessageAddressTableLookup => v0::MessageAddressTableLookup {
        account_key: Pod<Address>,
        writable_indexes: containers::Vec<Pod<u8>, ShortU16Len>,
        readonly_indexes: containers::Vec<Pod<u8>, ShortU16Len>,
    }
}

compound! {
    struct V0Message => v0::Message {
        header: MessageHeader,
        account_keys: containers::Vec<Pod<Address>, ShortU16Len>,
        recent_blockhash: Pod<Hash>,
        instructions: containers::Vec<Elem<CompiledInstruction>, ShortU16Len>,
        address_table_lookups: containers::Vec<Elem<MessageAddressTableLookup>, ShortU16Len>,
    }
}

compound! {
    struct VersionedTx => versioned::VersionedTransaction {
        signatures: containers::Vec<Pod<Signature>, ShortU16Len>,
        message: VersionedMsg,
    }
}

compound! {
    Entry {
        num_hashes: u64,
        hash: Pod<Hash>,
        transactions: Vec<VersionedTx>,
    }
}
```


wincode/src/len.rs
```diff
@@ -0,0 +1,225 @@
...
+/// against malicious input causing OOM. The default is 4 MiB. Users are
+/// free to override this limit by passing a different constant or by
+/// implementing their own `SeqLen` implementation.
+pub struct BincodeLen<const MAX_SIZE: usize = DEFAULT_BINCODE_LEN_MAX_SIZE>;
+
+impl<const MAX_SIZE: usize> SeqLen for BincodeLen<MAX_SIZE> {
+    /// Bincode's default fixint encoding writes lengths as `u64`.
+    #[inline(always)]
+    fn size_hint(reader: &mut Reader) -> Result<usize> {
+        u64::get(reader).map(|len| len as usize)
```
t-nelson: `usize::try_from`?

wincode/src/io.rs
```diff
@@ -0,0 +1,206 @@
...
+            // SAFETY:
+            // - `src` mustn't overlap with the internal buffer (shouldn't be the case unless the user is doing something they shouldn't).
+            // - We just checked that we have enough capacity in the internal buffer.
+            ptr::copy_nonoverlapping(
+                src.as_ptr(),
+                self.buffer.as_mut_ptr().add(self.pos).cast(),
+                src.len(),
+            );
+        }
+        self.pos += src.len();
```
t-nelson: is one of the explicit arithmetic methods appropriate here?

wincode/src/io.rs
```diff
@@ -0,0 +1,206 @@
...
+    pub unsafe fn write_with<F>(&mut self, len: usize, write: F) -> Result<()>
+    where
+        F: FnOnce(&mut [MaybeUninit<u8>]) -> Result<()>,
+    {
+        let dst = self
+            .buffer
+            .get_mut(self.pos..self.pos + len)
+            .ok_or_else(|| write_size_limit(len))?;
+        write(dst)?;
+        self.pos += len;
```
t-nelson: similarly here, explicit arithmetic?

wincode/src/error.rs
```diff
@@ -0,0 +1,74 @@
...
+    Error::WriteSizeLimit(len)
+}
+
+#[cold]
+pub fn preallocation_size_limit(needed: usize, limit: usize) -> Error {
+    Error::PreallocationSizeLimit { needed, limit }
+}
+
+#[cold]
+pub fn size_hint_overflow(max_length: &'static str) -> Error {
```
t-nelson: `const`?

wincode/src/len.rs
```diff
@@ -0,0 +1,225 @@
...
+    }
+
+    #[inline(always)]
+    fn bytes_needed(_len: usize) -> Result<usize> {
+        Ok(size_of::<u64>())
+    }
+}
+
+#[cfg(feature = "solana-short-vec")]
+pub mod short_vec {
```
t-nelson, line 69: one day we will kill this abomination

t-nelson, line 69: any wins for explicit arithmetic in this module?

wincode/src/schema/impls.rs
```diff
@@ -0,0 +1,794 @@
...
+            type Src = $type;
+
+            #[inline(always)]
+            fn size_of(_src: &Self::Src) -> Result<usize> {
+                Ok(size_of::<$type>())
+            }
+
+            #[inline(always)]
+            fn write(writer: &mut Writer, src: &Self::Src) -> Result<()> {
+                #[cfg(target_endian = "little")]
```
t-nelson, line 59: does `to_le_bytes()` not perform this optimization for us?

wincode/src/schema/impls.rs
```diff
@@ -0,0 +1,794 @@
...
+                Ok(())
+            }
+        }
+
+        impl SchemaRead<'_> for $type {
+            type Dst = $type;
+
+            #[inline(always)]
+            fn read(reader: &mut Reader, dst: &mut MaybeUninit<Self::Dst>) -> Result<()> {
+                #[cfg(target_endian = "little")]
```
t-nelson, line 79: ditto

wincode/src/schema/mod.rs
```diff
@@ -0,0 +1,794 @@
...
+};
+
+pub mod containers;
+mod impls;
+
+/// Types that can be written (serialized) to a byte buffer.
+pub trait SchemaWrite {
+    type Src: ?Sized;
+    /// Get the serialized size of `Self::Src`.
+    fn size_of(src: &Self::Src) -> Result<usize>;
```
t-nelson: bunch of these impls have implicit arithmetic, maybe should be explicit?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,39 @@
+name: CI
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+
+jobs:
```
t-nelson, line 8: can we hookup coverage as well?

cpubot: Sure, which action(s) do we typically use for this?

wincode/src/error.rs
cpubot: Yeah, technically no reason they couldn't all be `const`.

wincode/src/len.rs
cpubot: Good call

.github/workflows/ci.yml
t-nelson: ah i guess it's still run from buildkite on monorepo... @yihau do we have any opinions on a coverage action for new repos?

wincode/src/error.rs
t-nelson: right. this was the only one i noticed with const usage today. no harm in hosing them all down for consistency

wincode/src/io.rs
cpubot: I don't think it's necessary here. The above capacity check enforces `self.pos + src.len() <= self.buffer.len()`

wincode/src/io.rs
cpubot: Here, yeah I think it makes sense on the `get_mut(self.pos..self.pos + len)` -- `self.pos + len` should use `checked_add`

wincode/src/schema/impls.rs
cpubot: It technically does turn `to_le_bytes` into a noop, but I'm unsure if the compiler is smart enough to avoid the copy given `&src.to_le_bytes()` (the copy is trivial, but made my OCD flare up)

wincode/src/schema/impls.rs
cpubot: Similar to the above, I don't know if the compiler is smart enough to optimize this

- On LE, the we do a single memcpy directly into dst
- On BE, we necessarily do load + byteswap + store

BE path introduces an extra load+store via a temporary (unless the compiler eliminates it ); the LE path guarantees a single copy into dst

Again the amount of bytes we're dealing with here is trivial, but it felt reasonable to ensure fewer instructions given the small amount of code it entails (I'd probably feel differently if it was a lot of code)

wincode/src/len.rs
cpubot: Everything in here is either checked or guaranteed not to overflow, unless I've overlooked something

wincode/src/schema/mod.rs
cpubot: Yeah -- probably worth it for some of the variable sized implementations (heap allocd types like `Vec`)

wincode/src/schema/containers.rs
```diff
@@ -0,0 +1,602 @@
+//! This module provides specialized "container" types that can be used to opt
```
alessandrod, line 1: nit: this should probably be `collections` to match `std::collections`?

wincode/src/lib.rs
```diff
@@ -0,0 +1,277 @@
...
+//!     pub address: [u8; 32],
+//! }
+//! ```
+//!
+//! `serde`'s default, naive, implementation will perform per-element visitation of all bytes
+//! in `Vec<u8>` and `[u8; 32]`. Because these fields are "plain old data", ideally we would
+//! avoid per-element visitation entirely and read / write these fields in a single pass.
+//! The situation worsens if this struct needs to be written into a heap allocated data structure,
+//! like a `Vec<A>` or `Box<[A]>`. Due to lack of placement initialization of heap memory, all
+//! those bytes will be put on the stack before being copied into the heap allocation.
```
alessandrod, line 84: I think that this is true for `Box<A>` but not for `Vec<A>`? The latter doesn't
have a `new(the elements)` constructor, but it's built with `from_iter` and friends

.github/workflows/ci.yml
yihau: maybe we can steal the idea from the sdk repo haha
https://github.com/anza-xyz/solana-sdk/blob/5c389a1dc95b4a267aae44f4e43bf32386aab027/.github/workflows/main.yml#L499-L520

however, it requires some environment setup. we can add that later so it wouldn't block the important part!

wincode/src/schema/containers.rs
```diff
@@ -0,0 +1,596 @@
...
+///
+/// let my_struct = MyStruct {
+///     address: Address(array::from_fn(|i| i as u8)),
+/// };
+/// let wincode_bytes = wincode::serialize(&my_struct).unwrap();
+/// let bincode_bytes = bincode::serialize(&my_struct).unwrap();
+/// assert_eq!(wincode_bytes, bincode_bytes);
+/// # }
+/// ```
+pub struct Pod<T>(PhantomData<T>);
```
alessandrod: is there a reason for this and Elem not to be marker traits? If they were we
could implement them for all builtins and then people would implement them for
their types and `#[wincode(with)]` would be a useful but fringe API? 

Actually does Elem even need to exist? Couldn't we have a marker trait for Pod
and then things that aren't Pod do the slower per-item thing? Am I missing something?

wincode-derive/src/schema_read.rs
```diff
@@ -0,0 +1,329 @@
...
+    let dst = get_src_dst(args);
+    let impl_generics = append_de_lifetime(&args.generics);
+    let (_, ty_generics, where_clause) = args.generics.split_for_impl();
+
+    let helpers = fields.iter().enumerate().map(|(i, field)| {
+        let ty = override_ref_lifetime(&field.ty);
+        let target = override_ref_lifetime(field.target());
+        let ident = field.struct_member_ident(i);
+        let ident_string = field.struct_member_ident_to_string(i);
+        let get_uninit_mut_ident = format_ident!("get_uninit_{}_mut", ident_string);
```
alessandrod: nit: I'd drop the get_ prefix

(thing()/set_thing() is idiomatic rust)

wincode/src/schema/containers.rs
```diff
@@ -0,0 +1,596 @@
...
+    type Dst = T;
+
+    fn read(reader: &mut Reader, dst: &mut MaybeUninit<Self::Dst>) -> Result<()> {
+        // SAFETY: `T` is plain ol' data.
+        unsafe { reader.read_t(dst) }
+    }
+}
+
+#[cfg(feature = "alloc")]
+impl<T, Len> SchemaWrite for Vec<Elem<T>, Len>
```
alessandrod, line 221: if Pod was a trait, we also wouldn't need a separate schema impl for the alloc
containers I think?

.github/workflows/ci.yml
t-nelson: ok yeah let's punt to a follow up. would be good to get coverage enabled and ensure everything is well tested before 1.0.0

wincode/src/io.rs
t-nelson: given the extent to which this crate is applicable and that it will be acting on user controlled inputs, it might pay to deny the `arithmetic_side_effects` clippy lint and just be explicit everywhere. we've done this selectively (though unlikely with good selection criteria) throughout the monorepo

wincode/src/schema/impls.rs
t-nelson: makes sense! just caught my eye as a little high up in the stack to be making this kind of decision. codegen paranoia seems reasonable tho

wincode/src/schema/containers.rs
```diff
@@ -315,89 +323,118 @@ where
...
-    }
-}
-
-#[cfg(feature = "alloc")]
-impl<T, Len> SchemaRead<'_> for BoxedSlice<Pod<T>, Len>
-where
-    Len: SeqLen,
-{
-    type Dst = alloc::boxed::Box<[T]>;
+macro_rules! impl_heap_slice {
```
t-nelson: try to do these refactors in standalone commits in the future. easier to review

wincode/src/schema/containers.rs
cpubot: > is there a reason for this and Elem not to be marker traits? 

It's a good question, and I did consider this. Two reasons:

----

Firstly, two different trait bounds can't have separate implementations `∀ T`. Meaning, the following would be rejected by the compiler:

```rs
trait Pod {}
trait Elem {}

impl<T> SchemaWrite for Vec<T> where T: Pod {}
impl<T> SchemaWrite for Vec<T> where T: Elem {}
```
A ZST allows this exact differentiation between implementations

Or perhaps you were thinking that specifically the `containers::Vec` would be gated by `T: Pod`. `containers::Vec` exists for more than just the `Pod` specialization -- it's also what allows for a specialized length encoding. In the solana sdk, for example, we have non-POD `Vec`s that use the short-u16 length encoding. In this case, we need to opt into `containers::Vec` implementation, but cannot use `Pod`, which entails `containers::Vec<Elem<T>, ShortU16Len>`.

> #[wincode(with)] would be a useful but fringe API?

It still is for the most part. Users _only_ need to reach for `#[wincode(with)]` if:
1. the data is POD,
2. or they need a length encoding other than bincode's default length encoding.

We still have a blanket implementation for `alloc::vec::Vec<T>`:
https://github.com/anza-xyz/wincode/blob/127e8d576cbf2a877f15373327c6f112e8753e16/wincode/src/schema/impls.rs#L266-L296

Blanket implementations for all collection types follow this pattern.

> Couldn't we have a marker trait for Pod and then things that aren't Pod do the slower per-item thing?

That's exactly the behavior the blanket implementations provide.

In summary, 
- if user simply annotates `Vec<T>`, we use the per-item implementation, which doesn't make any assumptions about the POD-nature of `T` and uses bincode's default length encoding (`containers::Vec<Elem<T> BincodeLen>`).
- If user needs a custom length encoding or to opt into the optimized POD Vec implementation, they use `#[wincode(with)]`

----

Additionally, `Pod` as a ZST allows annotating foreign types as POD. For example, `Signature` in `solana-signature` is:
```rs
#[repr(transparent)]
struct Signature([u8; 64)]
```

This allows:
```rs
struct MyStruct {
    #[wincode(with = "Pod<_>")]
    signature: Signature,
}
```

without requiring the solana sdk to `impl Pod for Signature`

wincode/src/schema/containers.rs
cpubot: We would if we want to be able to specialize `Vec`'s (or other contiguous containers) impl to read and write all bytes in a single pass like:
https://github.com/anza-xyz/wincode/blob/127e8d576cbf2a877f15373327c6f112e8753e16/wincode/src/schema/containers.rs#L311-L312

If `Pod` is a trait, we can have an efficient direct byte read / write for `Pod` _elements_, but can't have differentiated implementations for `Vec<T>` that allow reading _all_ elements in a single pass, due to the overlapping implementations entailed by
```rs
trait Pod {}
trait Elem {}

impl<T> SchemaWrite for Vec<T> where T: Pod {}
impl<T> SchemaWrite for Vec<T> where T: Elem {}
```

But let me know if you have something else in mind that I'm missing

wincode/src/lib.rs
cpubot: `Vec<A>` in serde is built by `SeqAccess::next_element` with a subsequent call to `Vec::push`, which almost certainly has the same problem [[₁]](https://github.com/serde-rs/serde/blob/master/serde_core/src/de/impls.rs#L1057-L1060) [[₂]](https://hackmd.io/XXuVXH46T8StJB_y0urnYg#The-code-ordering-problem). 

It's not clear if `from_iter` is able to write directly to heap memory either without some temporary, given that the iterator must yield a `T` before it can be `ptr::write`n [[₃]](https://github.com/rust-lang/rust/blob/4a54b26d30dac43778afb0e503524b763fce0eee/library/alloc/src/vec/mod.rs#L3785-L3807) -- the destination pointer is not available to the constructor of `T` (unless there's some special optimization implemented in the compiler somewhere that I'm not aware of)

wincode/src/schema/containers.rs
cpubot: I thought about this too, but we have more than collections in here, like `Pod`. Though, `Elem` is specific to these specific collections, whereas `Pod` can be used on its own. I don't feel strongly. Happy to change it if you prefer.

wincode/src/lib.rs
alessandrod: Duh I'm dumb I wasn't thinking about deserialization at all 😂

wincode/src/error.rs
```diff
@@ -0,0 +1,95 @@
+//! Error types and helpers.
+use {core::str::Utf8Error, thiserror::Error};
+
+#[derive(Error, Debug)]
+pub enum Error {
```
alessandrod, line 5: this is the first commit in a 0.1.0 crate, so we don't have to fix this now, just mentioning for future reference: something I learned the hard way workingon aya is that in libraries you almost never want this sort of catch all errors.Say that I'm a mission critical app - say like something that manages user funds - and I want to exhaustively handle all errors; say I call Writer::write_exact() then I have to handle ReadSizeLimit, which makes no sense.

wincode/src/len.rs
```diff
@@ -0,0 +1,226 @@
...
+};
+
+/// Behavior to support heterogenous sequence length encoding.
+///
+/// It is possible for sequences to have different length encoding schemes.
+/// This trait abstracts over that possibility, allowing users to specify
+/// the length encoding scheme for a sequence.
+pub trait SeqLen {
+    /// Read the length of a sequence from the reader.
+    fn size_hint(reader: &mut Reader) -> Result<usize>;
```
alessandrod: nit: I found it a bit confusing that this is called size hint.

If you don't return the actual exact size here you get UB.

wincode/src/len.rs
```diff
@@ -0,0 +1,226 @@
...
+    /// Get the length of a sequence from the reader, potentially
+    /// returning an error if some length condition is not met
+    /// (e.g., size constraints, overflow, etc.).
+    #[inline(always)]
+    fn size_hint_cautious<T>(reader: &mut Reader) -> Result<usize> {
```
alessandrod: We only ever call size_hint_cautious, is there a reason for exposing two trait
methods?

Couldn't we just expose one and document that it's allowed to return the limit
error? (which I'd probably call SizeLimit and drop the Preallocation part).

Looking at the BincodeLen impl I sort of understand why it's structured the way
it is, but I'm not sure anyone would ever call size_hint() without wanting to
cap the result.

wincode/src/len.rs
```diff
@@ -0,0 +1,226 @@
...
+        Self::size_hint(reader)
+    }
+    /// Write the length of a sequence to the writer.
+    fn encode_len(writer: &mut Writer, len: usize) -> Result<()>;
```
alessandrod: nit: I'd call this write

...and this made me realize that I'd probably also rename size_hint => read 😂

wincode/src/io.rs
alessandrod: ack, in the vm where applicable we've added allow lints + explanation of why overflow isn't possible to perf sensitive subsections of code

wincode/src/schema/containers.rs
alessandrod: Sounds like Pod should go somewhere else? 😋

I don't feel strongly either up to you!

wincode/src/schema/containers.rs
alessandrod: discussed on discord - I'm dumb and assumed that marker + blanket _was_ supported on stable, which it isn't.

alexpyattaev:
Just FYI adding this as a dep into agave crate results in version conflicts:
```
Updating git repository `https://github.com/anza-xyz/wincode.git`
    Updating crates.io index
error: failed to select a version for `paste`.
    ... required by package `ark-ff v0.4.2`
    ... which satisfies dependency `ark-ff = "^0.4.0"` (locked to 0.4.2) of package `solana-bn254 v3.0.0`
    ... which satisfies dependency `solana-bn254 = "^3.0.0"` (locked to 3.0.0) of package `agave-syscalls v3.1.0 (/home/sol/agave/syscalls)`
    ... which satisfies path dependency `agave-syscalls` (locked to 3.1.0) of package `agave-ledger-tool v3.1.0 (/home/sol/agave/ledger-tool)`
versions that meet the requirements `^1.0` (locked to 1.0.9) are: 1.0.9

all possible versions conflict with previously selected packages.

  previously selected package `paste v1.0.15`
    ... which satisfies dependency `paste = "^1.0.15"` of package `wincode v0.1.0 (https://github.com/anza-xyz/wincode.git?branch=initial#5373282f)`
    ... which satisfies git dependency `wincode` of package `solana-tls-utils v3.1.0 (/home/sol/agave/tls-utils)`
    ... which satisfies path dependency `solana-tls-utils` (locked to 3.1.0) of package `solana-core v3.1.0 (/home/sol/agave/core)`
    ... which satisfies path dependency `solana-core` (locked to 3.1.0) of package `agave-ledger-tool v3.1.0 (/home/sol/agave/ledger-tool)`

failed to select a version for `paste` which could resolve this conflict
```
suggest change paste version to 1.0.9 (or 1.0.*)

wincode/src/io.rs
```diff
@@ -0,0 +1,217 @@
...
+    #[inline]
+    pub fn peek(&self) -> Result<&u8> {
+        self.cursor.first().ok_or_else(|| read_size_limit(0))
+    }
+
+    /// Copy exactly `dst.len()` bytes from the [`Reader`] into `dst`.
+    #[inline]
+    pub fn read_exact(&mut self, dst: &mut [MaybeUninit<u8>]) -> Result<()> {
+        let Some((src, rest)) = self.cursor.split_at_checked(dst.len()) else {
+            return Err(read_size_limit(dst.len()));
```
alessandrod, line 40: this doesn't seem to have coverage

wincode/src/io.rs
```diff
@@ -0,0 +1,217 @@
...
+        };
+        self.read_exact(slice)?;
+        Ok(())
+    }
+
+    /// Read exactly `len` bytes from the cursor into a slice.
+    #[inline]
+    pub fn read_borrowed(&mut self, len: usize) -> Result<&'a [u8]> {
+        let Some((src, rest)) = self.cursor.split_at_checked(len) else {
+            return Err(read_size_limit(len));
```
alessandrod, line 80: same here I think no coverage

wincode/src/io.rs
```diff
@@ -0,0 +1,217 @@
...
+    #[inline(always)]
+    pub fn consume_unchecked(&mut self, amt: usize) {
+        self.cursor = &self.cursor[amt..];
+    }
+
+    /// Advance `amt` bytes from the reader and discard them.
+    #[inline]
+    pub fn consume(&mut self, amt: usize) -> Result<()> {
+        if self.cursor.len() < amt {
+            return Err(read_size_limit(amt));
```
alessandrod, line 113: this guy too

wincode/src/io.rs
```diff
@@ -0,0 +1,217 @@
...
+    pos: usize,
+}
+
+impl<'a> Writer<'a> {
+    pub fn new(buffer: &'a mut [MaybeUninit<u8>]) -> Self {
+        Self { buffer, pos: 0 }
+    }
+
+    #[cfg(feature = "alloc")]
+    pub fn from_vec(buffer: &'a mut Vec<u8>) -> Self {
```
alessandrod: unused and undocumented? 🤔

wincode/src/io.rs
```diff
@@ -0,0 +1,217 @@
...
+    }
+
+    /// Get the number of bytes written to the buffer.
+    #[inline]
+    pub fn finish(self) -> usize {
+        self.pos
+    }
+
+    /// Get the number of bytes written to the buffer, and error if there are trailing bytes.
+    pub fn finish_disallow_trailing_bytes(self) -> Result<usize> {
```
alessandrod: missing tests methinks

wincode/src/io.rs
```diff
@@ -0,0 +1,217 @@
...
+            ));
+        }
+        Ok(self.pos)
+    }
+
+    /// Write exactly `src.len()` bytes from the given `src` into the internal buffer.
+    #[inline]
+    pub fn write_exact(&mut self, src: &[u8]) -> Result<()> {
+        if self.buffer.len().saturating_sub(self.pos) < src.len() {
+            return Err(write_size_limit(src.len()));
```
alessandrod, line 172: no coverage?

wincode/src/io.rs
```diff
@@ -0,0 +1,217 @@
...
+    ///
+    /// - `write` must write EXACTLY `len` bytes into the given buffer.
+    pub unsafe fn write_with<F>(&mut self, len: usize, write: F) -> Result<()>
+    where
+        F: FnOnce(&mut [MaybeUninit<u8>]) -> Result<()>,
+    {
+        let upper_bound = self
+            .pos
+            .checked_add(len)
+            .ok_or_else(|| buffer_length_overflow(len))?;
```
alessandrod: no coverage for the else path

wincode/src/io.rs
```diff
@@ -0,0 +1,217 @@
...
+        let dst = self
+            .buffer
+            .get_mut(self.pos..upper_bound)
+            .ok_or_else(|| write_size_limit(len))?;
```
alessandrod, line 212: no coverage for the else path

wincode/src/schema/impls.rs
```diff
@@ -0,0 +1,970 @@
...
+{
+    type Dst = VecDeque<T::Dst>;
+
+    #[inline]
+    fn read(reader: &mut Reader<'de>, dst: &mut MaybeUninit<Self::Dst>) -> Result<()> {
+        <containers::VecDeque<Elem<T>, BincodeLen>>::read(reader, dst)
+    }
+}
+
+impl<T> SchemaWrite for [T]
```
alessandrod, line 378: this impl doesn't seem to be tested

wincode/src/schema/impls.rs
```diff
@@ -0,0 +1,970 @@
...
+        {
+            type Dst = $container<T::Dst>;
+
+            #[inline]
+            fn read(reader: &mut Reader<'de>, dst: &mut MaybeUninit<Self::Dst>) -> Result<()> {
+                struct DropGuard<T>(*mut MaybeUninit<T>);
+                impl<T> Drop for DropGuard<T> {
+                    #[inline]
+                    fn drop(&mut self) {
+                        drop(unsafe { $container::from_raw(self.0) });
```
alessandrod, line 555: this doesn't seem to be tested

alessandrod (commented):
some things that jumped out while looking at coverage (there are a few more uncovered things, but I pointed out only the ones that could potentially break with malformed inputs)

wincode/src/schema/containers.rs
```diff
@@ -0,0 +1,649 @@
...
+//! Vector with non-POD elements and custom length encoding:
+//!
+//! ```
+//! # #[cfg(all(feature = "solana-short-vec", feature = "alloc", feature = "derive"))] {
+//! # use wincode_derive::SchemaWrite;
+//! # use wincode::{containers::{self, Elem}, len::ShortU16Len};
+//! # use serde::Serialize;
+//! # use solana_short_vec;
+//! #[derive(Serialize, SchemaWrite)]
+//! struct Point {
```
alessandrod, line 60: this is a bit confusing because this type _is_ pod?

alessandrod:
we should hook up miri since we do a lot of unsafe stuff

wincode/src/schema/containers.rs
```diff
@@ -0,0 +1,649 @@
...
+    type Src = T;
+
+    #[inline]
+    fn size_of(_src: &Self::Src) -> Result<usize> {
+        Ok(size_of::<T>())
+    }
+
+    #[inline]
+    fn write(writer: &mut Writer, src: &Self::Src) -> Result<()> {
+        // SAFETY: `T` is plain ol' data.
```
alessandrod, line 203: this is trust me bro tho right? 😂 not sure if there's a way to make this less
of a footgun. I think in aya we define Pod as Copy + 'static? Maybe add those
bounds here?

wincode/src/lib.rs
```diff
@@ -0,0 +1,277 @@
...
+pub mod len;
+mod schema;
+pub use schema::*;
+mod serde;
+pub use serde::*;
+
+#[doc(hidden)]
+pub mod __private {
+    #[doc(hidden)]
+    pub use paste::paste;
```
alessandrod: what is this for?

wincode/src/schema/impls.rs
```diff
@@ -0,0 +1,970 @@
...
+            fn read(reader: &mut Reader, dst: &mut MaybeUninit<Self::Dst>) -> Result<()> {
+                // SAFETY: `$type` is plain ol' data.
+                unsafe { reader.read_t(dst) }
+            }
+        }
+    };
+}
+
+impl_pod!(u8);
+impl_pod!(i8);
```
alessandrod, line 180: why are the other primitives not pod?

cavemanloverboy: for integers u have to do endianess checks, impl_int! covers these

wincode/src/schema/containers.rs
cpubot: Haha yeah definitely a "trust me bro". Good idea though -- I'll add `Copy + 'static`

wincode/src/schema/containers.rs
cpubot: It's POD in the sense that it's Copy + 'static, but has to consider endianness and isn't #[repr(C)] (no layout guarantee)

wincode/src/error.rs
cpubot: Good call -- I'll split the errors up

wincode/src/io.rs
cpubot: ack

wincode/src/io.rs
cpubot: ack

wincode/src/io.rs
cpubot: ack

wincode/src/io.rs
cpubot: Yeah... I think I used it at one point in `Serialize`. Will remove

wincode/src/io.rs
cpubot: ack

wincode/src/io.rs
cpubot: ack

wincode/src/io.rs
cpubot: ack

wincode/src/io.rs
cpubot: ack

wincode/src/len.rs
cpubot: Yeah good call. This came out of looking at `serde`'s implementation for initial implementation, but you're right that we only ever call `size_hint_cautious`.

wincode/src/len.rs
cpubot: Works for me 🙂

wincode/src/lib.rs
cpubot: Leftover from `compound!` (before the derive macro) -- will remove

wincode/src/proptest_config.rs
```diff
@@ -0,0 +1,20 @@
...
+/// Configuration for proptest tests.
+///
+/// Disable FS I/O in Miri.
+pub(crate) fn proptest_cfg() -> Config {
+    #[cfg(miri)]
+    {
+        Config {
+            failure_persistence: None,
+            // Avoid excruciatingly long test times in Miri.
+            cases: 5,
```
alessandrod, line 12: would it make sense to lower this only on cfg(miri)?

wincode/src/schema/containers.rs
alessandrod: This is why this PR broke my brain a bit btw: I think we should clarify that
when we use POD we mean POD+, that is POD + some other requirements.

This falls under we don't have to fix in 0.1.0.

wincode/src/schema/impls.rs
```diff
@@ -0,0 +1,978 @@
...
+    type Dst = [T::Dst; N];
+
+    #[inline]
+    fn read(reader: &mut Reader<'de>, dst: &mut MaybeUninit<Self::Dst>) -> ReadResult<()> {
+        // SAFETY: MaybeUninit<[T::Dst; N]> trivially converts to [MaybeUninit<T::Dst>; N].
+        let dst =
+            unsafe { transmute::<&mut MaybeUninit<Self::Dst>, &mut [MaybeUninit<T::Dst>; N]>(dst) };
+        let mut base = dst.as_mut_ptr();
+        let mut guard = SliceDropGuard::<T::Dst>::new(base);
+        // Avoid slice reborrow of `dst` (triggered by `dst.iter_mut()`)
```
alessandrod, line 409: nit: do we not have the clippy "useless for" lint enabled? (we could bind to i
and do base.add(i))

wincode/src/proptest_config.rs
cpubot: Should be gated by the `#[cfg(miri)]` above

alessandrod: lmao this is what happens if you do reviews at 7am

wincode/src/schema/impls.rs
cpubot: Done

PR #6, opened 2025-10-10
v0.1.1

cpubot:
- Fix docsrs build for `wincode` #5
- Make #[derive(SchemaRead)] less brittle with generic lifetimes #4


Cargo.toml
```diff
@@ -12,7 +12,6 @@ arithmetic_side_effects = "deny"
 warnings = "deny"
 
 [workspace.package]
-version = "0.1.0"
```
cpubot, line 15: No reason to lock this to the workspace

PR #7, opened 2025-10-10
Implement specialization for contiguous byte structures

cpubot:
Users must currently opt into direct buffer reads / writes with `Pod`, and `#[wincode(with)]` makes that easy enough. But, it would be nice if the library just "did the right thing" when it can tell that the src / dst type is `u8`.

This steals some logic from the [`unty`](https://crates.io/crates/unty) crate, but force inlines it so the compiler can reliably DCE the branch (it does not reliably do so when importing the crate). I have verified that indeed the compiler is able to eliminate the branch: https://rust.godbolt.org/z/1WMY8qn8K, which means the types that use this (basically all contiguous data structures (vec, arrays, etc) shouldn't have any runtime overhead with this check.

E.g., the following type should opt into these optimizations

```rs
struct MyStruct {
    bytes: Vec<u8>,
    signature: [u8; 64]
}
```

### DIFF
--- a/wincode/src/schema/impls.rs
+++ b/wincode/src/schema/impls.rs
@@ -15,9 +15,30 @@ use std::{
     collections::{HashMap, HashSet},
     hash::Hash,
 };
+use {
+    crate::{
+        containers::SliceDropGuard,
+        error::{
+            invalid_bool_encoding, invalid_char_lead, invalid_tag_encoding, invalid_utf8_encoding,
+            pointer_sized_decode_error, ReadResult, WriteResult,
+        },
+        io::{Reader, Writer},
+        len::{BincodeLen, SeqLen},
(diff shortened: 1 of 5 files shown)


wincode/src/schema/impls.rs
```diff
@@ -15,34 +15,39 @@ use std::{
...
-        schema::{size_of_elem_iter, write_elem_iter, SchemaRead, SchemaWrite},
+        schema::{size_of_elem_slice, write_elem_slice, SchemaRead, SchemaWrite},
+        util::type_equal,
     },
     core::{
         marker::PhantomData,
         mem::{self, transmute, MaybeUninit},
     },
 };
+#[cfg(feature = "alloc")]
```
cpubot, line 35: Not sure why rustfmt decided to move this

PR #15, opened 2025-10-16
add custom error variants to `ReadError` and `WriteError`

### LINKED ISSUE #14: `ReadError` and `WriteError` should support custom errors
Currently, there seems to be no support for throwing custom error types. This is useful for any type that needs to do arbitrary verification during (de)serialization, e.g. a signature type that needs to check whether the point is on the elliptic curve.

There should probably be a way to throw custom `ReadError`s and `WriteError`s, e.g. by using `map_err()` on some error type within the implementations of `SchemaRead::read()` and `SchemaWrite::write()`. This is analogous to this use case in `serde`: https://serde.rs/convert-error.html

qkniep:
Adds variants `ReadError::Custom` and `WriteError::Custom` for turning a `&'static str` into a variant of the corresponding `wincode` error type.

These can for example be used in `map_err()` as follows:
```rust
impl<'de> SchemaRead<'de> for IndividualSignature {
    type Dst = IndividualSignature;

    fn read(
        reader: &mut wincode::io::Reader<'de>,
        dst: &mut MaybeUninit<Self::Dst>,
    ) -> wincode::ReadResult<()> {
        let sig_bytes = reader.read_borrowed(96)?;
        // NOTE: We can now map the error here.
        let sig = BlstSignature::deserialize(sig_bytes)
            .map_err(|_| wincode::ReadError::Custom("bad BLS encoding"))?;
        dst.write(IndividualSignature(sig));
        wincode::ReadResult::Ok(())
    }
}
```

Closes #14.

### DIFF
--- a/wincode/src/error.rs
+++ b/wincode/src/error.rs
@@ -1,7 +1,7 @@
 //! Error types and helpers.
 use {crate::io, core::str::Utf8Error, thiserror::Error};
 
-#[derive(Error, Debug, PartialEq, Eq, Clone)]
+#[derive(Error, Debug, PartialEq, Eq, Clone, Copy)]
 pub enum Error {
     #[error(transparent)]
     WriteError(#[from] WriteError),
@@ -9,7 +9,7 @@ pub enum Error {
     ReadError(#[from] ReadError),
 }
 
-#[derive(Error, Debug, PartialEq, Eq, Clone)]
+#[derive(Error, Debug, PartialEq, Eq, Clone, Copy)]
 pub enum WriteError {
     #[error(transparent)]
     Io(#[from] io::WriteError),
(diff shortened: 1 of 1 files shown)


wincode/src/error.rs
```diff
@@ -1,25 +1,33 @@
...
+#[derive(Error, Debug, PartialEq, Eq, Clone)]
 pub enum WriteError {
     #[error(transparent)]
     Io(#[from] io::WriteError),
     #[error(transparent)]
     InvalidUtf8Encoding(#[from] Utf8Error),
     #[error("Sequence length would overflow length encoding scheme: {0}")]
     LengthEncodingOverflow(&'static str),
+    #[error("Custom error: {0}")]
+    Custom(Box<str>),
```
cpubot: This would need to be gated by `#[cfg(feature = "alloc")]`, so we should either require `&'static str` or feature gate appropriately. My gut is that we should go with `&'static str` for maximum target flexibility

cpubot (changes requested):
Thanks for the contribution! Just a note regarding feature gating or using `&'static str`

qkniep:
Yes, `&'static str` should be good enough in most cases.

Now I made the changes as minimal as possible, just adding the two `Custom` error variants.

PR #16, opened 2025-10-20
Add compile-time type metadata to tune implementations

cpubot:
# Summary

This reworks various internals of `wincode` to provide more compile time information that we can use to better optimize `SchemaWrite` and `SchemaRead` implementations for `core`, `alloc`, and user defined types. By lifting these details into the compiler, we can infer the optimizations available that a user would currently have to explicitly specify manually (by annotating with `Pod`, for example).

Commits are broken down logically to ease the review process.

The core observation is that types can be classified in three ways that directly influence how their implementations can be optimized.

1. Dynamic: the type has a variable length (`Vec<T>`, `Box<[T]>`, etc).
2. Static: the type has a compile time known size, and may have some specific runtime initialization (e.g., `Box<T>`) or validation (endianness, layout, etc).
3. Zero-copy: The type has a compile time known size _and_ can be serialized / deserialized using its in-memory representation directly.

This classification can be represented with the following `enum`:
```rs
enum TypeMeta {
    Static {
        size: usize,
        zero_copy: bool,
    },
    Dynamic,
}
```

And is now a constant on `SchemaWrite` and `SchemaRead` traits:
https://github.com/anza-xyz/wincode/blob/342563208f11b4acdf9c72ed800699f98e8222c4/wincode/src/schema/mod.rs#L86-L95

https://github.com/anza-xyz/wincode/blob/342563208f11b4acdf9c72ed800699f98e8222c4/wincode/src/schema/mod.rs#L98-L118

This makes this information available to implementors at compile time.

# `wincode::io` changes

Depending on the `TypeMeta` of a particular type, its `SchemaRead` / `SchemaWrite` implementation may need a `Reader` or `Writer` with adjusted internal semantics than those provided by the current implemenation. In particular, if it is known that a compound type (like a struct or tuple) is comprised entirely of fields with `TypeMeta::Static`, ideally the implementation could sum the total serialized size of all fields and fetch a single contiguous block of memory for reading / writing, and subsequent reads / writes within that block would elide bounds checking.

To support this, `Reader` and `Writer` have been factored into traits. This allows us to define typical bounds-checked `Reader` and `Writer`s with bounds-check-elided counterparts (c6af1bd314c8b3b8aaa9fa7c38a1ddfbffdb3272). This sets the stage for implementations of `SchemaWrite` and `SchemaRead` on compound types to leverage buffer prefetch logic when their constituent types support it. This is the most significant breaking change of the changeset. `SchemaRead` and `SchemaWrite` no longer depend on `Reader` and `Writer` directly, but `impl Reader` and `impl Writer`.

# The rest of the changes

Outside of adding `TypeMeta` to `SchemaRead` / `SchemaWrite` and factoring `Reader` and `Writer` into traits, the rest of the changeset updates the internals to support these changes. This includes:
- defining `TypeMeta` on all types that aren't `Dynamic` (0696d75015e3c7c879345a461c8bb5f98e9bf3bd)
- updating tuple, `SchemaWrite`, and `SchemaRead` derive macros to derive `TypeMeta` based on their constituent types (and struct `repr`s in the case of `zero_copy`) (044c77a50b15385da0e8151462ec7172a2c3b9aa, bead12c72a7b0a477dfda5d9e1445263e08bea9f)
- adding compile time branches to implementations of higher order types (`Vec<T>`, `Box<[T]>`, `[T; N]`, etc) that make use of buffer prefetching (87b28c67508c792b9b0e587bc74744359432f8ac)
- tests that exercise both `Static` and `Dynamic` variants of the implementations (342563208f11b4acdf9c72ed800699f98e8222c4)

# What's not included

There are further optimizations that can be made when a type is `Static { zero_copy: true }`. We can also add some cool features that generalize the zero-copy deserialize implementation for `&'de [u8]` to any type with `Static { zero_copy: true }`. I have decided to save those for a subsequent PRs, since this PR has already grown so large.

# Summary of breaking changes

#### Hard breaking
- `Reader` and `Writer` are now traits and method names have been refined

#### API changes
- `TypeMeta` added to `SchemaWrite` and `SchemaRead` traits (not _hard_ breaking because we provide a default and derive macros will provide implementations on user defined types)

# Results

In microbenchmarks, these updates show a 6-10x (depending on the size of the input) improvement in both serializing and deserializing over the current version of `wincode` on data that can leverage the `TypeMeta::Static` designation. (datasets like structs or `Vec`s of structs that contain primitive types)

### DIFF
--- a/wincode/src/schema/mod.rs
+++ b/wincode/src/schema/mod.rs
@@ -245,11 +245,30 @@ mod tests {
         Hash,
     )]
     #[wincode(internal)]
-    struct SomeStruct {
+    struct StructStatic {
         a: u64,
         b: bool,
-        #[wincode(with = "Pod<_>")]
         e: [u8; 32],
+    }
+
+    #[derive(
+        serde::Serialize,
+        serde::Deserialize,
+        Debug,
+        PartialEq,
+        Eq,
+        Ord,
+        PartialOrd,
+        SchemaWrite,
+        SchemaRead,
+        proptest_derive::Arbitrary,
+        Hash,
+    )]
+    #[wincode(internal)]
+    struct StructNonStatic {
+        a: u64,
+        b: bool,
+        e: String,
     }
 
     thread_local! {
@@ -714,16 +733,16 @@ mod tests {
 
     #[test]
     fn test_phantom_data() {
-        let val = PhantomData::<SomeStruct>;
+        let val = PhantomData::<StructStatic>;
         let serialized = serialize(&val).unwrap();
         let bincode_serialized = bincode::serialize(&val).unwrap();
         assert_eq!(&serialized, &bincode_serialized);
         assert_eq!(
-            PhantomData::<SomeStruct>::size_of(&val).unwrap(),
+            PhantomData::<StructStatic>::size_of(&val).unwrap(),
             bincode::serialized_size(&val).unwrap() as usize
         );
-        let deserialized: PhantomData<SomeStruct> = deserialize(&serialized).unwrap();
-        let bincode_deserialized: PhantomData<SomeStruct> =
+        let deserialized: PhantomData<StructStatic> = deserialize(&serialized).unwrap();
+        let bincode_deserialized: PhantomData<StructStatic> =
             bincode::deserialize(&bincode_serialized).unwrap();
         assert_eq!(deserialized, bincode_deserialized);
     }
@@ -795,14 +814,29 @@ mod tests {
         }
 
         #[test]
-        fn test_vec_elem(vec in proptest::collection::vec(any::<SomeStruct>(), 0..=100)) {
+        fn test_vec_elem_static(vec in proptest::collection::vec(any::<StructStatic>(), 0..=100)) {
             let bincode_serialized = bincode::serialize(&vec).unwrap();
-            type Target = containers::Vec<Elem<SomeStruct>, BincodeLen>;
+            type Target = containers::Vec<Elem<StructStatic>, BincodeLen>;
             let schema_serialized = Target::serialize(&vec).unwrap();
 
             prop_assert_eq!(&bincode_serialized, &schema_serialized);
 
-            let bincode_deserialized: Vec<SomeStruct> = bincode::deserialize(&bincode_serialized).unwrap();
+            let bincode_deserialized: Vec<StructStatic> = bincode::deserialize(&bincode_serialized).unwrap();
+            let schema_deserialized = Target::deserialize(&schema_serialized).unwrap();
+            prop_assert_eq!(&vec, &bincode_deserialized);
+            prop_assert_eq!(vec, schema_deserialized);
+        }
+
+
+        #[test]
+        fn test_vec_elem_non_static(vec in proptest::collection::vec(any::<StructNonStatic>(), 0..=16)) {
+            let bincode_serialized = bincode::serialize(&vec).unwrap();
+            type Target = containers::Vec<Elem<StructNonStatic>, BincodeLen>;
+            let schema_serialized = Target::serialize(&vec).unwrap();
+
+            prop_assert_eq!(&bincode_serialized, &schema_serialized);
+
+            let bincode_deserialized: Vec<StructNonStatic> = bincode::deserialize(&bincode_serialized).unwrap();
             let schema_deserialized = Target::deserialize(&schema_serialized).unwrap();
             prop_assert_eq!(&vec, &bincode_deserialized);
             prop_assert_eq!(vec, schema_deserialized);
@@ -823,7 +857,7 @@ mod tests {
         }
 
         #[test]
-        fn test_serialize_slice(slice in proptest::collection::vec(any::<SomeStruct>(), 0..=100)) {
+        fn test_serialize_slice(slice in proptest::collection::vec(any::<StructStatic>(), 0..=100)) {
             let bincode_serialized = bincode::serialize(slice.as_slice()).unwrap();
             let schema_serialized = serialize(slice.as_slice()).unwrap();
             prop_assert_eq!(&bincode_serialized, &schema_serialized);
@@ -844,14 +878,28 @@ mod tests {
         }
 
         #[test]
-        fn test_vec_deque_elem(vec in proptest::collection::vec_deque(any::<SomeStruct>(), 0..=100)) {
+        fn test_vec_deque_elem_static(vec in proptest::collection::vec_deque(any::<StructStatic>(), 0..=100)) {
             let bincode_serialized = bincode::serialize(&vec).unwrap();
-            type Target = containers::VecDeque<Elem<SomeStruct>, BincodeLen>;
+            type Target = containers::VecDeque<Elem<StructStatic>, BincodeLen>;
             let schema_serialized = Target::serialize(&vec).unwrap();
 
             prop_assert_eq!(&bincode_serialized, &schema_serialized);
 
-            let bincode_deserialized: VecDeque<SomeStruct> = bincode::deserialize(&bincode_serialized).unwrap();
+            let bincode_deserialized: VecDeque<StructStatic> = bincode::deserialize(&bincode_serialized).unwrap();
+            let schema_deserialized = Target::deserialize(&schema_serialized).unwrap();
+            prop_assert_eq!(&vec, &bincode_deserialized);
+            prop_assert_eq!(vec, schema_deserialized);
+        }
+
+        #[test]
+        fn test_vec_deque_elem_non_static(vec in proptest::collection::vec_deque(any::<StructNonStatic>(), 0..=16)) {
+            let bincode_serialized = bincode::serialize(&vec).unwrap();
+            type Target = containers::VecDeque<Elem<StructNonStatic>, BincodeLen>;
+            let schema_serialized = Target::serialize(&vec).unwrap();
+
+            prop_assert_eq!(&bincode_serialized, &schema_serialized);
+
+            let bincode_deserialized: VecDeque<StructNonStatic> = bincode::deserialize(&bincode_serialized).unwrap();
             let schema_deserialized = Target::deserialize(&schema_serialized).unwrap();
             prop_assert_eq!(&vec, &bincode_deserialized);
             prop_assert_eq!(vec, schema_deserialized);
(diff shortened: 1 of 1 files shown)


wincode/src/io.rs
```diff
@@ -23,29 +17,79 @@ fn read_size_limit(len: usize) -> ReadError {
...
+    ///
+    /// Useful for sections of code where there is enough information to bulk prefetch bytes
+    /// with a single bounds check for the entire chunk.
+    type Trusted: Reader<'a>;
+    /// Peek up to `n_bytes` bytes without consuming them.
+    ///
+    /// This is _not_ required to return exactly `n_bytes`, it is required
+    /// to return _up to_ `n_bytes`, so take care to check the length
+    /// if you need to ensure you get exactly `n_bytes`.
+    fn peek_buffered(&self, n_bytes: usize) -> &'a [u8];
```
alessandrod: you don't like newlines?

not asking you to change - your crate your style, just telling you that I hate
it 😂

wincode/src/io.rs
```diff
@@ -23,29 +17,79 @@ fn read_size_limit(len: usize) -> ReadError {
...
+    /// Consume exactly `amt` bytes without checking bounds.
+    ///
+    /// May panic if the reader does not have enough bytes.
+    fn consume_unchecked(&mut self, amt: usize);
+    /// Consume exactly `amt` bytes, returning an error if the reader does not have enough bytes.
+    fn consume(&mut self, amt: usize) -> ReadResult<()>;
+    /// Bulk read `n_bytes` with a single bounds check, returning a [`Reader`]
+    /// that can elide bounds checking for the entire chunk.
+    ///
+    /// Should return an error if the reader does not have enough space.
```
alessandrod: must return an error right?

wincode/src/io.rs
```diff
@@ -23,29 +17,79 @@ fn read_size_limit(len: usize) -> ReadError {
...
+    fn read_array<const N: usize>(&mut self) -> ReadResult<&'a [u8; N]> {
+        let src = self.read_slice(N)?;
+        // SAFETY:
+        // - `read_slice` ensures we read N bytes.
+        Ok(unsafe { &*src.as_ptr().cast::<[u8; N]>() })
     }
+    /// Consume exactly `amt` bytes without checking bounds.
+    ///
+    /// May panic if the reader does not have enough bytes.
+    fn consume_unchecked(&mut self, amt: usize);
```
alessandrod: shouldn't this be unsafe?

wincode/src/io.rs
```diff
@@ -220,184 +310,466 @@ impl<'a> Writer<'a> {
     /// # Safety
     ///
     /// - `T` must be plain ol' data.
+    /// - `src` must not overlap with the internal buffer.
     #[inline]
-    pub unsafe fn write_t<T>(&mut self, value: &T) -> WriteResult<()> {
-        self.write_slice_t(slice::from_ref(value))
+    unsafe fn write_from_t<T>(&mut self, src: &T) -> WriteResult<()> {
```
alessandrod: why the _from renames?

wincode/src/io.rs
```diff
@@ -220,184 +310,466 @@ impl<'a> Writer<'a> {
...
+    }
+
+    /// Create a new [`TrustedSliceWriter`] from an untrusted buffer, providing
+    /// the length of the original source.
+    ///
+    /// Prefer [`TrustedSliceWriter::new`] if you're starting from a fresh buffer.
+    pub fn from_source(
+        buffer: &'a mut [MaybeUninit<u8>],
+        source_len: usize,
+        global_pos: usize,
```
alessandrod: I found global_pos confusing. source_offset?

wincode/src/io.rs
```diff
@@ -220,184 +310,466 @@ impl<'a> Writer<'a> {
...
+            return Err(writer_trailing_bytes(rem));
+        }
+
+        Ok(final_pos)
+    }
+
+    #[inline]
+    #[allow(clippy::arithmetic_side_effects)]
+    fn get_slice_mut(&mut self, len: usize) -> WriteResult<&mut [MaybeUninit<u8>]> {
+        let dst = &mut self.buffer[self.pos..self.pos + len];
```
alessandrod: here and in all the other trusted impls - can't you use get_unchecked if this is trusted?

wincode/src/schema/containers.rs
```diff
@@ -329,7 +340,7 @@ where
     /// # Safety
     ///
     /// - `T::read` must properly initialize elements.
-    fn read(reader: &mut Reader<'de>, dst: &mut MaybeUninit<Self::Dst>) -> ReadResult<()> {
+    fn read(reader: &mut impl Reader<'de>, dst: &mut MaybeUninit<Self::Dst>) -> ReadResult<()> {
         if type_equal::<T::Dst, u8>() {
             return <Vec<Pod<u8>, Len>>::read(reader, unsafe {
```
alessandrod, line 345: there are some places where we're missing safety docs for unsafe uses, maybe
enable the lint?

wincode/src/io.rs
cpubot: I often don't use them when just defining trait methods, but I don't feel strongly 🙂

wincode/src/io.rs
cpubot: Yes -- done

wincode/src/io.rs
cpubot: Done

wincode/src/io.rs
cpubot: Absolutely

wincode/src/schema/containers.rs
cpubot: Not a bad idea, though it can be a bit overbearing on extremely obvious cases -- we at least have it required for unsafe trait method docs. Perhaps something we can do it in a follow up PR?

wincode/src/io.rs
cpubot: Was trying to make it more consistent with `Reader` methods, but I've changed the naming there and I think we're in a good place now

wincode/src/io.rs
cpubot: Yes it must

wincode/src/serde.rs
```diff
@@ -88,22 +126,33 @@ pub trait Serialize: SchemaWrite {
     /// Serialize a serializable type into a `Vec` of bytes.
     #[cfg(feature = "alloc")]
     fn serialize(src: &Self::Src) -> WriteResult<Vec<u8>> {
-        let mut buffer = Vec::with_capacity(Self::size_of(src)?);
-        let len = Self::serialize_into(src, buffer.spare_capacity_mut())?;
+        let capacity = Self::size_of(src)?;
+        let mut buffer = Vec::with_capacity(capacity);
+        let mut writer = buffer.spare_capacity_mut();
+        Self::serialize_into(src, &mut writer)?;
+        let len = writer.len();
```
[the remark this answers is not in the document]
cpubot: Wrong, `writer` is `&mut [MaybeUninit<u8>]` (for which `Writer` is implemented)

wincode/src/io/mod.rs
```diff
@@ -0,0 +1,258 @@
...
+
+    /// Copy and consume exactly `dst.len() * size_of::<T>()` bytes from the [`Reader`] into `dst`.
+    ///
+    /// # Safety
+    ///
+    /// - `T` must be initialized by reads of `size_of::<T>()` bytes.
+    /// - `dst` must not overlap with the internal buffer.
+    #[inline]
+    unsafe fn copy_into_slice_t<T>(&mut self, dst: &mut [MaybeUninit<T>]) -> ReadResult<()> {
+        let len = size_of_val(dst);
```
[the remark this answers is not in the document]
cpubot: Doesn't need to be imported

wincode-derive/src/common.rs
```diff
@@ -66,44 +149,124 @@ impl Field {
...
+                )
             }
         })
     }
 }
 
+fn anon_ident_iter(prefix: Option<&str>) -> impl Iterator<Item = Ident> + Clone + use<'_> {
+    let prefix = prefix.unwrap_or("");
+    ('a'..='z').cycle().enumerate().map(move |(i, ch)| {
+        let wrap = i / 26;
```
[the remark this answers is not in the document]
cpubot: Irrelevant and useless suggestion

wincode/src/io/mod.rs
```diff
@@ -0,0 +1,258 @@
...
+    type Trusted<'a>: Writer
+    where
+        Self: 'a;
+
+    /// Perform any cleanup or flushing of the writer.
+    ///
+    /// Be advised that this may be called on trusted [`Writer`] variants for subsections of the writer,
+    /// so make sure that those implementations don't perform any cleanup operations that would
+    /// invalidate the parent writer.
+    fn finish(&mut self) -> WriteResult<()> {
```
alessandrod, line 226: This is a bit confusing to me. From what I understand it must be called from the
outermost SchemaWrite, but when is it also called for subsections? If you have
nested collections? An example would help

wincode/src/schema/mod.rs
```diff
@@ -134,15 +188,17 @@ where
...
 /// Variant of [`write_elem_iter`] specialized for slices, which can opt into
 /// an optimized implementation for bytes (`u8`s).
-fn write_elem_slice<T, Len>(writer: &mut Writer, src: &[T::Src]) -> WriteResult<()>
+fn write_elem_slice<T, Len>(writer: &mut impl Writer, src: &[T::Src]) -> WriteResult<()>
 where
     Len: SeqLen,
     T: SchemaWrite,
     T::Src: Sized,
 {
     if type_equal::<T::Src, u8>() {
```
alessandrod, line 197: would branching on T::TYPE_META not work here?

wincode-derive/src/common.rs
```diff
@@ -66,44 +149,124 @@ impl Field {
...
+                quote! { #(#items),* }
+            }
+        };
+        // No need to prefix, as this is only used in a struct context, where the static size is
+        // known at compile time.
+        let anon_idents = self.member_anon_ident_iter(None).collect::<Vec<_>>();
+        let zero_copy_idents = self.member_anon_ident_iter(Some("zc_")).collect::<Vec<_>>();
+        let is_zero_copy_eligible = repr.is_zero_copy_eligible();
+        quote! {
+            if let (#(TypeMeta::Static { size: #anon_idents, zero_copy: #zero_copy_idents }),*) = (#tuple_expansion) {
```
alessandrod, line 229: this is insane lmao

it's obviously cool, but it took me a good 5 minutes to understand what's going
on, please add a comment that explains what's happening

alessandrod (approved):
Amazing work. It's a big PR, but it's split so well it was a joy to review.

I don't really have any notes other than it took me a while to grasp the derive struct impl, and the semantics of finish() are not totally clear.

I really think that we should aim to have 100% unsafe coverage for this crate, since I suspect that it'll be more popular than we expect, but I'll leave that up to you and in any case yeah can be done in a follow up PR.

wincode/src/schema/mod.rs
cpubot: It certainly would, and that is the plan! Just opted to replace these `type_equal` sections with `T::TYPE_META` in a follow up PR to keep this one smaller. I plan to overhaul how this and containers work generally to leverage  `TYPE_META` in a way that is much nicer from a user API perspective after this is merged.

wincode-derive/src/common.rs
cpubot: Can do!

wincode/src/io/mod.rs
cpubot: `finish` is called at the end of the outermost `SchemaWrite` (as you mentioned), but also called when `TrustedWriter`s go out of scope (not via drop, but explicitly when they're finished). 

An example of where this is useful is file io: the trusted writer may do some buffering of file bytes to reduce syscalls, and `finish` can be implemented to actually flush remaining bytes in that buffer to the file. This could of course just be done in the outermost writer, but the outer writer and trusted writer may not share buffers (of course they _could_) and felt reasonable to allow implementors to have access to that same hook for trusted sections.

cpubot: Example of where it's called:
https://github.com/anza-xyz/wincode/blob/97082ef9492d9af78c27beba929cebba3b3a0ca4/wincode/src/schema/mod.rs#L169-L176

wincode/src/schema/mod.rs
cpubot: https://github.com/anza-xyz/wincode/pull/19

PR #17, opened 2025-10-28
Add Cursor and Vec Reader/Writer impls

cpubot:
Some quality of life implementations for `Reader` and `Writer`.

#16 adds the barebones necessary implementations of `Reader` and `Writer` over slices, but practically speaking users will want more ergonomics when interfacing with `wincode`; for example being able to serialize into a `Vec`, or being able to serialize _multiple_ times into a `Vec` (like [`std::io::Cursor<&mut Vec>`](https://doc.rust-lang.org/std/io/struct.Cursor.html#impl-Write-for-Cursor%3C%26mut+Vec%3Cu8,+A%3E%3E) provides).

This adds a new `Cursor` struct, and the following implementations:
- `Reader for Cursor<T> where T: AsRef<[u8]>`
- `Writer for Cursor<&mut [MaybeUninit<u8>]>`
- `Writer for Cursor<&mut MaybeUninit<[u8; N]>>`
- `Writer for Cursor<&mut Vec<u8>>`
- `Writer for Cursor<Vec<u8>>`

This allows serializing / deserializing into / from the same source multiple times or from specific offsets in the source.

Also included is a `Writer for Vec<u8>` implementation, which simply appends to the `Vec` (like [`std::io::Write`'s implementation](https://doc.rust-lang.org/std/vec/struct.Vec.html#impl-Write-for-Vec%3Cu8,+A%3E)).

### DIFF
--- a/wincode/src/io/cursor.rs
+++ b/wincode/src/io/cursor.rs
@@ -360,13 +360,19 @@ mod vec {
 /// Trusted writer for `Cursor<&mut Vec<u8>>` or `Cursor<Vec<u8>>` that continues
 /// overwriting the vector's memory.
 ///
-/// This will _not_ grow the vector, as it assumes the caller has already reserved enough capacity.
+/// Generally this should not be constructed directly, but rather by calling [`Writer::as_trusted_for`]
+/// on a trusted [`Writer`]. This will ensure that the safety invariants are upheld.
 ///
 /// Note that this does *not* update the length of the vector, as it only contains a reference to the
 /// vector's memory via `&mut [MaybeUninit<u8>]`, but it will update the _position_ of the cursor.
 /// Vec implementations will synchronize the length and position on subsequent writes or when the
 /// writer is finished. Benchmarks showed a roughly 2x performance improvement using this method
 /// rather than taking a `&mut Vec<u8>` directly.
+///
+/// # Safety
+///
+/// - This will _not_ grow the vector, as it assumes the caller has already reserved enough capacity.
+///   The `inner` buffer must have sufficient capacity for all writes. It is UB if this is not upheld.
 #[cfg(feature = "alloc")]
 pub struct TrustedVecWriter<'a> {
     inner: &'a mut [MaybeUninit<u8>],
--- a/wincode/src/io/vec.rs
+++ b/wincode/src/io/vec.rs
@@ -2,7 +2,14 @@ use super::*;
 
 /// Trusted writer for `Vec<u8>` that continues appending to the vector's spare capacity.
 ///
-/// This will _not_ grow the vector, as it assumes the caller has already reserved enough capacity.
+/// Generally this should not be constructed directly, but rather by calling [`Writer::as_trusted_for`]
+/// on a [`Vec<u8>`]. This will ensure that the safety invariants are upheld.
+///
+/// # Safety
+///
+/// - This will _not_ grow the vector, and it will not bounds check writes, as it assumes the caller has
+///   already reserved enough capacity. The `inner` Vec must have sufficient capacity for all writes.
+///   It is UB if this is not upheld.
 pub struct TrustedVecWriter<'a> {
     inner: &'a mut Vec<u8>,
 }
--- a/wincode/src/io/mod.rs
+++ b/wincode/src/io/mod.rs
@@ -41,6 +41,10 @@ const fn read_size_limit(len: usize) -> ReadError {
 /// Returns [`ReadError::UnsupportedZeroCopy`] for readers that do not support zero-copy.
 pub trait Reader<'a> {
     /// A variant of the [`Reader`] that can elide bounds checking within a given window.
+    ///
+    /// Trusted variants of the [`Reader`] should generally not be constructed directly,
+    /// but rather by calling [`Reader::as_trusted_for`] on a trusted [`Reader`].
+    /// This will ensure that the safety invariants are upheld.
     type Trusted<'b>: Reader<'a>
     where
         Self: 'b;
@@ -205,6 +209,10 @@ pub type WriteResult<T> = core::result::Result<T, WriteError>;
 /// Trait for structured writing of bytes into a source of potentially uninitialized memory.
 pub trait Writer {
     /// A variant of the [`Writer`] that can elide bounds checking within a given window.
+    ///
+    /// Trusted variants of the [`Writer`] should generally not be constructed directly,
+    /// but rather by calling [`Writer::as_trusted_for`] on a trusted [`Writer`].
+    /// This will ensure that the safety invariants are upheld.
     type Trusted<'a>: Writer
     where
         Self: 'a;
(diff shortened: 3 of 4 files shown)


wincode/src/io/cursor.rs
```diff
@@ -0,0 +1,610 @@
...
+/// }
+///
+/// let mut data = MaybeUninit::<[u8; 8]>::uninit();
+///
+/// let mut cursor = Cursor::new(&mut data);
+/// let bytes = rand_bytes();
+/// cursor.write(&bytes).unwrap();
+/// assert_eq!(unsafe { data.assume_init() }, bytes);
+///
+/// let mut cursor = Cursor::new(&mut data);
```
alessandrod, line 32: not super clear that you're running things again to check that cursor pos resets
to 0, maybe add a comment?

wincode/src/io/cursor.rs
```diff
@@ -0,0 +1,610 @@
...
+        Self { inner, pos: 0 }
+    }
+
+    /// Creates a new cursor at the given position.
+    ///
+    /// # Safety
+    /// - `pos` must be less than or equal to the length of the inner type.
+    ///
+    /// Providing a position greater than the length of the inner type will result in undefined behavior.
+    pub const unsafe fn new_at(inner: T, pos: usize) -> Self {
```
alessandrod: it's strange that this is unsafe and new isn't? 0 can be out of range too

wincode/src/io/cursor.rs
```diff
@@ -0,0 +1,610 @@
...
+    /// Returns a slice of the remaining bytes in the cursor.
+    #[inline]
+    fn cur_slice(&self) -> &[u8] {
+        // SAFETY: `pos` is less than or equal to the length of the slice.
+        unsafe { self.inner.as_ref().get_unchecked(self.pos..) }
+    }
+
+    /// Returns the number of bytes remaining in the cursor.
+    #[inline]
+    fn cur_len(&self) -> usize {
```
alessandrod: nit: I'd call this remaining

alessandrod (approved):

wincode/src/io/cursor.rs
cpubot: Will do!

wincode/src/io/cursor.rs
cpubot: Good catch for sure. I think I'll just adjust the slice logic a bit so that we bounds check when slicing at the current position

cpubot: Actually, is a position of `0` UB?

The only implementations where this could in theory be problematic is the slice impls.

I believe `slice.get_unchecked(0..)` is fine on an empty slice

wincode/src/io/cursor.rs
cpubot: 95f76c2257776f4ae35a8d469eaabe6e75b990ac

wincode/src/io/vec.rs
```diff
@@ -0,0 +1,164 @@
...
+impl<'a> Writer for TrustedVecWriter<'a> {
+    type Trusted<'b>
+        = TrustedVecWriter<'b>
+    where
+        Self: 'b;
+
+    fn write(&mut self, src: &[u8]) -> WriteResult<()> {
+        let spare = self.inner.spare_capacity_mut();
+        // SAFETY: Creator of this writer ensures we have sufficient capacity for all writes.
+        unsafe { ptr::copy_nonoverlapping(src.as_ptr(), spare.as_mut_ptr().cast(), src.len()) };
```
vadorovsky, line 32: This will cause UB if the spare capacity is smaller than `src`. I know that the point of `TrustedVecWriter` is to never grow, but I still think we could return an error, especially since `WriteError::WriteSizeLimit` looks like a great fit.

Or if we really want to shift the responsibility on the caller, this method needs to be `unsafe`, but that defeats the usage of the trait.

cpubot: The expectation of `TrustedWriter`s (and `TrustedReader`s) generally is that they are bounds checked (and in the case of the `Vec`, has sufficient capacity) before instantiation.

vadorovsky: Fair. But that expectation means this method needs to be `unsafe` (it can cause UB if misused), especially given that it's public. By making the method public and safe we are not making the expectation clear. And it's not even 100% clear from reading any docstring (the second paragraph of the `TrustedVector`'s description mentions capacity, but doesn't say that `write` is unchecked and can cause UB), so the caller needs to explicitly read the code to understand the risk. :upside_down_face: 

I don't think that adding a check with an error will kill the performance. And adding a separate trait with unsafe methods for writing seems like unnecessary work (I would be happy with it though if you think otherwise).

cpubot: I certainly understand where you're coming from, but it's not expected that the `Trusted` variants of `Reader`s and `Writer`s are constructed by users. They're really only public because they exist as associated types on the non-trusted variants. We actually do have documentation that calling `as_trusted_for` must ensure that `n_bytes` are available in the source / destination, and that behavior is unspecified if that is not upheld.
https://github.com/anza-xyz/wincode/blob/4ae814f33b5c0bd460bbff7f19f6aa5e14f590e7/wincode/src/io/mod.rs#L97-L104
https://github.com/anza-xyz/wincode/blob/4ae814f33b5c0bd460bbff7f19f6aa5e14f590e7/wincode/src/io/mod.rs#L233-L238

Calling `write` on a trusted variant in isolation is indeed unsafe, but acquiring a trusted variant via `as_trusted_for` (and subsequently calling `write` on it) should not be.

Perhaps I can add documentation to all the trusted variants that indicates that they should generally not manually constructed, and that doing so and using them without satisfying their invariants will lead to UB. What do you think about that?

--- 

> I don't think that adding a check with an error will kill the performance.

Removing the bounds checking within the trusted variants is actually the impetus for the abstraction in the first place. Without intermediate bounds checks between writes, the memcpys can be vectorized, which yields a significant performance improvement

> And adding a separate trait with unsafe methods for writing seems like unnecessary work

It also doesn't work in practice. The fact that they're unified over `Writer` or `Reader` means that `SchemaRead` and `SchemaWrite` operate over the same interface, which means that `SchemaRead` and `SchemaWrite` don't need separate implementations depending on whether the reader/writer is trusted. In theory implementing `Writer for T where T: TrustedWriter` would solve this, but the blanket implementation precludes us from making specific implementations for the various types.

vadorovsky: OK, thanks for the explanation and sorry for missing that!

> Perhaps I can add documentation to all the trusted variants that indicates that they should generally not manually constructed, and that doing so and using them without satisfying their invariants will lead to UB. What do you think about that?

Yes, that sounds great. Or maybe even just extending the safety comment to something like:

```
// SAFETY: Creator of this writer ensures we have sufficient capacity for all writes,
// and the writer is never constructed manually.
```

would've already prevented my past self from writing this comment. :sweat_smile:

vadorovsky (approved):
Thanks!

alessandrod (approved):

PR #19, opened 2025-11-08
Automatically infer Pod; deprecate Elem

cpubot:
Now that we have [compile time metadata about the size and zero-copy nature of types](https://github.com/anza-xyz/wincode/pull/16), we can simplify `container` usage significantly.

Refer to #16 for more details on `TypeMeta`.

### Background on `container`

The `container` module provides `SchemaRead` and `SchemaWrite` implementations for built-in types (`Vec`, `VecDeque`, `LinkedList`, etc) with an additional type parameter that allows configuring a length encoding. Additionally (as of current `master`), these `SchemaRead` and `SchemaWrite` implementations are _only_ implemented for either `Elem<T>` and `Pod<T>` (e.g., `Vec<Elem<T>>` or `Vec<Pod<T>>`). The `Pod` case allows users to assert that the collection's inner type `T`'s serialized representation is the same as its in-memory representation. `Elem` is used for all other cases. So, `container`s serve double duty: 
1. Specifying a custom length encoding.
2. Specifying the POD-nature (or lack thereof) of a collection's inner type `T`.

Refer to the [docs](https://docs.rs/wincode/latest/wincode/containers/index.html) for additional explanation and examples.

The reason we do not currently provide blanket implementations for _any_ type `T` (i.e., users must specify `Elem<T>` or `Pod<T>`) is that we cannot garner enough information at compile time for an arbitrary `T` to determine whether we can optimize the implementation. The reason we pack both length encoding and type metadata information into `container` implementations is that separating these out into separate container types for the matrix of length encodings and type metadata encodings would result in a proliferation of container types and additional cognitive burden for users / annoyingly large API surface.

#16 lays the foundation to solve this by introducing a `TypeMeta` constant on `SchemaRead` and `SchemaWrite`, which provides the necessary compile time information to provide blanket implementations for `container` types (i.e., we can _infer_ `Pod` or `Elem` without user intervention).

#### This PR

We can now provide blanket implementations across all container types and use the compile-time `TypeMeta` on the inner type to determine the implementation. As such specific (non-blanket) implementations for `Container<Elem<T>>` and `Container<Pod<T>>` are no longer necessary. This also makes `Elem` completely unnecessary / useless, so it is now marked deprecated. Pass-through `SchemaWrite` and `SchemaRead` implementations of `Elem` are now implemented for backwards compatibility -- container declarations that use `Elem` will see no change in behavior, given the new compile-time `TypeMeta` analysis (`Elem<T>` forwards directly to `T`).

The result of this is that users generally should never have to annotate with `Pod` for:
- Raw byte types (`u8`, `[u8; N]`)
- `SchemaWrite` / `SchemaRead` derived structs comprised entirely of those raw byte types annotated with `#[repr(transparent)]` or `#[repr(C)]`.
- Sequence containers (like `containers::Vec<T>`, `containers::VecDeque<T>`, `container::Box<[T]>`, etc) where `T` is any of the above types.

In other words, `wincode` should generally do the optimal thing without any user intervention.

`Pod` is preserved, as I believe it still has utility for users that explicitly want to ignore things like endianness or where the user is interfacing with a foreign POD type for which they cannot derive `SchemaWrite` or `SchemaRead`.

### DIFF
--- a/wincode-derive/src/schema_read.rs
+++ b/wincode-derive/src/schema_read.rs
@@ -2,7 +2,7 @@ use {
     crate::common::{
         default_tag_encoding, extract_repr, get_crate_name, get_src_dst,
         get_src_dst_fully_qualified, suppress_unused_fields, Field, FieldsExt, SchemaArgs,
-        StructRepr, TraitImpl, TypeExt, Variant,
+        StructRepr, TraitImpl, TypeExt, Variant, VariantsExt,
     },
     darling::{
         ast::{Data, Fields, Style},
@@ -123,8 +123,12 @@ fn impl_struct(
                     unsafe { reader.copy_into_t(dst)? };
                 }
                 TypeMeta::Static { size, zero_copy: false } => {
+                    // SAFETY: `size` is the serialized size of the struct, which is the sum
+                    // of the serialized sizes of the fields.
+                    // Calling `read` on each field will consume exactly `size` bytes,
+                    // fully consuming the trusted window.
+                    let reader = &mut unsafe { reader.as_trusted_for(size) }?;
                     #init_guard
-                    let reader = &mut reader.as_trusted_for(size)?;
                     #(#read_impl)*
                     mem::forget(guard);
                 }
@@ -256,20 +260,7 @@ fn impl_enum(
     let default_tag_encoding = default_tag_encoding();
     let tag_encoding = tag_encoding.unwrap_or(&default_tag_encoding);
 
-    // Note that all enums except unit enums are never static.
-    let mut type_meta_impl = quote!(TypeMeta::Dynamic);
-    if variants.iter().all(|variant| variant.fields.is_unit()) {
-        // If all variants are unit, we know up front that the static size is the size of the discriminant.
-        type_meta_impl = quote! {
-            const {
-                match <#tag_encoding as SchemaRead<'de>>::TYPE_META {
-                    // Unit enums are never zero-copy, as they have invalid bit patterns.
-                    TypeMeta::Static { size, .. } => TypeMeta::Static { size, zero_copy: false },
-                    TypeMeta::Dynamic => TypeMeta::Dynamic,
-                }
-            }
-        };
-    }
+    let type_meta_impl = variants.type_meta_impl(TraitImpl::SchemaRead, tag_encoding);
 
     let read_impl = variants.iter().enumerate().map(|(i, variant)| {
         let variant_ident = &variant.ident;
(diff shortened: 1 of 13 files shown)


wincode-derive/src/schema_read.rs
```diff
@@ -116,15 +118,22 @@ fn impl_struct(
             }
 
             match <Self as SchemaRead<'de>>::TYPE_META {
-                TypeMeta::Static { size, .. } => {
+                TypeMeta::Static { zero_copy: true, .. } => {
+                    // SAFETY: `T` is zero-copy eligible (no invalid bit patterns, no layout requirements, no endianness checks, etc.).
+                    unsafe { reader.copy_into_t(dst)? };
```
kskalski, line 123: somehow I find it odd that in reaction to `zero_copy==true` we need to invoke `copy`... I suppose it's the mismatch of semantics between `zero_copy` (as discussed in other places, it's akin to transmute safety) and what reader needs to do - it actually reads data from its source to the destination, so it always needs to copy, unless it operates on struct with references, which make the "copy" just copy the refs not the content, right?

seems like I keep having similar confusion about `zero_copy`, maybe it should just use a more precise name like `can_transmute` or even something relating to the compatibility with bincode serialization format (I remember your comments about enum not being `zero_copy` because some bytes are not exactly as bincode serializes them) `has_bincode_byte_layout`

cpubot: `zero_copy` is a property of the type as it relates to the serialization format. It denotes that the type is _eligible_ for zero-copy (i.e., the serialized bytes == the in-memory bytes), but does not impose that the type _must_ be interpreted as a reference. In that vein, if a type's serialized bytes == its in-memory bytes, it's safe to copy as raw bytes when the user wants a copy or the reader itself does not support zero-copy.

In other words, `zero_copy` denotes a _capability_, not a mandate.

wincode/src/serde.rs
```diff
@@ -148,14 +164,23 @@ pub trait Serialize: SchemaWrite {
...
+            TypeMeta::Static {
+                size,
+                zero_copy: false,
+            } => {
                 // SAFETY: `Self::TYPE_META` specifies a static size, so a single write of `Self::Src`
                 // will consume `size` bytes, fully consuming the trusted window.
-                let mut writer = unsafe { dst.as_trusted_for(size) }?;
-                Self::write(&mut writer, src)?;
-                writer.finish()?;
+                let trusted = &mut unsafe { dst.as_trusted_for(size) }?;
```
kskalski: the previous name looked more informational, maybe it should be `trusted_dst` or `trusted_writer`

cpubot: Yeah looks like i accidentally clobbered it when fixing merge conflicts. Now fixed

bw-solana (commented):
LGTM.

Obviously inferring the ability to ZC is critical to getting this right. My understanding is
- #16 is where most of that inference happens.
- Additional padding check for ZC eligibility was added in #25.
- And #27 added strict checking for enum variants.

The only other pitfall I can think of has to do with endianness. I think basically everyone runs on little endian targets, but we still run miri on big endian targets for vote state view for example. Do we have any safeguards there that I'm missing?

bw-solana:
I'm good with what you have.

But waiting to click the button in case Kamil or Alessandro have more comments

alessandrod (approved):

PR #22, opened 2025-11-10
Allow specifying custom enum discriminant encodings

cpubot:
We have various cases in agave and alpenglow where the default serde/bincode enum discriminant encoding is overridden to use something like a `u8`. This is currently unsupported in `wincode-derive`, requiring users to manually implement `SchemaRead` / `SchemaWrite` for those enums.

This PR adds another top-level derive attribute that allows specifying the schema to use when encoding enum discriminants. Any type that implements `SchemaRead` and `SchemaWrite` is supported, so the user is free to use any built-in type or their own custom encoding.

Note this also includes: #23 

#### Custom encoding
```rs
#[wincode(tag_encoding = "u8")]
enum Enum {
    A { name: String, id: u64 },
    B(String, Vec<u8>),
    C,
}
```

#### Custom tag
```rs
enum Enum {
    #[wincode(tag = 1)]
    A { name: String, id: u64 },
    #[wincode(tag = 2)]
    B(String, Vec<u8>),
    #[wincode(tag = 3)]
    C,
}
```

### DIFF
--- a/wincode-derive/src/common.rs
+++ b/wincode-derive/src/common.rs
@@ -403,13 +403,24 @@ pub(crate) struct SchemaArgs {
     /// Specifies whether to generate placement initialization struct helpers on `SchemaRead` implementations.
     #[darling(default)]
     pub(crate) struct_extensions: bool,
-    /// Specifies the encoding to use for enum variants.
+    /// Specifies the encoding to use for enum discriminants.
     ///
-    /// If specified, the enum variants will be encoded using the given type's `SchemaWrite`
+    /// If specified, the enum discriminants will be encoded using the given type's `SchemaWrite`
     /// and `SchemaRead` implementations.
-    /// Otherwise, the enum variants will be encoded using the default encoding (`u32`).
+    /// Otherwise, the enum discriminants will be encoded using the default encoding (`u32`).
     #[darling(default)]
-    pub(crate) variant_encoding: Option<Type>,
+    pub(crate) tag_encoding: Option<Type>,
+}
+
+/// The default encoding to use for enum discriminants.
+///
+/// Bincode's default discriminant encoding is `u32`.
+///
+/// Note in the public APIs we refer to `tag` to mean the discriminant encoding
+/// for friendlier naming.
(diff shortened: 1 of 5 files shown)


kskalski:
> We have various cases in agave and alpenglow where the default serde/bincode enum encoding is overridden to use something like a `u8`. This is currently unsupported in `wincode-derive`, requiring users to manually implement `SchemaRead` / `SchemaWrite` for those enums.
> 
> This PR adds another top-level derive attribute that allows specifying the schema to use when encoding enum variants. Any type that implements `SchemaRead` and `SchemaWrite` is supported, so the user is free to use any built-in type or their own custom encoding.
> 
> Note this also includes: #23
> #### Custom encoding
> 
> ```rust
> #[wincode(variant_encoding = "u8")]

Here and in the code-base there is confusion whether it's about *variant* or *discriminant* encoding.
I guess this attribute won't force the way variants are encoded, right?

This attribute should be renamed to `discriminant_encoding` (or `discriminant_repr`, or `tag_repr` / `tag_encoding` -> I guess `tag` is a nicer word and it would match the attributes per-variant, but on the other hand it's not consistent with how Rust calls discriminants https://doc.rust-lang.org/reference/items/enumerations.html#r-items.enum.discriminant, on the other hand `serde` uses `tag` for this https://serde.rs/container-attrs.html#tag)

kskalski:
Yes, one of the possible interpretations is that `variant_encoding` affects how say `Enum::A` variant including its data is encoded (not that it makes much sense, but the confusion remains). In a way it specifies how part of the variant is encoded, so since `Enum::C` variant is composed only of discriminant, then it in fact affects its entire encoding. It is just too ambiguous when enum contains variants with discriminant + data.

I guess `tag_*` will be best, especially that you already use `tag` in attributes attached to variants `#[wincode(tag = 3)]`, so it will naturally relate to the values specified there.

Now for the actual implementation, I guess there is no reason to use ambiguous naming there at all, but for consistency the code should operate also on `tag` and just mention `discriminant` in comments.

PR #24, opened 2025-11-13
Add ZeroCopy marker and SchemaRead support

cpubot:
We have simple zero-copy support for byte slices (`&'de [u8]`), but we can provide enough compile time information to make this more robust.

This PR adds a marker trait `ZeroCopy` that is used to provide blanket `SchemaRead` implementations for:
- `&'de T where T: SchemaRead<'de> + ZeroCopy`
- `&'de [T] where T: SchemaRead<'de> + ZeroCopy`

Primitive types that implement `ZeroCopy` are:
- `u8`
- `i8`
- `[T; N] where T: ZeroCopy`

`#[derive(SchemaRead)]` will implement `ZeroCopy` on structs when its `TypeMeta` is `TypeMeta::Static { zero_copy: true }`

### Vs `TypeMeta`

#16 introduced `TypeMeta` which also provides compile-time `zero_copy` information. So, why do we need this?

`TypeMeta` is geared for usage inside `SchemaWrite` and `SchemaRead` implementations. Implementations can provide optimized compile-time branches depending on a type's `TypeMeta`. Due to limitations in the type system, `TypeMeta` is unusable for implementing something like a `ZeroCopy` marker trait. In particular, constants in generics can't be used to constrain implementations. E.g., we cannot 
```rs
struct Assert<const B: bool>;
trait IsTrue {}
impl IsTrue for Assert<true> {}

impl<'de, T> ZeroCopy for T 
where 
    T: SchemaRead<'de>,
    Assert<{ matches!(T::TYPE_META, TypeMeta::Static { zero_copy: true, .. }) }>: IsTrue,
```
And as such we can't do anything similar to provide `impl<'de, T> SchemaRead<'de> for T where SchemaRead<'de> ...TypeMeta...`

Similarly, we could not use the `ZeroCopy` marker trait to provide blanket implementations for built in types like `Vec`, `Box`, etc. because the compiler limits us to a single blanket implementation for each type, so we'd have to get into newtypes, which would yield a terrible API.

So, there's a bit of duplication between where we advertise zero-copy, but the surface area is small since only `u8`, `i8` and `[T;N]` are marked zero-copy and it's a relatively small amount of code to support. Users generally wont have to interface with any of these mechanics, so it's just a net win from a user-facing perspective.

### DIFF
--- a/wincode/src/schema/impls.rs
+++ b/wincode/src/schema/impls.rs
@@ -1144,7 +1144,7 @@ mod zero_copy {
     #[inline(always)]
     fn cast_ensure_aligned<T, U>(ptr: *const T) -> ReadResult<*const U> {
         let ptr = ptr.cast::<U>();
-        if align_of::<T>() > 1 && !ptr.is_aligned() {
+        if align_of::<U>() > 1 && !ptr.is_aligned() {
             return Err(unaligned_pointer_read());
         }
         Ok(ptr)


wincode/src/schema/impls.rs
```diff
@@ -1123,3 +1126,162 @@ where
...
+
+    fn read(reader: &mut impl Reader<'de>, dst: &mut MaybeUninit<Self::Dst>) -> ReadResult<()> {
+        let TypeMeta::Static {
+            size,
+            zero_copy: true,
+        } = T::TYPE_META
+        else {
+            // Should be impossible to reach.
+            unreachable!("Type is not zero-copy");
+        };
```
kskalski, line 1276: the `const TYPE_META` declaration above guarantees the same at compile time, so I guess this could just do `let size = T::TYPE_META.size;`

wincode/src/schema/impls.rs
```diff
@@ -1123,3 +1126,162 @@ where
...
+
+        let len = <BincodeLen>::read::<T::Dst>(reader)?;
+        #[allow(clippy::arithmetic_side_effects)]
+        let bytes = reader.borrow_exact(len * size)?;
```
kskalski: is this overflow case actually worth handling at runtime? deserialized data could be user-provided and even if the overflow actually lowers the size of `bytes` the safety comment below won't hold, i.e. bytes.len() != len * size and 
`unsafe { zero_copy::cast_slice_to_slice_t::<T::Dst>(bytes, len)? }` becomes scary

wincode/src/schema/impls.rs
cpubot: As far as I know this isn't possible without a `match` or `if let` to extract `TypeMeta::Static`. The condition is constant, so it should be compiled out anyway

wincode/src/schema/impls.rs
cpubot: Good point -- I'll add a check

cpubot: Actually, this is safe at the moment because `BincodeLen` has a size limit check.
https://github.com/anza-xyz/wincode/blob/f15758b0c17a3422ed82b7ef53b5c545b52f4889/wincode/src/len.rs#L43-L50

But perhaps we should do something different here because there's no preallocation occurring like all other cases, and 4MiB is probably too restrictive for borrowed payloads

cpubot: https://github.com/anza-xyz/wincode/pull/24/commits/0f134af08f5d57cd9064e28bc2585ef2632233d9

wincode/src/schema/impls.rs
kskalski: Ah, right, might make sense to add const accessor to `TypeMeta` since this kind of code already occur in several places and will probably in more:
```rust
impl TypeMeta {
    const fn expect_static_size(self) -> usize {
        match self {
            TypeMeta::Static { size, .. } => size,
            TypeMeta::Dynamic => panic!("type is dynamic"),
        }
    }
}

...
        const STATIC: TypeMeta = TypeMeta::Static {
            size: 8,
            zero_copy: true,
        };
        const DYN: TypeMeta = TypeMeta::Dynamic;
        const SIZE: usize = DYN.expect_static_size();
```
giving
```
689 |         const SIZE: usize = DYN.expect_static_size();
    |                             ^^^^^^^^^^^^^^^^^^^^^^^^ evaluation panicked: type is dynamic
    |
note: inside `schema::TypeMeta::expect_static_size`
   --> wincode/src/schema/mod.rs:89:34
    |
89  |             TypeMeta::Dynamic => panic!("type is dynamic"),
    |                                  ^^^^^^^^^^^^^^^^^^^^^^^^^ the failure occurred here
```

I believe this would also allow simplifying some definitions of TYPE_META in the traits, as they could do
```rust
    const TYPE_META: TypeMeta =
        TypeMeta::Static {
            size: T::TYPE_META.expect_static_size(),
            zero_copy: true,
        }
```

(maybe it's an idea for separate PR though)

cpubot: Yeah, definitely think we could add some helpers to `TypeMeta`, but agree something for a separate PR.

wincode/src/schema/mod.rs
```diff
@@ -2058,6 +2123,92 @@ mod tests {
...
+        #[repr(C)]
+        struct MyStruct {
+            #[wincode(with = "Pod<_>")]
+            address: Address,
+        }
+
+        #[derive(SchemaWrite, SchemaRead, Debug, PartialEq, Eq)]
+        #[wincode(internal)]
+        struct MyStructRef<'a> {
+            inner: &'a MyStruct,
```
kskalski, line 2203: are we able to test here that Pod<_> implements `ZeroCopy`? I guess we would need a field that uses ref to a Pod... maybe
```
#[wincode(with = "&Pod<_>")]
address: &Address,
```
?

wincode/src/schema/mod.rs
```diff
@@ -2058,6 +2123,92 @@ mod tests {
...
 
+    #[test]
+    fn test_struct_zero_copy_refs() {
+        // Owned zero-copy type.
+        #[derive(SchemaWrite, SchemaRead, Debug, PartialEq, Eq, proptest_derive::Arbitrary)]
+        #[wincode(internal)]
+        #[repr(C)]
+        struct Zc {
+            a: u8,
+            b: [u8; 64],
```
kskalski, line 2134: I suppose we should test i8 and [i8; N] too

wincode/src/schema/mod.rs
cpubot: We are technically testing for this given that `MyStruct` is a reference in `MyStructRef`. It wouldn't compile if MyStruct wasn't marked as `ZeroCopy`. Because `Address` doesn't itself derive `SchemaRead`, the fact that it's annotated with `Pod` is what allows it to compile.

I'm unsure if this would compile, but I can try
```rs
#[wincode(with = "&Pod<_>")]
address: &Address,
```

cpubot: Apparently it does work! 78e085adadb52291d93ae8bda6122df228cfb366

wincode/src/schema/mod.rs
cpubot: Sure, can add

cpubot: 418bead4ac12fae191149367f0ddab3936773940

PR #26, opened 2025-11-14
Make trusted reader/writer variants private + `unsafe as_trusted_for`

cpubot:
### Background
The notion of "trusted" `Reader` and `Writer`s was introduced in #16.

In short, trusted `Reader` and `Writer`s are implementations of the `Reader` and `Writer` traits that can be used when a serializer / deserializer for a type determines that the type that it is encoding / decoding has a compile-time known serialized size. They can call `reader.as_trusted_for(n_bytes)` to specify "I assert that the next set of reads or writes performed on this `Reader` / `Writer` will read/write precisely `n_bytes`". This allows the `Reader` / `Writer` to elide bounds checks within that `n_bytes` window, which yields a significant performance improvement, especially on large datasets.

### This PR
Trusted `Reader` and `Writer` types must be public because they're referenced as associated types of the untrusted variants. E.g.,
https://github.com/anza-xyz/wincode/blob/1a79bc32aa39264e4d3f5fe7d381a183850d20dd/wincode/src/io/slice.rs#L156-L160

But, we should prevent them from being manually constructed given the potential for unsafe usage with safe APIs (e.g., we should prevent users calling `TrustedSliceReader::new()` manually). These trusted variants are always available via `as_trusted_for` on the corresponding untrusted variant, and they are safe to use if acquired with `as_trusted_for` and the safety contract is upheld (no reads or writes are attempted past the requested `n_bytes`).

Additionally in line with the above, `as_trusted_for` is now marked as `unsafe`. Even though `as_trusted_for` does an initial bounds check, there is nothing preventing the user from over reading / writing through safe APIs after the initial bounds check succeeds. Marking this function as `unsafe` forces the user to be aware of the safety invariants.

### DIFF
--- a/wincode/src/serde.rs
+++ b/wincode/src/serde.rs
@@ -52,7 +52,10 @@ pub trait Deserialize<'de>: SchemaRead<'de> {
     fn deserialize_into(mut src: &'de [u8], dst: &mut MaybeUninit<Self::Dst>) -> ReadResult<()> {
         match Self::TYPE_META {
             TypeMeta::Static { size, .. } => {
-                Self::read(&mut src.as_trusted_for(size)?, dst)?;
+                // SAFETY: `Self::TYPE_META` specifies a static size, so a single read of `Self::Dst`
+                // will consume `size` bytes, fully consuming the trusted window.
+                let mut reader = unsafe { src.as_trusted_for(size) }?;
+                Self::read(&mut reader, dst)?;
             }
             _ => {
                 Self::read(&mut src, dst)?;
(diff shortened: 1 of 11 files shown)


wincode/src/serde.rs
```diff
@@ -143,16 +149,19 @@ pub trait Serialize: SchemaWrite {
...
         match Self::TYPE_META {
             TypeMeta::Static { size, .. } => {
-                Self::write(&mut dst.as_trusted_for(size)?, src)?;
-                dst.finish()?;
-                Ok(())
+                // SAFETY: `Self::TYPE_META` specifies a static size, so a single write of `Self::Src`
+                // will consume `size` bytes, fully consuming the trusted window.
+                let mut writer = unsafe { dst.as_trusted_for(size) }?;
+                Self::write(&mut writer, src)?;
+                writer.finish()?;
```
kskalski, line 156: was this `writer.finish()` missing here before?

cpubot: It was, yes

kskalski: I wonder if there could be some kind of mechanism for warning about that... or even just logging in trusted writers that they are dropped without finish, bug I suppose that would introduce overhead and it can't be forced on all implementors to do it

cpubot: It's a good idea -- I've thought about this too. Could put add a `flushed` flag and something like this behind `#[cfg(debug_assertions)]` in the trusted writers we provide
```rs
debug_assert!(
    !std::thread::panicking() && self.finished,
    "MyWriter dropped without finish() being called"
);
```

But you're right we can't force all implementors to do it, but may provide reasonable assertions for the crate internally

cpubot: In fact I may do this in a follow up PR -- I like the idea

alessandrod (approved):

PR #27, opened 2025-11-14
Implement more robust TypeMeta inference for enums

cpubot:
Currently, enums are only `TypeMeta::Static` if all variants are unit enums. While a reasonable starting point, a more general method for inferring a whether an enum is statically sized would actually check the size of all variants for equality. This subsumes the existing method (check if all variants are unit) and enables more valid cases.

Trusted `Reader`s and `Writer`s can only be instantiated if a type has `TypeMeta::Static`, and types are only `TypeMeta::Static` if all their constituent types are `TypeMeta::Static`. This means that there is a fairly significant performance penalty for using an enum in a struct or a sequence that has otherwise statically sized members. So, more robust `TypeMeta` inference has significant performance upside.

This PR overhauls the `TypeMeta` inference for enums to compare the sizes of all enum variants and return `TypeMeta::Static` if all variants have the same serialized size (this trivially holds for enums with a single variant).

For example:

The following enum's `TypeMeta` will yield `TypeMeta::Static { size: 12, zero_copy: false }`
```rs
enum Enum {
    A {
        a: u64,
    },
    B {
        x: u32,
        y: u32,
    },
    C {
        a: u8,
        b: u8,
        c: u8,
        d: u8,
        e: u8,
        f: u8,
        g: u8,
        h: u8,
    },
}
```

Similarly the following will yield `TypeMeta::Static { size: 37, zero_copy: false }` (single variant)

```rs
enum Enum {
    A { a: u8, b: [u8; 32] },
}
```

### DIFF
--- a/wincode/src/schema/mod.rs
+++ b/wincode/src/schema/mod.rs
@@ -981,6 +981,38 @@ mod tests {
     }
 
     #[test]
+    fn enum_unit_and_non_unit_dynamic() {
+        #[derive(
+            SchemaWrite,
+            SchemaRead,
+            Debug,
+            PartialEq,
+            proptest_derive::Arbitrary,
+            serde::Serialize,
+            serde::Deserialize,
+        )]
+        #[wincode(internal)]
+        enum Enum {
+            Unit,
+            NonUnit(u8),
+        }
+
+        assert_eq!(<Enum as SchemaWrite>::TYPE_META, TypeMeta::Dynamic);
+        assert_eq!(<Enum as SchemaRead<'_>>::TYPE_META, TypeMeta::Dynamic);
+
+        proptest!(proptest_cfg(), |(e: Enum)| {
+            let serialized = serialize(&e).unwrap();
+            let bincode_serialized = bincode::serialize(&e).unwrap();
+            prop_assert_eq!(&serialized, &bincode_serialized);
+
+            let deserialized: Enum = deserialize(&serialized).unwrap();
+            let bincode_deserialized: Enum = bincode::deserialize(&bincode_serialized).unwrap();
+            prop_assert_eq!(&deserialized, &bincode_deserialized);
+            prop_assert_eq!(deserialized, e);
+        });
+    }
+
+    #[test]
     fn test_phantom_data() {
         let val = PhantomData::<StructStatic>;
         let serialized = serialize(&val).unwrap();


wincode-derive/src/common.rs
```diff
@@ -307,6 +305,128 @@ impl Variant {
...
+        if self.is_empty() {
+            return quote! { TypeMeta::Static { size: 0, zero_copy: false } };
+        }
+
+        // Enums have a statically known size in a very specific case: all variants have the same serialized size.
+        // This holds trivially for enums where all variants are unit enums (the size is just the size of the discriminant).
+        // In other cases, we need to compute the size of each variant and check if they are all equal.
+        // Otherwise, the enum is dynamic.
+        //
+        // Enums are never zero-copy, as the discriminant may have invalid bit patterns.
```
kskalski, line 324: Invalid bit patterns influence the safety of transmute, but not necessarily whether type is zero-copyable (in the common-sense meaning - I suppose currently in wincode those two concepts are not distinguished), right?

It would also different for write vs read operations:
- write -> serialization is safe (in-mem enum has always correct bit-patterns) and zero-copyable
- read -> deserialization should read some of the bytes to verify valid bits and return Result, but no copying needs to be done

wincode/src/schema/mod.rs
kskalski, line 844: should there be a test that checks that unit enum's type meta is static?

wincode/src/schema/mod.rs
```diff
@@ -885,6 +886,100 @@ mod tests {
...
+
+    #[test]
+    fn enum_dynamic_non_uniform_variants() {
+        #[derive(SchemaWrite, SchemaRead, Debug, PartialEq, proptest_derive::Arbitrary)]
+        #[wincode(internal)]
+        enum Enum {
+            A { a: u64 },
+            B { x: u32, y: u32 },
+            C { a: u8, b: u8 },
+        }
```
kskalski, line 945: maybe also add a test checking
```
enum EnumMix {
   Unit,
   NonUnit(u8)
}
```
is dynamic

wincode-derive/src/common.rs
kskalski: Having separate marker for "needs verifycation, but not copying" is probably more of an idea for better optimizations support in future, however given that `SchemaWrite` and `SchemaRead` have their own `TYPE_META`, would it make sense in this case to make enums zero-copyable in `SchemaWrite` and only non-zero-copyable in `SchemaRead`?

wincode/src/schema/mod.rs
cpubot: We have these tests:
https://github.com/anza-xyz/wincode/blob/9957c4a11f3a950fc6398dec3a88be934bbb75ea/wincode/src/schema/mod.rs#L783-L834

wincode-derive/src/common.rs
cpubot: Let me give a breakdown of `zero_copy: true` first before answering your questions

### zero-copy

`zero-copy`, as used in wincode, means the type's serialized bytes are the same as its in-memory representation. I.e., it is more analogous to transmute safety. 

You're correct that we don't distinguish between something being completely transmute safe and something having "zero-copy parts" with validity checks at the `TypeMeta` level (it always means the former). But practically speaking, as of https://github.com/anza-xyz/wincode/pull/24, a type with `zero-copy: true` _is_ valid to specify as a reference in a `SchemaRead` derive without any adjustments to `TypeMeta`. In other words, we get "_zero-copy with validity checks_" by virtue of having `zero_copy: true` -- and enums naturally fit into this. For example, it is valid to define the following enum:
```rs
pub enum MyEnumRef<'a> {
    A { a: &'a u8 },
    B { b: &'a [u8; 64] },
}
```

Because `u8` and `[u8; 64]` are `zero_copy: true`, they can be used as references in compound types like structs, tuples, and enums. And #24 makes this possible with structs as well. E.g., the following would receive `zero_copy: true`
```rs
#[repr(transparent)]
struct Address([u8; 32]);
```
which makes it valid to use as a reference in a compound type.

So in summary, you can think of `zero_copy: true` meaning that "it's valid to use a reference to this type in a `SchemaRead` derive".

You're also correct that this is generally trivial for `SchemaWrite`, and we already have a blanket implementation for `SchemaWrite for &'a T`, regardless of whether the type is `zero-copy: true`.

### enums and `zero_copy`

As explained above, `zero_copy: true` means the type's in memory-representation is the same as its serialized bytes. Even if an enum _itself_ is not `zero_copy: true`, this doesn't mean that enums can't be used with `zero-copy: true` _parts_.

Enums could theoretically be `zero_copy: true` for _writes_ with a narrow set of qualifications and some caveats:
1. Fieldless enums with a `#[repr]` that is also `zero_copy: true` (e.g., `u8`).
    - This would also require us to ensure that encoded discriminant values match assigned values (if any). E.g., `enum { A = 3 }` would need the serialized discriminant to be `3`, which would not be "default bincode compatible". If users _don't_ specify specific values (i.e., the discriminant is the integer field ordering), this would work, but still not "default bincode compatible" unless the discriminant matches bincode's default `u32` and the user is on a LE platform.
2. `#[repr(C, Int)]` enums where `Int` is some `zero_copy: true` type and none of the variants have padding (bincode doesn't serialize padding).
    - This has similar issues with not being "default bincode compatible" unless the discriminant is encoded as `u32` and the user is on an LE platform.
    - `TypeMeta` doesn't have a variant like `TypeMeta::DynamicZeroCopy`, so we'd additionally have to ensure that all variants have the same size (so that we can produce `TypeMeta::Static`) or introduce another `TypeMeta` variant.

For _reads_, enums can basically never be `zero_copy: true`, unless all the above write conditions are met, variants all have the same size with no padding, _and_ the number of variants matches the cardinality of the discriminant (e.g., `#[repr(u8)]` with 256 variants).

These are technically feasible to implement, and perhaps we implement them later on, but they're low priority in my estimation given how specific the qualifications are.

Additionally, with #24, we get "zero-copy with validation", which should cover a fairly broad number of cases. Even without #24 it's still possible to use "zero-copy with validation" today for enums that contain byte slices (we have a `SchemaRead` impl for `&[u8]`) -- #24 just broadens the valid cases for this substantially.

wincode/src/schema/mod.rs
cpubot: Done: 7a4ba321606faab03e7c7fb5be62ba6bc4791658

alessandrod (approved):

PR #33, opened 2025-11-17
Optimize char deserialization with manual UTF-8 decoder

tanmay4l:
addresses the TODO comment at line 247-250 which noted: _"Could implement a manual decoder that avoids UTF-8
   validate + chars() and instead performs the UTF-8 validity checks and produces a `char` directly. Some quick 
  micro-benchmarking revealed a roughly 2x speedup is possible."_

  ## Changes

  **Before:**
  ```rust
  let str = core::str::from_utf8(buf).map_err(invalid_utf8_encoding)?;
  let c = str.chars().next().unwrap();

  After:
  - Manual UTF-8 decoding for 2-4 byte characters using bit masks
  - Inline validation of continuation bytes (must be 10xxxxxx)
  - Overlong encoding validation (3-byte: >= U+0800, 4-byte: >= U+10000)
  - Surrogate validation (rejects U+D800..U+DFFF)
  - Out of range validation (rejects > U+10FFFF)


TanmayDhobale:
do we need this, or should I close it? @cpubot

kskalski:
You mention using a microbenchmark? Does it make sense to include it in the PR (e.g. add to `wincode/benches`) and get the comparison numbers in the PR description?

kskalski:
Thanks, I used the code and did a few other changes on top of it including the benchmark too in https://github.com/anza-xyz/wincode/pull/187

PR #40, opened 2025-11-28
feat: implement consolidated benchmarking suite

tanmay4l:
cc @cpubot

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,7 +21,7 @@ jobs:
       - name: rustfmt (check)
         run: cargo +nightly fmt --all -- --check
       - name: miri (check)
-        run: cargo +nightly miri test --all-features --all-targets --no-fail-fast
+        run: cargo +nightly miri test --all-features --lib --bins --no-fail-fast
 
   lint-test:
     name: clippy & tests (stable)


wincode/Cargo.toml
```diff
@@ -29,14 +29,21 @@ solana-short-vec = { version = "3.0.0", optional = true }
...
+criterion = { version = "0.5.1", features = ["html_reports"] }
 
 [build-dependencies]
 proc-macro2.workspace = true
 quote.workspace = true
+
+[[bench]]
+name = "benchmarks"
+harness = false
+
```
cpubot, line 49: Should also add `required-features = ["std", "derive"]`

wincode/benches/benchmarks.rs
```diff
@@ -0,0 +1,159 @@
...
+        );
+        
+        // Vec<Pod<PodStruct>> - optimized
+        #[derive(SchemaWrite, SchemaRead)]
+        struct WithPod {
+            #[wincode(with = "wincode::containers::Vec<Pod<_>>")]
+            data: Vec<PodStruct>,
+        }
+        
+        let with_pod = WithPod { data: data.clone() };
```
cpubot: This actually isn't measuring anything meaningful. `Pod` is no longer required for types comprised of bytes or byte arrays -- so these benchmarks are comparing virtually the same code. It would be useful to compare `Vec<PodStruct>` between `bincode` and `wincode` though.

wincode/benches/benchmarks.rs
```diff
@@ -0,0 +1,159 @@
...
+    
+    group.bench_function("bincode", |b| {
+        b.iter(|| bincode::deserialize::<SimpleStruct>(black_box(&serialized)).unwrap());
+    });
+    
+    group.finish();
+}
+
+fn bench_hashmap_comparison(c: &mut Criterion) {
+    let mut group = c.benchmark_group("HashMap<u64, u64>");
```
cpubot, line 153: We should add another one for HashMaps comprised of Pod, e.g., `HashMap<[u8; 16], PodStruct>`

wincode/benches/benchmarks.rs
```diff
@@ -0,0 +1,159 @@
...
+#[derive(Serialize, Deserialize, SchemaWrite, SchemaRead, Clone)]
+struct SimpleStruct {
+    id: u64,
+    value: u64,
+    flag: bool,
+}
+
+#[repr(transparent)]
+#[derive(Clone, Copy, SchemaWrite, SchemaRead, Serialize, Deserialize)]
+struct PodStruct([u8; 32]);
```
cpubot: Perhaps make this more interesting, like
```rs
#[repr(C)]
struct PodStruct {
    a: [u8; 32],
    b: [u8; 16],
    c: [u8; 8],
}
```

wincode/benches/benchmarks.rs
```diff
@@ -0,0 +1,159 @@
...
+            BenchmarkId::new("bincode", size),
+            &serialized,
+            |b, s| b.iter(|| bincode::deserialize::<Vec<u64>>(black_box(s)).unwrap()),
+        );
+    }
+    
+    group.finish();
+}
+
+fn bench_struct_comparison(c: &mut Criterion) {
```
cpubot, line 89: Perhaps add one for `PodStruct` too.

wincode/Cargo.toml
```diff
@@ -29,14 +29,21 @@ solana-short-vec = { version = "3.0.0", optional = true }
 thiserror = { version = "2.0.16", default-features = false }
 wincode-derive = { path = "../wincode-derive", optional = true }
 
+
```
cpubot: Extra space

wincode/Cargo.toml
tanmay4l: oh good catch.

wincode/benches/benchmarks.rs
tanmay4l: um yea,  I've replaced this with a wincode vs bincode comparison for Vec<PodStruct> instead.

wincode/benches/benchmarks.rs
```diff
@@ -108,39 +131,68 @@ fn bench_hashmap_comparison(c: &mut Criterion) {
...
     
     for size in [1_000, 10_000] {
-        let data: Vec<PodStruct> = (0..size).map(|i| PodStruct([i as u8; 32])).collect();
-        let bytes = (size * 32) as u64;
+        let data: Vec<PodStruct> = (0..size).map(|i| PodStruct {
+            a: [i as u8; 32],
+            b: [i as u8; 16],
+            c: [i as u8; 8],
+        }).collect();
+        let bytes = (size * 56) as u64;
```
cpubot: Should probably be `wincode::serialized_size(&data).unwrap()` to get an accurate reading of the total bytes serialized / deserialized

cpubot (commented):
Perhaps we should measure serialize as well?

tanmay4l:
> Perhaps we should measure serialize as well?

we can do same pattern for it , ill add serialization benchmarks

README.md
```diff
@@ -22,4 +22,20 @@ let val = MyStruct { data: vec![1,2,3], win: true };
...
+cargo bench --features derive
+```
+
+### Performance Results
+
+Preliminary benchmarks show `wincode` significantly outperforming `bincode`:
+
+- **Deserialization**: **3x faster** for simple structs (1.25ns vs 3.85ns)
+- **Vec<u64>**: **2x higher throughput** (28 GiB/s vs 12.8 GiB/s)
+- **Byte-array structs**: Optimized automatically for zero-copy deserialization
```
cpubot: Lets remove this from the README for now, and we can generate a more comprehensive markdown table later.

tanmay4l:
> Nice, work! Thanks

seems like benchmarks cannot run under miri because they use criterion which does file I/O and process spawning  know 

Benchmarks should NOT be run with miri  ?

tanmay4l:
> > > Nice, work! Thanks
> > 
> > 
> > seems like benchmarks cannot run under miri because they use criterion which does file I/O and process spawning know
> > Benchmarks should NOT be run with miri ?
> 
> Right, they shouldn't run with miri

ig we need updated the CI workflow to exclude benchmarks from miri (--lib --bins instead of --all-targets). should be good now sounds good to u ?

PR #41, opened 2025-12-01
feat(io): implement Writer directly for [u8] and [MaybeUninit<u8>]

kskalski:
#### Problem
`bincode` allows serializing directly into the `&mut [u8]` slice without tracking the length of the serialized data, which is useful in scenarios where user knows exactly the serialized representation size / layout.

In `wincode` the `Writer` trait is implemented for `&mut [u8]`, but since `write` function uses `&mut self`, the user is actually required to use `&mut &mut [u8]` for writes, which sometimes forces explicitly creating temporary variable.

`wincode`'s approach has a benefit of tracking how much of the slice was filled, since `write` will update the slice to point beyond written data, but it actually imposes slight performance penalty for doing that.

#### Changes
* add `impl` of `Writer` for `[u8]`
* add `impl` of `Writer` for `[MaybeUninit<u8>]`
* update current tests to use `writer = &mut buffer.as_mut_slice()` instead of `mut writer = &mut buffer.as_mut_slice()`, since now conversion of `mut var: &mut` to `Writer` is ambiguous (this is technically a breaking change)

### DIFF
--- a/wincode/src/io/slice.rs
+++ b/wincode/src/io/slice.rs
@@ -255,7 +255,10 @@ impl<'a> Writer for TrustedSliceWriter<'a> {
 /// Get a slice of `len` bytes for writing, advancing the writer by `len` bytes, or
 /// returning an error if the input slice does not have at least `len` bytes remaining.
 #[inline]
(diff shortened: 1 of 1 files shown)


cpubot (approved):
Nice, thank you

PR #42, opened 2025-12-01
Implement drop-safe struct_extensions builder

cpubot:
## Background

The [`struct_extensions` attribute](https://docs.rs/wincode/0.2.1/wincode/#struct_extensions), introduced in `v0.1.0`, provides convenience methods for manual `SchemaRead` implementations. In particular, it provides the following methods on the derived struct:

`uninit_<field_name>_mut`: Gets a mutable `MaybeUninit` projection to the `<field_name>` slot.
`read_<field_name>`: Reads into a `MaybeUninit`’s `<field_name>` slot from the given Reader.
`write_uninit_<field_name>`: Writes a `MaybeUninit`’s `<field_name>` slot with the given value.

While this does significantly reduce boilerplate, it doesn't do anything to help users with drop safety.

Consider the following example:

```rs
#[wincode(struct_extensions)]
struct Bytes {
    bytes: Vec<u8>,
    ar: Vec<[u8; 32]>,
}

struct Data {
    bytes: Bytes,
}

impl<'de> SchemaRead<'de> for Data {
    type Dst = Data;

        let mut bytes = MaybeUninit::<Bytes>::uninit();
        Bytes::read_bytes(reader, &mut bytes)?;
        Bytes::read_ar(reader, &mut bytes)?;
        let init = unsafe { bytes.assume_init() };
        dst.write(Data { bytes: init });
        Ok(())
    }
}
```

Because there is no tracking of initialized fields, an error in the middle of deserializing will not drop initialized fields. Specifically these lines from the above example:
```rs
Bytes::read_bytes(reader, &mut bytes)?;
// If the following line errors, the above Vec<u8> leaks
Bytes::read_ar(reader, &mut bytes)?;
```

This isn't a problem with `struct_extensions` per se, but rather an implicit challenge when manually initializing `MaybeUninit` structs.

## This PR

This PR reworks `struct_extensions` to provide drop-safe "Builder" structs that take care of field initialization bookkeeping. In particular, it generates a struct that takes a mutable reference to the `MaybeUninit<T>`, and provides the same functionality as the original, but with automatic initialization tracking and a `Drop` implementation that drops all initialized fields. The same code as above, with the builder:

```rs
#[wincode(struct_extensions)]
struct Bytes {
    bytes: Vec<u8>,
    ar: Vec<[u8; 32]>,
}

struct Data {
    bytes: Bytes,
}

impl<'de> SchemaRead<'de> for Data {
    type Dst = Data;

        let mut bytes = MaybeUninit::<Bytes>::uninit();
        let mut bytes_builder = BytesUninitBuilder::from_maybe_uninit_mut(&mut bytes);
        bytes_builder.read_bytes(reader)?;
        bytes_builder.read_ar(reader)?;
        bytes_builder.finish()
        let init = unsafe { bytes.assume_init() };
        dst.write(Data { bytes: init });
        Ok(())
    }
}
```

This code is completely drop safe. If deserialization fails, the initialized fields are dropped when the builder is dropped. `builder.assume_init_forget` forgets the builder so that drop logic is no longer executed, so that we can subsequently `assume_init` on the underlying `MaybeUninit`.

There are additional affordances for more complicated cases, like multiple nested structs, for example.

```rs
#[wincode(struct_extensions)]
struct Bytes {
    bytes: Vec<u8>,
    ar: Vec<[u8; 32]>,
}

#[wincode(struct_extensions)]
struct Data {
    bytes: Bytes,
    other_byte_vec: Vec<u8>,
}

struct Outer {
    data: Data,
}

impl<'de> SchemaRead<'de> for Outer {
    type Dst = Outer;

        let data = unsafe { &mut *(&raw mut (*dst.as_mut_ptr()).data).cast::<MaybeUninit<Data>>() };
        let mut data_builder = DataUninitBuilder::from_maybe_uninit_mut(data);
        unsafe {
            // Marks `bytes` field of `Data` as initialized if the function succeeds.
            data_builder.init_bytes_with(|bytes| {
                // Individual fields of `Bytes` are marked as initialized as reading progresses.
                // Initialized fields are dropped in reverse order on error or panic.
                let mut bytes_builder = BytesUninitBuilder::from_maybe_uninit_mut(bytes);
                bytes_builder.read_bytes(reader)?.read_ar(reader)?;
                bytes_builder.finish();
                Ok(())
            })?;
        }
        data_builder.read_other_byte_vec(reader)?;
        data_builder.finish();
        // Outer now fully initialized.
        Ok(())
    }
}
```

This could alternatively be written like the following (rather than `init_<field>_with` function), where nested struct is manually marked as initialized:
```rs
#[wincode(struct_extensions)]
struct Bytes {
    bytes: Vec<u8>,
    ar: Vec<[u8; 32]>,
}

#[wincode(struct_extensions)]
struct Data {
    bytes: Bytes,
    other_byte_vec: Vec<u8>,
}

struct Outer {
    data: Data,
}

impl<'de> SchemaRead<'de> for Outer {
    type Dst = Outer;

        let data = unsafe { &mut *(&raw mut (*dst.as_mut_ptr()).data).cast::<MaybeUninit<Data>>() };
        let mut data_builder = DataUninitBuilder::from_maybe_uninit_mut(data);
        // Project `bytes` field of `Data` into a `&mut MaybeUninit<Bytes>`.
        let mut bytes_builder =
            BytesUninitBuilder::from_maybe_uninit_mut(data_builder.uninit_bytes_mut());
        // Individual fields of `Bytes` are marked as initialized as reading progresses.
        // Initialized fields are dropped in reverse order on error or panic.
        bytes_builder.read_bytes(reader)?.read_ar(reader)?;
        // `Bytes` is fully initialized, so we forget the builder.
        bytes_builder.finish();

        // `bytes` field of `Data` can now be marked as initialized
        // so that it's dropped if the next field fails to read.
        unsafe { data_builder.assume_init_bytes() };

        data_builder.read_other_byte_vec(reader)?;

        // `Data` is fully initialized, so we forget the builder.
        data_builder.finish();
        // Outer now fully initialized.
        Ok(())
    }
}

```

### DIFF
--- a/wincode-derive/src/schema_read.rs
+++ b/wincode-derive/src/schema_read.rs
@@ -302,7 +302,7 @@ fn impl_struct_extensions(args: &SchemaArgs, crate_name: &Path) -> Result<TokenS
             let real_index = fields.len() - 1 - index;
             let field_ident = field.struct_member_ident(real_index);
             // The corresponding bit for the field.
-            let bit_set_index = LitInt::new(&(1 << real_index).to_string(), Span::call_site());
+            let bit_set_index = LitInt::new(&(1u128 << real_index).to_string(), Span::call_site());
             quote! {
                 if self.init_set & #bit_set_index != 0 {
                     // SAFETY: We are dropping an initialized field.
@@ -388,7 +388,7 @@ fn impl_struct_extensions(args: &SchemaArgs, crate_name: &Path) -> Result<TokenS
         let init_with_field_ident = format_ident!("init_{ident_string}_with");
 
         // The bit index for the field.
-        let index_bit = LitInt::new(&(1 << i).to_string(), Span::call_site());
+        let index_bit = LitInt::new(&(1u128 << i).to_string(), Span::call_site());
         let set_index_bit = quote! {
             self.init_set |= #index_bit;
         };


wincode/src/schema/mod.rs
```diff
@@ -750,6 +750,205 @@ mod tests {
...
+            proptest!(proptest_cfg(), |(test: Test)| {
+                let serialized = serialize(&test).unwrap();
+                let mut uninit = MaybeUninit::<Test>::uninit();
+                let reader = &mut serialized.as_slice();
+                let mut builder = TestUninitBuilder::from_maybe_uninit_mut(&mut uninit);
+                builder
+                    .read_a(reader)?
+                    .read_b(reader)?
+                    .read_c(reader)?;
+                let init = unsafe { builder.assume_init_mut() };
```
kskalski: I think typically this function would be called something like `into_assume_init_mut` or possibly `build_assume_init_mut`

wincode-derive/src/schema_read.rs
```diff
@@ -225,27 +230,232 @@ fn impl_struct_extensions(args: &SchemaArgs) -> Result<TokenStream> {
...
+            }
+
+            /// Assume the builder is fully initialized, and return a mutable reference to the inner `MaybeUninit` struct.
+            ///
+            /// The builder will be forgotten, so the drop logic will not longer run.
+            ///
+            /// # Safety
+            ///
+            /// Calling this when the content is not yet fully initialized causes undefined behavior: it is up to the caller
+            /// to guarantee that the `MaybeUninit<T>` really is in an initialized state.
```
kskalski: maybe we could assert that all `init_set` bits are all set...? 

I guess this could be over-restrictive, i.e. if user passed in `MaybeUninit` that was already partially initialized and won't like the builder to require calling assume init for those fields. This could be supported by providing `assume_all_init` though.

wincode-derive/src/schema_read.rs
```diff
@@ -205,6 +205,7 @@ fn impl_struct_extensions(args: &SchemaArgs) -> Result<TokenStream> {
     let dst = get_src_dst(args);
     let impl_generics = append_de_lifetime(&args.generics);
     let (_, ty_generics, where_clause) = args.generics.split_for_impl();
+    let builder_ident = format_ident!("{struct_ident}UninitBuilder");
```
kskalski, line 208: isn't polluting the namespace a concern here? should those builder be put into some `wincode_gen` mod-scope... just thinking, maybe it's not necessary

wincode-derive/src/schema_read.rs
```diff
@@ -225,27 +230,232 @@ fn impl_struct_extensions(args: &SchemaArgs) -> Result<TokenStream> {
...
+    };
+
+    // Generate the helper methods for the builder.
+    let builder_helpers = fields.iter().enumerate().map(|(i, field)| {
+        let target = field.target_resolved();
+        let target_reader_bound = target.with_lifetime("de");
+        let ident = field.struct_member_ident(i);
+        let ident_string = field.struct_member_ident_to_string(i);
+        let uninit_mut_ident = format_ident!("uninit_{ident_string}_mut");
+        let read_field_ident = format_ident!("read_{ident_string}");
```
kskalski, line 385: btw, is there some mechanism that prevents those generated functions form triggering "unused" lint errors? I know this is an annoying issue with `modular_bitfield` crate code-gen.

wincode-derive/src/schema_read.rs
cpubot: Yeah it's a good idea, but I didn't do it for the reason you mentioned.

I could add a function `is_init()` that checks if all tracking bits are set, which would allow users to add assertions around that if they want.

wincode/src/schema/mod.rs
cpubot: I liked `assume_init_mut` here because it mirrors [`assume_init_mut`](https://doc.rust-lang.org/stable/std/mem/union.MaybeUninit.html#method.assume_init_mut) from `MaybeUninit`.

wincode-derive/src/schema_read.rs
cpubot: I don't think it's necessarily a problem because the struct name is suffixed with `UninitBuilder`, so unlikely to conflict. This feature is also opt-in, and the docs specify that it will be added to the same namespace, so I think it's okay personally.

wincode-derive/src/schema_read.rs
cpubot: We could emit `#[allow(unused)]` on the struct and the methods. But I actually don't get any unused warnings when using this extension.

On a clean project
image

image

wincode-derive/src/schema_read.rs
cpubot: Added 38323b1c405917a4689f405b193d82ff28db0490

wincode/src/schema/mod.rs
kskalski: Well, this association is actually misleading, since `assume_init_mut` takes `&mut self` not `self`, so the caller above might wrongly assume that he is getting a reference to innards of builder without dropping it.
The normal rust convention is to use `into_*` for destructive conversions or `build` for builder pattern.

cpubot: Sure, I'll change it. I don't feel strongly

cpubot: f3b67514bc50128e9ef6d6d7dc7272aa4bb75bf1

wincode-derive/src/schema_read.rs
```diff
@@ -225,27 +230,249 @@ fn impl_struct_extensions(args: &SchemaArgs) -> Result<TokenStream> {
...
         }
-    ))
+    };
+
+    let builder_impl = {
+        let is_fully_init_mask = if fields.len() as u32 == builder_bit_set_bits {
+            quote!(#builder_bit_set_ty::MAX)
+        } else {
+            let field_bits = LitInt::new(&fields.len().to_string(), Span::call_site());
+            quote!(((1 as #builder_bit_set_ty) << #field_bits) - 1)
```
kskalski, line 330: I imagine this could be calculated as fully compile-time constant, right?
```rust
let field_init_bits = LitInt::new(((1u128 << fields.len() - 1).to_string(), Span::call_site());
quote!((#field_init_bits as #builder_bit_set_ty))
```

wincode-derive/src/schema_read.rs
```diff
@@ -225,27 +230,249 @@ fn impl_struct_extensions(args: &SchemaArgs) -> Result<TokenStream> {
...
+                }
+
+                /// Forget the builder, disabling the drop logic.
+                ///
+                /// # Safety
+                ///
+                /// Assuming the source `MaybeUninit<T>` when the content is not yet fully initialized causes undefined behavior.
+                /// It is up to the caller to guarantee that the `MaybeUninit<T>` really is in an initialized state.
+                #[inline]
+                #vis const unsafe fn assume_init_forget(self) {
```
kskalski: Not sure I like this name either... since it doesn't actually run the `assume_init*` on underlying `MaybeUninit`, which the name might suggest.
Maybe it should be something like `finish_init_forget` or `init_done_forget` ?

We could make it return the underlying `&mut MaybeUnnit` and call it `into_inner_done`, which might be convenient in some cases (this way the other `into_assume_init_mut` would basically be equivalent to `into_inner_done().assume_init_mut()`), but also be a bit confusing if user just calls it ignoring the result.

wincode-derive/src/schema_read.rs
```diff
@@ -225,27 +230,249 @@ fn impl_struct_extensions(args: &SchemaArgs) -> Result<TokenStream> {
...
+                // SAFETY:
+                // - `self.inner` is a valid reference to a `MaybeUninit<#builder_dst>`.
+                // - We return the field as `&mut MaybeUninit<#target>`, so
+                //   the field is never exposed as initialized.
+                unsafe { &mut *(&raw mut (*self.inner.as_mut_ptr()).#ident).cast() }
+            }
+
+            /// Write a value to the maybe uninitialized field.
+            #[inline]
+            #vis const fn #write_uninit_field_ident(&mut self, val: #target) -> &mut Self {
```
kskalski, line 409: I think the choice of `read_field_name` and `write_field_name` names is a bit confusing here (though I guess it's already inherited from previous design of `struct_extensions`), since `read` operates on `Reader`, while `write` doesn't have anything to do with wincode io `Writer`.

I would definitely prefer to change `write` into something like `store_field_name` or `init_field_name`, which would also convey the message that field is not manually initialized

wincode-derive/src/schema_read.rs
cpubot: It'll be constant folded by the compiler https://rust.godbolt.org/z/qaPE5Krbj

wincode-derive/src/schema_read.rs
cpubot: I think `write` is fine, it matches the semantics of `MaybeUninit::write`

wincode-derive/src/schema_read.rs
cpubot: We could technically just call this function `forget` and remove `unsafe` as a simple convenience over `mem::forget`. The user still has to `assume_init_mut` on the original reference or `assume_init` on the original `MaybeUninit<T>` to get access to `T` -- the contract of both of those methods requires that they ensure the underlying `T` is fully initialized.

I would also be fine with just `unsafe finish`. Finish implies that we're done with the builder, and the fact that it's `unsafe` requires the user to understand the safety assumptions.

wincode-derive/src/schema_read.rs
kskalski: ok, if you think this makes the library code simpler, though with explicit calculation we could get rid of the branch with `quote!(#builder_bit_set_ty::MAX)` and just have one code always calculating the constant

wincode-derive/src/schema_read.rs
kskalski: That's a fair argument, there are some issues with that though - `MaybeUninit` also has `_read` method, which means something completely different than `read` generated for builder.
I think because in context of `wincode` functions doing `read` and `write` have their specific meaning, we should try to avoid naming things with different purpose that way.

```rust
            builder
                .read_a(reader)?
                .read_b(reader)?
                .write_c(test.c);
```

is basically an instant head-scratcher... are we reading a few things into the builder and then writing out the builder contents to Writer stored in `test.c` ?

wincode-derive/src/schema_read.rs
kskalski: `finish` seems fine, not sure it really needs `unsafe`, since it doesn't do anything unsafe by itself and the user should be aware of need to initialize all parts of memory when starting to use the builder in the first place. It's also a call that always needs to be made unless `into_assume_init_mut` is used, so forcing to add another `unsafe` block seems off.

 I would be in fact more concerned with how to ensure user do not forget to call `finished`.

alessandrod: +1 for finish and no unsafe.

I don't think we need to ensure it's called?

cpubot: Cool, I'll call it `finish`. I like that

wincode-derive/src/schema_read.rs
alessandrod: I've thought about this too in the past and I think I've discussed it with zach.
I think write should definitely be write, it's read I'm not sure about.

That said this PR is fixing a concrete bug, and I think we should land it. We
can improve the naming later imo, this is 0.2.2 we have a long way to 1.0

cpubot: I don't think it's confusing if you look at the type signature or read the function documentation. Would you prefer `read_a_from` or `write_from_reader`?

alessandrod (dismissed):
I'm approving as this fixes a bug. I do think the naming can probably be improved, but I don't have good proposals at the moment, don't think that calling write store is right, and just generally think we can improve the naming as we keep iterating

wincode-derive/src/schema_read.rs
cpubot: > I don't think we need to ensure it's called?

We actually do -- otherwise the drop logic will execute, which will call drop on all the fields. UB if the user then calls `assume_init` or any of the `assume_init*` variants

alessandrod (dismissed):

wincode-derive/src/schema_read.rs
kskalski: personally I would avoid the names that have overloaded meaning around the same API / snippets of code and in those cases would opt for `init_a` and `init_a_from_reader`.

kskalski (dismissed):
Ok, we should be open to naming adjustments in the future then. Otherwise LGTM

PR #43, opened 2025-12-03
Ensure HashMaps are preallocated in read

cpubot:
`HashMap`s are currently not preallocated in deserialization, resulting in suboptimal performance.

The issue is subtle:
https://github.com/anza-xyz/wincode/blob/8499c0092428c5efb2e7e569d1e778079ad0e6a0/wincode/src/schema/impls.rs#L1006-L1012

Normally, the `FromIter` implementation of `HashMap` would preallocate capacity based on the source iterator's `size_hint`. The problem here is that we're actually collecting into a `Result<HashMap<_, _>, _>`, and `Result`'s `FromIterator` prevents this from happening.

This PR adjusts the implementation to avoid `Result`'s `FromIterator` and preallocates and inserts explicitly.

Results:
image

Performance of the other types used in the macro are unaffected, as they don't support preallocation.

### DIFF
--- a/wincode/src/schema/impls.rs
+++ b/wincode/src/schema/impls.rs
@@ -1003,7 +1003,6 @@ macro_rules! impl_seq {
                     // SAFETY: `$key::TYPE_META` and `$value::TYPE_META` specify static sizes, so `len` reads of `($key::Dst, $value::Dst)`
                     // will consume `(key_size + value_size) * len` bytes, fully consuming the trusted window.
                     let reader = &mut unsafe { reader.as_trusted_for((key_size + value_size) * len) }?;
-                    #[allow(clippy::redundant_closure)]
                     let mut map = $with_capacity(len);
                     for _ in 0..len {
                         let k = $key::get(reader)?;
@@ -1012,7 +1011,6 @@ macro_rules! impl_seq {
                     }
                     map
                 } else {
-                    #[allow(clippy::redundant_closure)]
                     let mut map = $with_capacity(len);
                     for _ in 0..len {
                         let k = $key::get(reader)?;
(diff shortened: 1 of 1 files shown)


wincode/src/schema/impls.rs
```diff
@@ -1081,11 +1089,11 @@ macro_rules! impl_seq {
     };
 }
 
-impl_seq! { "alloc", BTreeMap<K: Ord, V> }
-impl_seq! { "std", HashMap<K: Hash | Eq, V> }
-impl_seq! { "alloc", BTreeSet<K: Ord> }
-impl_seq! { "std", HashSet<K: Hash | Eq> }
-impl_seq! { "alloc", LinkedList<K:> }
+impl_seq! { "alloc", BTreeMap<K: Ord, V>, |_| BTreeMap::new() }
+impl_seq! { "std", HashMap<K: Hash | Eq, V>, |len| HashMap::with_capacity(len) }
```
kskalski: I guess it could be passed as a function instead of using lambda:
`impl_seq! { "std", HashMap<K: Hash | Eq, V>, HashMap::with_capacity }`

kskalski: and maybe this would fix the lints that you had to allow, does it?

cpubot: Wouldn't work because `BTreeMap::new`, for example, doesn't take a parameter

cpubot: Actually if we keep the closure for `BtreeMap::new`, but omit for `HashMap::with_capacity` I think it should work

alessandrod (approved):

PR #46, opened 2025-12-04
impl SchemaRead/SchemaWrite for ShortU16

cpubot:
There is a desire in agave to use `ShortU16` directly in structs (https://github.com/anza-xyz/agave/pull/7694).

Right now we only have a `SeqLen` implementation for `ShortU16`, so it can only be used in container length type-parameters. But there's no reason we can't give it a `SchemaRead` and `SchemaWrite` implementation. We already have all the logic to support this, so it is trivial to provide. This will allow it to be embedded into structs as a member.

Additionally, rather than wincode providing a separate `ShortU16Len` ZST, we can just make it a type alias to `ShortU16`. This allows for a single import of `ShortU16` from `solana-short-vec` that can be used in both structs and container length type-parameters, while not breaking anyone downstream.

### DIFF
--- a/wincode/src/len.rs
+++ b/wincode/src/len.rs
@@ -71,9 +71,52 @@ pub mod short_vec {
             mem::{transmute, MaybeUninit},
             ptr,
         },
-        solana_short_vec::decode_shortu16_len,
+        solana_short_vec::{decode_shortu16_len, ShortU16},
     };
-    pub struct ShortU16Len;
+
+    impl<'de> SchemaRead<'de> for ShortU16 {
+        type Dst = Self;
+
+        fn read(reader: &mut impl Reader<'de>, dst: &mut MaybeUninit<Self::Dst>) -> ReadResult<()> {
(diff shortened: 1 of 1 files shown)


alessandrod (approved):

t-nelson:
this is fine, but generally we should be removing this abomination rather than making it easier to use

alexpyattaev:
> this is fine, but generally we should be removing this abomination rather than making it easier to use

Maybe we can drop the max length to 2 bytes? (it would be able to store up to 32768 which is plenty for our needs)

alexpyattaev:
> > Maybe we can drop the max length to 2 bytes? (it would be able to store up to 32768 which is plenty for our needs)
> 
> The dynamic nature of it makes it problematic performance-wise. Ideally we'd just use a u16 if protocol admits.

In case of gossip and repair we can expect plenty of messages with very small length (<100 bytes). Votor has only >127 bytes messages afaict. But we could go with u16 if there is a massive performance win.

t-nelson:
we could also just merge this and open an issue on the monorepo for the other stuff

PR #47, opened 2025-12-05
Implement faster, unrolled ShortU16 decoder

cpubot:
The `solana-short-vec` `ShortU16` implementation incurs unnecessary overhead from:
- Value mask before checking continuation bit (wasted instructions on theoretically hot 1 byte path)
- Loop with dynamic i * 7 shift 
- Defensive and technically unnecessary saturating_* / checked_* ops to guard against theoretical loop edge cases
- All the serde badness when paired with bincode

The compiler actually does a decent job optimizing this, but it can be improved.
https://rust.godbolt.org/z/rojK31Wf4

This implementation:
- Fully unrolled with constant shifts (7, 14) 
- No unnecessary widening or defensive ops — unrolled structure makes overflow impossible by construction
- `const fn`
- no serde obviously

https://rust.godbolt.org/z/eYzq1G8oT

results
```
ShortU16/solana_short_vec:decode_shortu16_len/127
                        time:   [8.1727 ns 8.1882 ns 8.2045 ns]
                        thrpt:  [116.24 MiB/s 116.47 MiB/s 116.69 MiB/s]
ShortU16/wincode:decode_short_u16/127
                        time:   [1.4131 ns 1.4156 ns 1.4179 ns]
                        thrpt:  [672.57 MiB/s 673.71 MiB/s 674.90 MiB/s]
```
```
ShortU16/solana_short_vec:decode_shortu16_len/16383
                        time:   [8.0641 ns 8.0804 ns 8.0978 ns]
                        thrpt:  [235.54 MiB/s 236.05 MiB/s 236.52 MiB/s]
ShortU16/wincode:decode_short_u16/16383
                        time:   [1.5453 ns 1.5522 ns 1.5588 ns]
                        thrpt:  [1.1950 GiB/s 1.2000 GiB/s 1.2054 GiB/s]
```
```
ShortU16/solana_short_vec:decode_shortu16_len/65535
                        time:   [8.4347 ns 8.4469 ns 8.4589 ns]
                        thrpt:  [338.23 MiB/s 338.71 MiB/s 339.20 MiB/s]
ShortU16/wincode:decode_short_u16/65535
                        time:   [1.7430 ns 1.7457 ns 1.7484 ns]
                        thrpt:  [1.5980 GiB/s 1.6005 GiB/s 1.6030 GiB/s]
```

results are obviously more dramatic without `black_box`ing since the compiler can constant fold e.g.,
```
ShortU16/solana_short_vec:decode_shortu16_len/65535
                        time:   [8.1254 ns 8.1398 ns 8.1548 ns]
                        thrpt:  [350.84 MiB/s 351.48 MiB/s 352.11 MiB/s]
                 change:
                        time:   [-3.9026% -3.6384% -3.3188%] (p = 0.00 < 0.05)
                        thrpt:  [+3.4328% +3.7757% +4.0611%]
                        Performance has improved.
ShortU16/wincode:decode_short_u16/65535
                        time:   [599.87 ps 600.92 ps 602.20 ps]
                        thrpt:  [4.6396 GiB/s 4.6495 GiB/s 4.6576 GiB/s]
                 change:
                        time:   [-65.392% -65.287% -65.146%] (p = 0.00 < 0.05)
                        thrpt:  [+186.91% +188.07% +188.95%]
                        Performance has improved.
```

### DIFF
--- a/wincode/src/len.rs
+++ b/wincode/src/len.rs
@@ -241,7 +241,7 @@ pub mod short_vec {
         if b2 == 0 {
             return Err(non_canonical_err());
         }
-        if b2 > 3 || b2 >= 0x80 {
+        if b2 > 3 {
             return Err(overflow_err());
         }
         let val = ((b0 & 0x7f) as u16) | (((b1 & 0x7f) as u16) << 7) | ((b2 as u16) << 14);


t-nelson:
we need to be very careful with review here. this type has been responsible for at least two bounty payouts

wincode/src/len.rs
```diff
@@ -167,14 +159,108 @@ pub mod short_vec {
...
+
+        // Byte 2
+        if bytes.len() < 3 {
+            return Err(incomplete_err());
+        }
+        let b2 = bytes[2];
+        if b2 == 0 {
+            return Err(non_canonical_err());
+        }
+        if b2 > 3 || b2 >= 0x80 {
```
kskalski: looks like there is redundant comparison here, 
`b2 > 3  => b2 >= 0x80`, right?

cpubot: Whoops, yep you're right

wincode/src/len.rs
```diff
@@ -167,14 +159,108 @@ pub mod short_vec {
...
+    /// ```
+    /// use wincode::len::decode_short_u16;
+    ///
+    /// let bytes = [0x80, 0x80, 0x01];
+    /// let (len, read) = decode_short_u16(&bytes).unwrap();
+    /// assert_eq!(len, 16384);
+    /// assert_eq!(read, 3);
+    /// ```
+    #[inline]
+    pub const fn decode_short_u16(bytes: &[u8]) -> ReadResult<(u16, usize)> {
```
kskalski, line 196: the assembly generated in the compiler explorer from that function itself is not that good, copying things around the stack to create `Result`, but seems like with inlining here and in `decode_short_u16_from_reader` compiler optimizes out lots of those operations (creating a large result object and getting values out of it)

wincode/src/len.rs
```diff
@@ -167,14 +159,108 @@ pub mod short_vec {
...
+
+        // Byte 0
+        if bytes.is_empty() {
+            return Err(incomplete_err());
+        }
+        let b0 = bytes[0];
+        if b0 < 0x80 {
+            return Ok((b0 as u16, 1));
+        }
+
```
kskalski, line 222: probably won't matter for performance after all the optimizatons compiler does, but for readability we could define
`let b0 = b0 & 0x7f) as u16;`
here and simplify / deduplicate uses below

kskalski (commented):
LGTM
The added tests give us proper coverage, at least assuming that reference implementation in `solana_short_vec::decode_shortu16_len` handles positive and negative cases properly.

t-nelson (approved):
finally found time to refamiliarize myself with the serde decoder (fml). this lgtm

PR #48, opened 2025-12-06
Fix struct_extensions projections on mapped and container types

cpubot:
`struct_extensions` currently uses `field.target_resolved()` to determine member projection types. This is incorrect for cases where a struct member is annotated with a `container` type, as `target_resolved` will return the fully inferred _container_ type, rather than the underlying type.

For example
```rs
struct A {
    #[wincode(with = "containers::Pod<_>")]
    signature: [u8; u32]
}
```

`target_resolved()` returns `containers::Pod<[u8; 32]>`, whereas we actually want the `[u8; 32]` directly.

> Note that `target_resolved` is still the correct type to use for `read_<field>`, as `read` should use the type's `read` implementation directly (and `read_<field>` implementation is unchanged). But for `write_<field>` or `uninit_<field>_mut`, we need to use the destination type (`Dst`) since they circumvent the `SchemaRead::read` impl.

The fix here is simple -- just use `<field.ty as SchemaRead>::Dst` rather than `field.target_resolved()`. This works for mapped types, container types, and plain built-in types.

Example mapped type
```rs
struct A {
    signature: [u8; u32]
}

#[wincode(from = "A", struct_extensions)]
struct AMapped {
    signature: containers::Pod<[u8; u32]>,
}
```

Added tests that will fail to compile on current `master`.

### DIFF
--- a/wincode/src/schema/mod.rs
+++ b/wincode/src/schema/mod.rs
@@ -922,8 +922,11 @@ mod tests {
                 .write_b(test.b)
                 .write_c(test.c);
             prop_assert!(builder.is_init());
-            let init = unsafe { builder.into_assume_init_mut() };
-            prop_assert_eq!(&test, init);
+            let init_mut = unsafe { builder.into_assume_init_mut() };
+            prop_assert_eq!(&test, init_mut);
+            // Ensure `uninit` is marked initialized so fields are dropped.
+            let init = unsafe { uninit.assume_init() };
+            prop_assert_eq!(test, init);
         });
     }
 
--- a/wincode-derive/src/schema_read.rs
+++ b/wincode-derive/src/schema_read.rs
@@ -356,10 +356,10 @@ fn impl_struct_extensions(args: &SchemaArgs, crate_name: &Path) -> Result<TokenS
                 /// Calling this when the content is not yet fully initialized causes undefined behavior: it is up to the caller
                 /// to guarantee that the `MaybeUninit<T>` really is in an initialized state.
                 #[inline]
-                #vis const unsafe fn into_assume_init_mut(mut self) -> &'_wincode_inner mut #builder_dst {
+                #vis unsafe fn into_assume_init_mut(mut self) -> &'_wincode_inner mut #builder_dst {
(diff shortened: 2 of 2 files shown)


wincode/src/schema/mod.rs
```diff
@@ -902,6 +902,64 @@ mod tests {
...
+        proptest!(proptest_cfg(), |(test: Test)| {
+            let mut uninit = MaybeUninit::<Test>::uninit();
+            let mut builder = TestUninitBuilder::from_maybe_uninit_mut(&mut uninit);
+            builder
+                .write_a(test.a.clone())
+                .write_b(test.b)
+                .write_c(test.c);
+            prop_assert!(builder.is_init());
+            builder.finish();
+            let init = unsafe { uninit.assume_init() };
```
kskalski: could use int_uninit_mut() here

wincode/src/schema/mod.rs
```diff
@@ -902,6 +902,64 @@ mod tests {
         });
     }
 
+    #[test]
+    fn test_struct_extensions_with_container() {
+        #[derive(SchemaWrite, SchemaRead, Debug, PartialEq, Eq, proptest_derive::Arbitrary)]
+        #[wincode(internal, struct_extensions)]
+        struct Test {
```
kskalski, line 909: since there is a non-trivial `SchemaRead` lifetime handling, could you add a test that works on struct with lifetimes, say
```rust
struct TestRef<'a> {
    a: &'a [u8],
    b: Option<'a u8>
}
```

it should be possible to initialize it with a (locally scoped) slice, right?

cpubot: Sure, 5ed3c6bcd14fbd54eafcea706ac0c8cbc5cb5488

wincode/src/schema/mod.rs
cpubot: e38f188876575baf2f19a56c8c4ef4c71800a7f9

kskalski (approved):

kskalski:
Interestingly, tests show UB now?

PR #53, opened 2025-12-12
RFC: Impl BufReader for std::io::Read

cpubot:
wincode currently doesn't provide direct support for interfacing with file-io (e.g., via `Read`) due to its specific trait semantics. In particular, 
- wincode has a concept of "trusted" readers that can elide bounds checking
- in general prefers a `BufRead` style reader due the fact that `SchemaRead` implementations need to deal with variable sized type encodings (e.g., variable length encodings like `ShortU16`). For example, an implementation may need to inspect the next 3 bytes, but only use a subset of those bytes. Of course those implementations could get by with incremental reading as bytes are inspected, but that would be terribly inefficient without buffering, and so it's preferable for `SchemaRead` implementations in particular to deal with a `BufRead` style interface.

We can provide our own `BufReader` over any `std::io::Read` that implements the appropriate wincode semantics. The implementation is fairly straightforward until we need to deal with `as_trusted_for` on the `Reader` trait. Put simply, calling `as_trusted_for` means "I know for a fact that I will need `n` bytes for all subsequent reads on the returned `Reader`". This is trivial and obvious implementation-wise when the underlying source is in-memory, but the "right" way to handle this when dealing with what could be file or network IO is more nuanced.

Perhaps the most straightforward solution is to simply attempt to fill the `BufReader`'s buffer up to `capacity.min(n_bytes_requested)`, and then proceed as usual. This is great for cases where `n_bytes_requested` is less than `capacity`. It can become suboptimal however if `n_bytes_requested` is significantly larger than the buffer capacity. In particular, we incur additional syscall overhead for subsequent reads past capacity that could otherwise be avoided if the buffer were larger. If we were to instead _grow_ the buffer to support the requested number of bytes, we could conceivably read most if not all required bytes in a single syscall.

So, to summarize the above options for dealing with `as_trusted_for`:
1. Fill buffer capacity as much as possible without growth, `capacity.min(n_bytes_requested)`, then proceed as usual.
2. Grow the buffer to accommodate all future reads on the returned trusted reader and simply return a `TrustedSliceWriter` over those bytes.

Option 1:
Simpler conceptually, more predictable memory usage (no growth), and thus probably closer semantically to what users expect from `std::io::BufReader`. We can't really provide a "Trusted" variant of the `BufReader` in this case (in the way that we can with in-memory sources), because we still have to deal with bounds checking and potential syscalls.

Option 2: 
Less predictable memory usage (may grow), but will reduce syscall overhead for large trusted windows. Additionally benefits from the performance of `TrustedSliceWriter` for all reads in that trusted window (no bounds checking + no syscalls -- high likelihood of vectorized reads)

I'd like to have a discussion on which route is better for wincode, so I'm providing both implementations in two draft PRs.

This PR includes an implementation of Option 2 (growable `BufReader`).

See #54 for Option 1 implementation.

Note the code in these implementations shouldn't be considered "final" per se, but working proof of concepts that are close to finalization.

### DIFF
--- a/wincode/src/io/std_io.rs
+++ b/wincode/src/io/std_io.rs
@@ -1,0 +1,539 @@
+use {
+    crate::io::{read_size_limit, ReadError, ReadResult, Reader, TrustedSliceReader},
+    core::{
+        mem::{transmute, MaybeUninit},
+        ops::Range,
+        ptr,
+        slice::{from_raw_parts, from_raw_parts_mut},
+    },
+    std::io::{ErrorKind, Read},
+};
+
+/// Add buffered reading to any [`Read`] instance.
+///
+/// It can be excessively inefficient to work directly with a [`Read`] instance.
+/// For example, every call to read on [`TcpStream`](std::net::TcpStream) results in a system call.
+/// A [`BufReader<R>`] performs large, infrequent reads on the underlying [`Read`] and
+/// maintains an in-memory buffer of the results.
+///
+/// [`BufReader<R>`] can improve the speed of programs that make small and repeated
+/// read calls to the same file or network socket. It does not help when reading very
+/// large amounts at once, or reading just one or a few times. It also provides no advantage
+/// when reading from a source that is already in memory, like a [`Vec<u8>`].
+pub struct BufReader<R: ?Sized> {
+    buf: Vec<u8>,
+    filled: Range<usize>,
+    inner: R,
+}
+
+const DEFAULT_BUF_SIZE: usize = 8 * 1024;
+
+impl<R> BufReader<R> {
+    /// Create a new [`BufReader<R>`] with the default buffer capacity.
+    ///
+    /// The default buffer capacity is currently 8KiB.
+    pub fn new(inner: R) -> Self {
+        Self {
+            buf: Vec::with_capacity(DEFAULT_BUF_SIZE),
+            filled: 0..0,
+            inner,
+        }
+    }
+
+    /// Create a new [`BufReader<R>`] with the specified capacity.
+    pub fn with_capacity(inner: R, capacity: usize) -> Self {
+        Self {
+            buf: Vec::with_capacity(capacity),
+            filled: 0..0,
+            inner,
+        }
+    }
+
+    /// Consume the [`BufReader<R>`] and return the underlying reader.
+    ///
+    /// Note that any leftover data in the buffer will be lost.
+    pub fn into_inner(self) -> R {
+        self.inner
+    }
+}
+
+/// Fill the buffer with up to `n_bytes` from the reader.
+///
+/// Note this implementation differs from the semantics of [`std::io::BufRead`]
+/// in that wincode [`Reader`]s take an `n_bytes` argument.
+/// Importantly, implementations should try to read at least `n_bytes`
+/// bytes, retrying until either `n_bytes` are read or EOF is hit.
+///
+/// This function does NOT grow the buffer. If `n_bytes > capacity`, it returns an error.
+/// Use [`maybe_grow_and_fill_buf`] if buffer growth is needed.
+fn fill_buf<'a, R: Read + ?Sized>(
+    r: &mut R,
+    buf: &'a mut Vec<u8>,
+    filled: &mut Range<usize>,
+    n_bytes: usize,
+) -> ReadResult<&'a [u8]> {
+    // Number of bytes already buffered.
+    let buffered_len = filled.len();
+    // We already have sufficient bytes in the buffer.
+    if buffered_len >= n_bytes {
+        // SAFETY: `filled` always points to an initialized portion of the buffer.
+        return Ok(unsafe { from_raw_parts(buf.as_ptr().add(filled.start), n_bytes) });
+    }
+
+    let capacity = buf.capacity();
+
+    // Error if request exceeds buffer capacity (this function doesn't grow).
+    if n_bytes > capacity {
+        #[cold]
+        fn out_of_memory() -> ReadError {
+            ReadError::Io(ErrorKind::OutOfMemory.into())
+        }
+        return Err(out_of_memory());
+    }
+
+    // SAFETY: we check that `n_bytes > capacity` above, so this will not underflow.
+    let needed = unsafe { n_bytes.unchecked_sub(buffered_len) };
+    // SAFETY: we maintain the invariant that `filled.end` is always less than `capacity`.
+    let edge_capacity = unsafe { capacity.unchecked_sub(filled.end) };
+
+    // User requested more bytes than we have space for relative to filled.end.
+    // Compact the buffer by shifting existing bytes to the beginning.
+    if needed > edge_capacity {
+        let base = buf.as_mut_ptr();
+        // SAFETY: `filled` always points to an initialized portion of the buffer.
+        let src = unsafe { base.add(filled.start) };
+        let dst = base;
+        // Use `copy_nonoverlapping` if we can, otherwise use `copy`.
+        if filled.start >= buffered_len {
+            // SAFETY:
+            // - `filled` always points to an initialized portion of the buffer.
+            // - we checked that `filled.start >= len`, src and dst don't overlap.
+            unsafe { ptr::copy_nonoverlapping(src, dst, buffered_len) };
+        } else {
+            // SAFETY:
+            // - `filled` always points to an initialized portion of the buffer.
+            // - we checked that `filled.start < len`, src and dst overlap.
+            unsafe { ptr::copy(src, dst, buffered_len) };
+        }
+
+        *filled = 0..buffered_len;
+    }
+
+    // SAFETY: we checked that the buffer has sufficient capacity for `n_bytes`.
+    unsafe { fill_buf_unchecked(r, buf, filled, n_bytes) }
+}
+
+/// Fill the buffer with up to `n_bytes` from the reader without any
+/// fast path checks (e.g., don't return early if we have enough bytes in the buffer)
+/// or whether the buffer has capacity for `n_bytes`.
+///
+/// # Safety
+/// - Caller guarantees the buffer has sufficient edge capacity (i.e., `buf.capacity() - filled.end`)
+///   for `n_bytes`.
+#[expect(clippy::arithmetic_side_effects)]
+unsafe fn fill_buf_unchecked<'a, R: Read + ?Sized>(
+    r: &mut R,
+    buf: &'a mut Vec<u8>,
+    filled: &mut Range<usize>,
+    n_bytes: usize,
+) -> ReadResult<&'a [u8]> {
+    let buffered_len = filled.len();
+    // Caller guarantees n_bytes is greater than the number of bytes already buffered.
+    let needed = unsafe { n_bytes.unchecked_sub(buffered_len) };
+    // SAFETY: we maintain the invariant that `filled.end` is always less than `capacity`.
+    let edge_capacity = unsafe { buf.capacity().unchecked_sub(filled.end) };
+
+    let mut read = 0;
+    // SAFETY:
+    // - `filled.end` is always less than `buf.capacity()` by invariant.
+    let mut dst = unsafe {
+        from_raw_parts_mut(
+            buf.as_mut_ptr().cast::<MaybeUninit<u8>>().add(filled.end),
+            edge_capacity,
+        )
+    };
+    while read < needed {
+        // SAFETY: `read` only writes to uninitialized bytes.
+        match r.read(unsafe { transmute::<&mut [MaybeUninit<u8>], &mut [u8]>(dst) }) {
+            Ok(0) => break,
+            Ok(n) => {
+                read += n;
+                // SAFETY: `n` bytes were written to `dst`, so `dst` is advanced by `n` bytes.
+                dst = unsafe { dst.get_unchecked_mut(n..) };
+            }
+            Err(ref e) if e.kind() == ErrorKind::Interrupted => {}
+            Err(e) => return Err(e.into()),
+        }
+    }
+
+    filled.end += read;
+
+    // SAFETY: `filled` always points to an initialized portion of the buffer.
+    let out = unsafe { from_raw_parts(buf.as_ptr().add(filled.start), filled.len().min(n_bytes)) };
+    Ok(out)
+}
+
+/// Like [`fill_buf`], but grows the buffer if `n_bytes > capacity`.
+///
+/// Used by `as_trusted_for` where we need to guarantee a contiguous window of N bytes.
+fn maybe_grow_and_fill_buf<'a, R: Read + ?Sized>(
+    r: &mut R,
+    buf: &'a mut Vec<u8>,
+    filled: &mut Range<usize>,
+    n_bytes: usize,
+) -> ReadResult<&'a [u8]> {
+    let buffered_len = filled.len();
+    if buffered_len >= n_bytes {
+        // SAFETY: `filled` always points to an initialized portion of the buffer.
+        return Ok(unsafe { from_raw_parts(buf.as_ptr().add(filled.start), n_bytes) });
+    }
+
+    if n_bytes > buf.capacity() {
+        // Need to reallocate - copy directly to new buffer rather than shifting (copying) and reallocating.
+        let mut new_buf = Vec::with_capacity(n_bytes.next_power_of_two());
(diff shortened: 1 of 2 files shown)


kskalski:
I think it would be useful to first start with implementation for `BufRead` (this way we could also plug in any other implementation of `BufRead`, not necessarily rely on `BufReader` or implement it specifically).

In agave similar problem was handled by adding trait and impl that has overflow buffer (https://github.com/anza-xyz/agave/blob/4e031c8025f0c14ecc2b8652f1a75cf5051e00f3/fs/src/buffered_reader.rs#L87), my impression is that here we could use similar approach, but I will also read some of the code here and see if it or something else look better.

kskalski:
We could use traits for increasingly efficient implementations though:
* make best-effort (i.e. not use trusted reader when obtained buffer is too small) implementation for `BufRead`  - I think this could be useful if user already have some buffering and little control over it
* define a trait deriving from `BufRead` that adds extra functionality of obtaining at least `n_bytes` buffer, like `fill_buf_preferred_len`
  * this way user can plug-in other implementations (i.e. in agave we use io-uring reader that can use kernel registered buffers with read-ahead)
  * that kind of API will allow implementation to decide when it will grow and when not (`fill_buf_preferred_len` can simply provide smaller buf and we fallback to slower impl - though we could consider having API that distinguish returning short buf from EOF)
  * `wincode` can provide `BufReader` with this functionality that we can advertise to use instead of default `BufReader`
* the way to use wincode for `io::Read` should be through some of the above readers
* `wincode` buf reader could include parameters for initial buffer size and max growth

wincode/src/io/std_io.rs
```diff
@@ -0,0 +1,539 @@
...
+    r: &mut R,
+    buf: &'a mut Vec<u8>,
+    filled: &mut Range<usize>,
+    n_bytes: usize,
+) -> ReadResult<&'a [u8]> {
+    let buffered_len = filled.len();
+    // Caller guarantees n_bytes is greater than the number of bytes already buffered.
+    let needed = unsafe { n_bytes.unchecked_sub(buffered_len) };
+    // SAFETY: we maintain the invariant that `filled.end` is always less than `capacity`.
+    let edge_capacity = unsafe { buf.capacity().unchecked_sub(filled.end) };
```
kskalski, line 144: does `unchecked_sub` generate more efficient code than `wrapping_sub`?

kskalski:
> Additionally, I don't think `std::io::BufRead` is necessarily suitable for providing an implementation of `Reader`. In particular, `fill_exact(n_bytes)` and `fill_buf(n_bytes)` aren't implementable with `BufRead`'s 

Ok, that's right, I forgot that `Reader` trait requires buffers of exact size, not just for trusted reader.
Agreed that `Reader` should be directly implemented by other buffered readers, since it requires communicating the need to obtain larger buffer and benefits from direct mutable access to the buffer.

> > * `wincode` buf reader could include parameters for initial buffer size and max growth
> 
> This might be the best route -- giving users a choice as to whether they want a growable buffer. We could provide two implementations (e.g., a `BufReader` and `GrowableBufReader`) that the user could pick. Or, parameterize `BufReader` like `struct BufReader<const GROWABLE: bool>` over which we could provide two implementations with differing `Trusted` associated types based on whether its growable.

Not sure how trusted readers relate to the growth capacity, but even for regular `Reader::fill_buf` we should probably have the variants that are able to fulfill growth (and maybe this should have thresholds) or signal error / fallback to less efficient operation.
The most flexible API would allow specifying initial buffer capacity and growth limit. Do you think a constant "growable" boolean specialization would be noticeably more efficient than checking growth limit?

I suppose we have cases where large number of bytes are requested based on runtime sizes (e.g. deserializing to `&[u8]` need to point to a whole slice, so all of the bytes need to be available), but sometimes those are best effort? What are the cases here
* minimum required buffer size to deserialize necessary types --> user should probably just set initial buffer capacity to at least that
* minimum required size to deserialize some rarely occuring cases -> need to grow to that size depending on input, otherwise error
* ideal buffer size to efficiently deserialize necessary types -> is reallocation or less efficient deserialization better

kskalski:
Coming back to this with some more comments as I'm also considering implementation of `Reader` for io-uring reader in agave:
* I think the `as_trusted_for` API is currently too demanding to work well for buffered  / chunked inputs - it has "optimize all or error" approach, but for many use-cases it seems more reasonable to have "optimize most or fallback to slow". Say you want to efficiently read and deserialize 1 billion-len `Vec<StaticStruct>`, it's nice to avoid the bound checks for `read` calls deserializing each individual struct, but you don't really need 1B*`size_of::<StaticStruct>` bytes of input immediately and fully available to fill the destination collection - you could as well do that in chunks of 1 million elements at a time, re-request a new trusted reader for next chunk of input, still a 1billion -> 1k bound checks reduction. Requiring callers to handle trusted readers below requested capacity is breaking and maybe sometimes a bit inconvenient to implement, but seems reasonable (implementations could also fallback to slow path and for properly sized in-mem input it would never happen).
* your no-grow implementation is close to what I imagined for `BufRead` default implementation - of course there is the issue of `fill_buf` that even if operating on large buffer can return small slice when read position is at the boundary. For me the best approach would be to introduce a trait in `wincode` that would add `fill_buf_hint_at_least` to hint/force the reader to defragment / up-fill the buffer. Maybe there isn't that much more in the `impl Reader` besides `fill_buf_n`, but there is still some unsafe / wincode specific code, trusted reader, copying slices, etc.

kskalski:
> > introduce a trait in `wincode` that would add `fill_buf_hint_at_least` to hint/force the reader to defragment / up-fill the buffer. Maybe there isn't that much more in the `impl Reader` besides `fill_buf_n`, but there is still some unsafe / wincode specific code, trusted reader, copying slices, etc.
> 
> If I'm understanding your concern correctly, I believe that the `fill_buf` implementation in the no-grow impl will handle the case you're describing. `fill_buf` in wincode does actually take an `n_bytes` argument, and the impl in the no-grow branch will defragment when `n_bytes` would push the buffer over the boundary (or return an OOM error if `n_bytes` exceeds capacity).

Right, the impl in the draft PR does that, I'm thinking of making things more useful by:
* making that function pluggable by exposing a trait that user can implement on their own `BufRead` implementations, so for example:
```rust
impl BufReadWithSizeHint for MyBufReader {
   fn fill_buf_hint_at_least(&mut self, desired_len: usize) -> io::Result<&[u8]> { .. }
}
..
let wincode_reader = wincode::io::IoReader::from_refillable_buf_read(MyBufReader::new())
```
could be used and all the rest of the implementation in this / the other PR would be put into `wincode::io::IoReader` that only uses functions of `BufRead + BufReadWithSizeHint`

* the current implementation of `fill_buf(n_bytes)` could also be exposed by wincode crate to ease implementation of `BufReadWithSizeHint`, I guess such function would take mutable buffer and `Read` to defragment + refill available buffer from "source" reader (obviously it should be a different reader than self ;-))

* I suppose we would still need to provide our own buf reader implementation, hopefully with the above two functionalities extracted, it would actually be cleaner

> On this point, assuming we go the no-grow route, we could achieve this by storing the total number of requested trusted bytes in the trusted reader, and all calls to `fill_buf` and related calls could maximally fill the buffer relative to that trusted total (decrementing with each fill). That way we do the minimum number of IO syscalls relative to the size of the buffer when trusted readers are used.

In this mode the trusted reader would make its own decision whether to call source reader vs just provide existing data? Checking all the relevant conditions might destroy the purpose of trusted reader that should avoid bound checks...

There is another thing to consider here - the implementations we discussed above will try to fulfil the desired len hint / `n_bytes` by defragmenting data within current buffer and adding extra - this isn't always the best thing to do, `as_trusted_for` is used when caller wants to make accessing next n bytes cheaper, but moving the bytes around the internal buffer might actually be more expensive than the bound checks being avoided... Maybe we need to distinguish cases of:
* caller *must* get `>= n_bytes` -> supposedly this is just for zero copy cases and not applicable to buffered reader, just that overall API must support this case
* caller would prefer to get `>= n_bytes` because it would allow them to do batched / vectorized operation on them in one go (e.g. single memcopy for all), but can handle incremental operation -> typically `std::io` APIs just give "as many as possible" bytes
* caller wants to suspend bounds checks for the next `n_bytes` - it doesn't necessarily mean all those bytes need to be immediately available (I suppose this is what you proposed in the comment above)

The breaking change is that 2. and 3. as opposed to 1. would now need to handle getting less than requested number of bytes. We wouldn't de-fragment buffer in those cases (or it could be implementation specific, e.g. if available bytes are <1/8 of the buffer capacity or < some fixed size, it does consider move + refill).

kskalski:
>     * I think the `as_trusted_for` API is currently too demanding to work well for buffered  / chunked inputs - it has "optimize all or error" approach, but for many use-cases it seems more reasonable to have "optimize most or fallback to slow". Say you want to efficiently read and deserialize 1 billion-len 

Ok, actually the APIs might be usable without much change, though maybe the cleanest way would be to make `as_trusted_for` return value more explicit about "OK / Request Too large / Error" distinction. For now I hacked a POC that (ab)uses the `io::ReadError` a bit:
https://github.com/anza-xyz/wincode/pull/185

Benchmark run suggests it won't harm us.

kskalski:
Let's list cases where getting `n_bytes` size `&[u8]` is:
1. required for arbitrary `n_bytes` -> it's impossible to fulfill with buffered reader without arbitrary re-allocs
2. required for user-known limit `n_max` -> it's possible to implement with `Reader(R: BufRead, vec(n_max))` by using the extra pre-allocated vec whenever `BufRead::fill_buf` returns fewer than `n_bytes`, or with custom buf reader that uses single `n_max` buffer and defragments / re-fills
3. best-effort, meaning getting `n_bytes` gets better perf, but slow-path exists -> we can implement directly over `BufRead` (or make it trivial to implement by user for their own reader)

If I'm not mistaken zero-copy (especially deserializing `&Struct`) requires 1. I'm not sure if 2. would work, which brings my side-question. Is support for zero-copy with buffered reader simply impossible due to how `Reader`'s lifetime / API is designed or it's just a matter of ensuring that requested part of the data is in buffer? I suspect it's the former, meaning `borrow_exact` lifetimes force impl to *always* have / own the relevant part of the memory, not only for the duration of single call / restricted piece of data.

If we exclude the zero-copy case (ideally we would hard-enforce buffered reader not to be used for it at all), I wonder if we can make things work with best-effort (3.). I looked at cases where `fill_buf` or `fill_buf_exact` is called and I think the reader can prevent those (i.e. re-implementing `fill_array`, `copy_into_slice_t` without requiring >= n_bytes, replacing some calls to `fill_buf(n)` with `fill_array` - there is a couple that are used with small `N` like int or char parsing). 
This just leaves `as_trusted_for` - and I think we can force callers to handle the inability to get trusted reader for `n_bytes` (as in my https://github.com/anza-xyz/wincode/pull/185 or by using other enum as you suggested).

I guess one important use-case I have in mind here that would work quite well with best-effort trusted reader is `Vec<Vec<StaticSizeStruct>>`, since each element of outer vector is arbitrary, but derived from len, size, handling each of them by trusted read occasionally falling back to slower impl (that iterates over inner vec elements doing bounds checks) would be almost at no perf loss.

Given above, maybe the `BufReadWithSizeHint` doesn't make sense (its purpose was to introduce a smaller trait that could be put on top of `BufRead` to "help" the implementer handle case 2. such that they can use single existing buffer and de-fragment / refill it on demand). I would say if 2. doesn't bring us much compared to 1. (i.e. both won't support zero-copy), then we can drop the idea.

kskalski:
Ok, I noticed now that `Reader'`s `fill_buf` parameter `n_bytes` caps the returned slice at max `n_bytes`, not min `n_bytes`, so actually it can be trivially implemented on top of `BufRead::fill_buf`. The only calls we need to worry about are `fill_buf_exact` and `as_trusted_reader`:
* the only call to `fill_buf_exact` outside of the other functions in `Reader` (whose default impl can be replaced for `BufRead` easily), is in `char` deserialization - I will send PR chaging it with `fill_array` or something.
* your implementation skips `borrow_*` functions, so they get the default "error for zero-copy" behavior - I assume this is how we make the reader unusable for zero-copy and we don't worry about that
* if we change the `as_trusted_for` contract such that callers need to fallback to slow impl, we can also implement it on top of `BufRead`

If I'm not missing anything else, we could then provide
```rust
pub struct BufReader<R: std::io::BufRead>(R);
impl<'a, R: std::io::BufRead> Reader<'a> for BufReader<R> { ..}
```

We could also have the impl that makes a bit more effort by having access to internal buffer (move mem and re-fill at times we identify as worth it), but having above either in code or listing functions to re-implement in the doc, would provide contract on how to implement the `Reader` in user's own buf reader impl.

kskalski:
> if we change the as_trusted_for contract such that callers need to fallback to slow impl, we can also implement it on top of BufRead

Alternatively if we don't want to do that, we could indeed make buffered reader return a "trusted" reader that cheats and still does bound checks, at least the one that will make it re-fill on demand. Trusted reader just fulfills `Reader` trait, so there is no other extra guarantee of having contiguous memory of requested size...

kskalski:
> I think `fill_buf`, `fill_exact`, etc should always obey the `strict` policy -- if we cannot ever satisfy `n_bytes`, hard error.

This will force adding extra buffering layer on top of many implementations that are already implementing `BufRead` (for example `bytes` crate provides (https://github.com/tokio-rs/bytes/blob/00cc5ff2bd861f7b62603a268ec477c3cead5f1e/src/buf/reader.rs#L74-L81) `BufRead` that might be chunked and it's likely someone would want to use `wincode` with `bytes`, which facilitates zero-copy networking). It's imaginable that some of the implementations could handle copying data around to ensure requested exact number of bytes is available to borrow as contiguous slice, but it's going to be difficult for external crates and might imply performance penalty.

The fundamental issue is that when you operate on chunked input, say [<100 bytes>, <100 bytes>], you proceed with reading, say 96 bytes are consumed, then user requests `fill_exact(8)`, getting slice of len=8 requires shuffling data around. This could be avoided if user only calls `copy_into_*`, since then the reader can trivially just go over 4 bytes from first chunk and 4 bytes of next chunk.

I think we could basically distinguish 3 layers of `Reader` API (whether they should actually be split into separate traits it a separate question):
* base API that doesn't assume input is a single contiguous piece of memory -> follow the standard `std::io` pattern of looping over incremental io operations that might only read / write part of the requested size / slice - it's then compatible with `BufRead` and allows using it without extra buffer layer (see my impl in https://github.com/anza-xyz/wincode/pull/196/)
* API that allows temporarily borrowing single slice of data of specified minimum size (`fill_exact`) -> is available when we operate on underlying large(r) slices only providing sub-view for them or we can / are willing to defragment memory to allow it; this might enable optimized operation for parts of the impl, since we can get rid of loop + len check and use vectorized optimizations are you mentioned
* API that allows "permanent" (tired to reader's `'a`) borrowing of sub-slices of data, the & zero-copy case (`borrow_exact`) -> similarly as above, but now only operating on pre-existing single slice is supported

The `as_trusted_for` brings a few more issues, but I suppose those are less critical as making above API distinction. Namely we need to either push responsibility to loop to user as in https://github.com/anza-xyz/wincode/pull/185 or use the approach from your https://github.com/anza-xyz/wincode/pull/54/ to make trusted reader check len / refill on each op. Separately in https://github.com/anza-xyz/wincode/pull/196/ I got into an issue with trusted reader "consuming" bytes, but then operating on them - in general it's not safe (or even possible) to call `consume` on `BufRead` while still holding / using the buffer returned from `fill_buf`), using drop guard isn't easy either, I ended up delaying the consume call, but ideally would like to avoid those checks.

In summary, I think a good course of action is to:
* make `Reader'`s default implementation not assume `fill_buf` will provide `== n_bytes` (which happens in `fill_exact`), all functions should use `copy_into_slice` to obtain safe behavior compatible with the doc comment from `fill_buf` saying "This is _not_ required to return exactly `n_bytes`, it is required to return _up to_ `n_bytes`."
* return error in `fill_exact` in default `Reader` -> similarly to `borrow_mut` this function shouldn't be assumed to be always available
* *do* implement `fill_exact` on all the slice readers along with optimized versions of functions using `fill_exact` to hint compiler about possible optimizations
* resolving issues with trusted reader is up for discussion, there are a couple of trade-offs (e.g. using enum also adds dispatch to each call)
* provide `Reader` implementation that would work on arbitrary `BufRead` without extra layer of buffering
* `std::io::Read` could be implemented by using above with standard `BufReader`, but of course we could provide an optimized alternative

PR #55, opened 2025-12-12
replace `wincode::serialize`  with `serialize_into` 

### LINKED ISSUE #52: Use `serialize_into` in benchmarks
Benchmarks were introduced in https://github.com/anza-xyz/wincode/pull/40.

`serialize` in particular always computes the serialized size of the type and heap allocates a `Vec<u8>` with appropriate capacity. For a better sense of the pure serialization performance, we should use `serialize_into` with a preallocated buffer. This will eliminate noise associated with the vector allocation and `serialized_size` calculation. It may also be worth having a separate `serialized_size` bench so we don't lose that information -- wincode does a lot to ensure `serialized_size` is more efficient than other solutions.

tanmay4l:
closes #52  cc @cpubot

### DIFF
--- a/wincode/benches/benchmarks.rs
+++ b/wincode/benches/benchmarks.rs
@@ -20,7 +20,7 @@ struct PodStruct {
     c: [u8; 8],
 }
 
-/// verification helper: ensures wincode output matches bincode 
+/// verification helper: ensures wincode output matches bincode
 fn verify_serialize_into<T>(data: &T) -> Vec<u8>
 where
     T: SchemaWrite<Src = T> + Serialize + ?Sized,


undali:
HI,
do we need to replace `bincode::serialize()` with `bincode::serialize_into()` to make the comparison fair?

tanmay4l:
> HI, do we need to replace `bincode::serialize()` with `bincode::serialize_into()` to make the comparison fair?
Screenshot 2025-12-12 at 10 22 54 PM

tanmay4l:
> Looks good @TanmayDhobale. Let's also do the following:
> 
> 1. Include `bincode::serialized_size` benchmarks
> 2. Update `criterion` to latest (`v0.8.1`) -- only change here is moving to `std::hint::black_box`
> 3. Add a `.cargo/config.toml` with:
> 
> ```toml
> [build]
> rustflags = ["-C", "target-cpu=native"]
> ```

Updated `criterion` to v0.8.1 and moved to `std::hint::black_box`

tanmay4l:
> Nice work!

Can I get Wincode merch ,  pls 😃

PR #56, opened 2025-12-13
schema: Implement `ZeroCopy` for all integer types

vadorovsky:
Given that all integer types provide `to_le_bytes` methods and can be transmuted, implement `ZeroCopy` for them. Test that a struct with all supported primitive types and explicit padding supports zero-copy.

### DIFF
--- a/wincode/src/schema/impls.rs
+++ b/wincode/src/schema/impls.rs
@@ -58,6 +58,9 @@ macro_rules! impl_int {
 
             const TYPE_META: TypeMeta = TypeMeta::Static {
                 size: size_of::<$type>(),
+                #[cfg(target_endian = "little")]
+                zero_copy: true,
+                #[cfg(not(target_endian = "little"))]
                 zero_copy: $zero_copy,
             };
 
@@ -77,6 +80,9 @@ macro_rules! impl_int {
 
             const TYPE_META: TypeMeta = TypeMeta::Static {
                 size: size_of::<$type>(),
+                #[cfg(target_endian = "little")]
+                zero_copy: true,
+                #[cfg(not(target_endian = "little"))]
                 zero_copy: $zero_copy,
             };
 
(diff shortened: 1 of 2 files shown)


vadorovsky:
@codex review

wincode/src/schema/impls.rs
```diff
@@ -155,14 +155,14 @@ unsafe impl ZeroCopy for i8 {}
...
-impl_int!(u128, zero_copy: false);
-impl_int!(i128, zero_copy: false);
+impl_int!(u16, zero_copy: true);
+impl_int!(i16, zero_copy: true);
+impl_int!(u32, zero_copy: true);
+impl_int!(i32, zero_copy: true);
+impl_int!(u64, zero_copy: true);
+impl_int!(i64, zero_copy: true);
+impl_int!(u128, zero_copy: true);
+impl_int!(i128, zero_copy: true);
```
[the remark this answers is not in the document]
vadorovsky: ~I guess that could be solved by using `to_ne_bytes`, probably worth a separate change.~ Nah, https://github.com/anza-xyz/wincode/pull/56#discussion_r2619399285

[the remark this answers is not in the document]
vadorovsky: Done

wincode/src/schema/impls.rs
```diff
@@ -153,16 +153,26 @@ unsafe impl ZeroCopy for u8 {}
 // - i8 is similarly a canonical zero-copy type: no endianness, no layout, no validation.
 unsafe impl ZeroCopy for i8 {}
 
+// SAFETY: Primitive integer types with fixed size.
+unsafe impl ZeroCopy for u16 {}
```
cpubot: Bincode spec requires `LE` encoded integers, so we should gate `ZeroCopy` by little endian targets with
```rs
#[cfg(target_endian = "little")]
```

wincode/src/schema/impls.rs
```diff
@@ -153,16 +153,26 @@ unsafe impl ZeroCopy for u8 {}
...
 impl_int!(i8, zero_copy: true);
-impl_int!(u16, zero_copy: false);
-impl_int!(i16, zero_copy: false);
-impl_int!(u32, zero_copy: false);
-impl_int!(i32, zero_copy: false);
-impl_int!(u64, zero_copy: false);
-impl_int!(i64, zero_copy: false);
-impl_int!(u128, zero_copy: false);
-impl_int!(i128, zero_copy: false);
+impl_int!(u16, zero_copy: true);
```
cpubot: Similarly these implementations should only use `zero_copy` on little endian targets. If I'm not mistaken I think we could do this for float values too

cpubot: Could probably specify that here
https://github.com/anza-xyz/wincode/blob/53bc687d9eeef6847d0a024d6f8ff503d0f376f8/wincode/src/schema/impls.rs#L59-L62
and here https://github.com/anza-xyz/wincode/blob/53bc687d9eeef6847d0a024d6f8ff503d0f376f8/wincode/src/schema/impls.rs#L78-L81

Might be nice to do it like the following:
```rs
const TYPE_META: TypeMeta = TypeMeta::Static {
    size: size_of::<$type>(),
    #[cfg(target_endian = "little")]
    zero_copy: true,
    #[cfg(not(target_endian = "little"))]
    zero_copy: $zero_copy,
};
```

cpubot (dismissed):
Looks great @vadorovsky

PR #58, opened 2025-12-17
Add support for mutable zero-copy deserialization

cpubot:
As of `v0.2.3` we have zero-copy support for immutable references.

An important feature for Solana onchain programs in particular is the ability to mutate deserialized bytes in place. Surely there are other cases where this is useful as well -- deserializing a type into a mutable reference, mutating in place, and writing those updated bytes back to persistence can save a significant amount of copying.

Supporting mutable zero-copy is straightforward. We only need a few adjustments to make this work.

1. Implement `Reader` for `&mut [u8]`.
2. Implement an associated mutable trusted `Reader`.
3. Add a `borrow_exact_mut` method to the `Reader` trait and implement it in the new `Reader`s.
4. Implement `SchemaRead` and for `&mut T` and `&mut [T]` where `T: ZeroCopy`.

I've added some additional helper methods as well to make using zero-copy a bit more user friendly as well. It is possible to use zero-copy without these helpers, but I think they are nice for ergonomics. The following functions are added to the `ZeroCopy` trait:
- `from_bytes`
    - Can be used instead of `deserialize::<&T>`, e.g., `T::from_bytes`.
- `from_bytes_mut`
    - Can be used instead of `deserialize_mut::<&mut T>`, e.g., `T::from_bytes_mut`.

### DIFF
--- a/wincode/src/serde.rs
+++ b/wincode/src/serde.rs
@@ -227,14 +227,13 @@ where
     T::deserialize(src)
 }
 
-/// Deserialize a mutable reference to a type from the given bytes.
-///
-/// This allows mutating the serialized data in place.
-///
-/// Only available for types that are [`ZeroCopy`](crate::ZeroCopy).
+/// Deserialize a type from the given bytes, with the ability
+/// to form mutable references for types that are [`ZeroCopy`](crate::ZeroCopy).
+/// This can allow mutating the serialized data in place.
 ///
 /// # Examples
 ///
+/// ## Zero-copy types
 /// ```
 /// # #[cfg(all(feature = "alloc", feature = "derive"))] {
 /// # use wincode::{SchemaWrite, SchemaRead};
@@ -252,6 +251,34 @@ where
 /// let data_mut: &mut Data = wincode::deserialize_mut(&mut serialized).unwrap();
 /// data_mut.bytes = *b"wincode";
 /// data_mut.the_answer = 42;
+///
+/// let deserialized: Data = wincode::deserialize(&serialized).unwrap();
+/// assert_eq!(deserialized, Data { bytes: *b"wincode", the_answer: 42 });
+/// # }
+/// ```
+///
+/// ## Mutable zero-copy members
+/// ```
+/// # #[cfg(all(feature = "alloc", feature = "derive"))] {
+/// # use wincode::{SchemaWrite, SchemaRead};
+/// # #[derive(Debug, PartialEq, Eq)]
+/// #[derive(SchemaWrite, SchemaRead)]
+/// struct Data {
+///     bytes: [u8; 7],
+///     the_answer: u8,
+/// }
+/// # #[derive(Debug, PartialEq, Eq)]
+/// #[derive(SchemaRead)]
+/// struct DataMut<'a> {
+///     bytes: &'a mut [u8; 7],
+///     the_answer: u8,
+/// }
+///
+/// let data = Data { bytes: [0; 7], the_answer: 42 };
+///
+/// let mut serialized = wincode::serialize(&data).unwrap();
+/// let data_mut: DataMut<'_> = wincode::deserialize_mut(&mut serialized).unwrap();
+/// *data_mut.bytes = *b"wincode";
 ///
 /// let deserialized: Data = wincode::deserialize(&serialized).unwrap();
 /// assert_eq!(deserialized, Data { bytes: *b"wincode", the_answer: 42 });


wincode/src/schema/mod.rs
```diff
@@ -2525,4 +2538,57 @@ mod tests {
...
         });
     }
+
+    #[test]
+    fn test_zero_copy_mut_roundrip() {
+        use rand::random;
+        #[derive(SchemaWrite, SchemaRead, Debug, PartialEq, Eq, proptest_derive::Arbitrary)]
+        #[wincode(internal)]
+        #[repr(C)]
+        struct Zc {
```
vadorovsky: nit: We could use the `StructZeroCopy` from the same file, which uses all the primitive types supporting zero-copy.

wincode/src/schema/mod.rs
```diff
@@ -2525,4 +2538,57 @@ mod tests {
...
+    }
+
+    #[test]
+    fn test_zero_copy_deserialize_ref() {
+        #[derive(
+            SchemaWrite, SchemaRead, Debug, PartialEq, Eq, proptest_derive::Arbitrary, Clone, Copy,
+        )]
+        #[wincode(internal)]
+        #[repr(C)]
+        struct Zc {
```
vadorovsky: same here

vadorovsky (approved):
Amazing stuff, just one nit

wincode/src/serde.rs
```diff
@@ -46,6 +46,32 @@ pub trait Deserialize<'de>: SchemaRead<'de> {
         Ok(unsafe { dst.assume_init() })
     }
 
+    /// Deserialize `bytes` into a mutable reference to `Self::Dst`.
```
kskalski: update comment to match parameter name `bytes` vs `src`

wincode/src/serde.rs
```diff
@@ -46,6 +46,32 @@ pub trait Deserialize<'de>: SchemaRead<'de> {
...
+    /// Only available for types that are [`ZeroCopy`].
+    #[inline(always)]
+    fn deserialize_mut(mut src: &'de mut [u8]) -> ReadResult<&'de mut Self::Dst>
+    where
+        Self: ZeroCopy + Sized,
+    {
+        <&mut Self as SchemaRead<'de>>::get(&mut src)
+    }
+
+    /// Deserialize `bytes` into a mutable reference to `Self::Dst`.
```
kskalski: similar name mismatch

wincode/src/schema/impls.rs
```diff
@@ -1215,6 +1244,94 @@ mod zero_copy {
...
+                //   valid storage of `T`.
+                // - Advertising `zero_copy: true` here could incorrectly enable memcpy of reference
+                //   elements (e.g., Vec<&T>), which would be unsound.
+                //
+                // We borrow the underlying `T` here, knowing it is zero-copy, but the reference itself
+                // is never considered zero-copy.
+                zero_copy: false,
+            },
+            // Should be impossible to reach.
+            _ => panic!("Type is not zero-copy"),
```
kskalski, line 1294: if it's impossible, we could use `unreachable!` here :)

wincode/src/schema/impls.rs
```diff
@@ -1215,6 +1244,94 @@ mod zero_copy {
...
+    /// Cast a `&mut [u8]` to a `&mut [T]` of a zero-copy type `T`.
+    ///
+    /// Like [`cast_slice_to_slice_t`], but for mutable slices.
+    #[inline(always)]
+    #[allow(clippy::arithmetic_side_effects)]
+    pub(super) unsafe fn cast_slice_to_slice_t_mut<T>(
+        bytes: &mut [u8],
+        len: usize,
+    ) -> ReadResult<&mut [T]> {
+        debug_assert_eq!(bytes.len(), len * size_of::<T>());
```
kskalski: since it's a debug assert, we could compare like this to avoid clippy allow
```
debug_assert_eq!(Some(bytes.len()), len.checked_mul(size_of::<T>()));
```

wincode/src/schema/mod.rs
cpubot: Good call

wincode/src/schema/impls.rs
cpubot: Can do :)

cpubot: Nvm:
```
error[E0015]: cannot call non-const formatting macro in constant functions
    --> wincode/src/schema/impls.rs:1294:18
     |
1294 |             _ => unreachable!("Type is not zero-copy"),
```

wincode/src/schema/mod.rs
cpubot: Actually, I think this is probably going to have unaligned pointer read issues in Miri. ~I'll make a separate struct that guarantees align 1.~

cpubot: Made a `BufAligned` helper to make this simpler 51c48bf73ea2d1b523044662f8248158317e0dff

wincode/src/serde.rs
```diff
@@ -46,6 +46,32 @@ pub trait Deserialize<'de>: SchemaRead<'de> {
         Ok(unsafe { dst.assume_init() })
     }
 
+    /// Deserialize `src` into a mutable reference to `Self::Dst`.
+    ///
+    /// This allows mutating the serialized data in place.
+    ///
+    /// Only available for types that are [`ZeroCopy`].
+    #[inline(always)]
+    fn deserialize_mut(mut src: &'de mut [u8]) -> ReadResult<&'de mut Self::Dst>
```
febo: Wonder whether we should use a terminology slightly different than `deserialize_*` or not. In zero-copy, we are not actually deserializing the type as other `deserialize` and `deserialize_into` methods. Maybe it will not be clear to users which one is the zero-copy version given that the names are very similar.

In bytemuck you have `from_bytes` and `from_bytes_mut` – e.g.:
* `deserialize_mut` -> `from_bytes_mut`
* `deserialize_ref` -> `from_bytes`

cpubot: I think `deserialize_ref/mut` is consistent -- you can still perform partial or full zero-copy deserialization with the `deserialize` methods as well. A reference in any struct member gives you zero-copy. `deserialize_ref/mut` is just a convenience for getting a reference to an owned zero-copy type (rather than having to specify `deserialize<&Data>`).

For example
```rs
struct PartialZeroCopy<'a> {
    integer: u64,
    bytes: &'a [u8; 8],
}
let mut buf = [0u8; 16];
buf[0..8].copy_from_slice(&42u64.to_le_bytes());
buf[8..16].copy_from_slice(&[1u8; 8]);
let data = deserialize::<PartialZeroCopy<'_>>(&buf).unwrap();
assert_eq!(data.integer, 42);
assert_eq!(data.bytes, &[1; 8]);
```

febo: > I think `deserialize_ref/mut` is consistent

Hmmm, I am not entirely convinced. 😅 `deserialize_mut` feels closer to `deserialize` than `deserialize_ref`. So maybe `deserialize_ref`/`deserialize_ref_mut` would be a better match? Then the zero-copy API is `deserialize_ref*`, which reflects the fact that you are working with references.

kskalski: the (non-public) `zero_copy` module uses `cast` for very similar signature:
` pub(super) unsafe fn cast_slice_to_t_mut<T>(bytes: &mut [u8]) -> ReadResult<&mut T>`
so there is some point to that

clearly `deserialize` and `deserialize_ref|mut` do quite different things, since the former copies data into new memory, while the others convert in-place

For the purpose of this PR `deserialize_mut` is consistent with `deserialize_ref` and if we change a name, then both should change. Seems like something to put into our "how to name things discussion", which might be postponed to future.

cpubot: How about the following:

the `Deserialize` trait provides:
- `deserialize_as_ref`
- `deserialize_as_mut`

Both of which are geared towards usage on an owned zero-copy type:
- `Account::deserialize_as_ref`
- `Account::deserialize_as_mut`

And then we provide the global function `deserialize_mut`, which is effectively the same as `deserialize`, but takes a _mutable_ byte slice `&'de mut [u8]`. This way usage between the two feels the same, where `deserialize_mut` would allow you to perform partial mutable zero-copy in the same way that `deserialize` does.

```rs
struct PartialZeroCopyMut<'a> {
    integer: u64,
    bytes: &'a mut [u8; 8],
}
let data = deserialize_mut::<PartialZeroCopyMut<'_>>(&buf).unwrap();
```

kskalski: Hm, `deserialize_mut` is constrained on `ZeroCopy` and `PartialZeroCopyMut` doesn't satisfy it, right?
The `_ref` and `_mut` versions can only do cast / transmute with static validation that such conversion is valid

Not sure if that is possible, but maybe those helpers should be part of `ZeroCopy` trait instead?

febo: > And then we provide the global function deserialize_mut, which is effectively the same as deserialize, but takes a mutable byte slice &'de mut [u8]

I think this is a good option. The fact that `deserialize` can be used with both  copy and zero-copy is cool. My comment is related to the discoverability of the zero-copy only API. Having very similar names might be a bit error prone to someone starting to use wincode/zero-copy. 

> the Deserialize trait provides:
>
> * deserialize_as_ref
> * deserialize_as_mut

I am not a big fan of `deserialize_as_mut` since it does not explicitly indicates that it is a mutable reference. The `as_ref`/`as_mut` pair is consistent with other APIs, the `deserialize` on the other hand is what will probably confuse people.

I mentioned `from_bytes`, which both bytemuck and zerocopy libraries use. I haven't seen a zero-copy library use `deserialize`. My impression is that `deserialize` is usually associated to copying data. I am not suggesting that we go down the `from_bytes` naming, just mentioning as an example. 😊

cpubot: > Hm, `deserialize_mut` is constrained on `ZeroCopy` and `PartialZeroCopyMut` doesn't satisfy it, right? The `_ref` and `_mut` versions can only do cast / transmute with static validation that such conversion is valid
> 
> Not sure if that is possible, but maybe those helpers should be part of `ZeroCopy` trait instead?

`deserialize_mut` in the previous suggestion acts pretty much exactly like `deserialize` does now.
```rs
pub fn deserialize_mut<'de, T>(src: &'mut de [u8]) -> ReadResult<T>
where
    T: SchemaRead<'de, Dst = T>,
```

Calling this on a struct with references to non-zero copy types will not compile, which is the same as how `deserialize` currently works. In particular, `SchemaRead` for `&T` or `&mut T` is only implemented when `T: ZeroCopy`, so we still get compile-time validation.

cpubot: The only reason we need constraints on the `deserialize_as_ref` or `deserialize_as_mut` in this example is because of the return types (`&Self::Dst` and `&mut Self::Dst`), which only compiles if `Self: ZeroCopy`

cpubot: > I am not a big fan of `deserialize_as_mut` since it does not explicitly indicates that it is a mutable reference. The `as_ref`/`as_mut` pair is consistent with other APIs, the `deserialize` on the other hand is what will probably confuse people.

I think `as_mut` / `as_ref` feels semantically appropriate here. There are various APIs in the std library with similar semantics, e.g., `Option::as_mut` or the `AsMut` trait generally, which should provide some semantic familiarity to users.

I personally don't think it can get much clearer semantically than `deserialize_as_mut` or `deserialize_as_ref` -- deserialize the type as a reference or a mutable reference.

> I mentioned `from_bytes`, which both bytemuck and zerocopy libraries use. I haven't seen a zero-copy library use `deserialize`. My impression is that `deserialize` is usually associated to copying data. I am not suggesting that we go down the `from_bytes` naming, just mentioning as an example. 😊

The reason I don't like `from_bytes` is that basically all deserialization is "from bytes" in some sense.

To your previous point about not explicitly indicating that the function returns a reference,
> I am not a big fan of `deserialize_as_mut` since it does not explicitly indicates that it is a mutable reference.

`from_bytes` itself doesn't indicate that you will get a reference, whereas `deserialize_as_ref` does explicitly state as such.

`from_bytes_borrowed` or `from_bytes_as_ref` might be clearer if we go the `from_bytes` route, but in my opinion we don't need to separate zero-copy APIs from typical deserialize APIs, since both paths support zero-copy, just with different signatures.

febo: > Calling this on a struct with references to non-zero copy types will not compile, which is the same as how `deserialize` currently works. In particular, `SchemaRead` for `&T` or `&mut T` is only implemented when `T: ZeroCopy`, so we still get compile-time validation.

Would it be terrible to add something like `from_bytes/from_bytes_mut` to the `ZeroCopy` trait that underneath just calls `deserialize/deserialize_mut` since you can use them with `&Self::Dst` and `&mut Self::Dst`? Then we would not need to introduce other `deserialize_*` that are only applicable to "fully" zero-copy types.

cpubot: > Would it be terrible to add something like `from_bytes/from_bytes_mut` to the `ZeroCopy` trait that underneath just calls `deserialize/deserialize_mut` since you can use them with `&Self::Dst` and `&mut Self::Dst`? Then we would not need to introduce other `deserialize_*` than are only applicable to "fully" zero-copy types.

Sure, I don't mind

cpubot: 03eb86f6b2885b650ec67193da0843ff60bce11b

wincode/src/schema/mod.rs
```diff
@@ -2525,4 +2608,110 @@ mod tests {
...
+            let mut serialized = serialize_aligned(&data).unwrap();
+            let deserialized: StructZeroCopy = deserialize(&serialized).unwrap();
+            prop_assert_eq!(deserialized, data);
+
+
+            // Mutate the serialized data in place
+            {
+                let ref_mut: &mut StructZeroCopy = deserialize_mut(&mut serialized).unwrap();
+                *ref_mut = data_rand;
+            }
```
kskalski, line 2698: that's an interesting polymorphism, I assume this would also work similar to `deserialize(&serialized)`:
```rust
let deserialized: StructZeroCopy = deserialize_mut(&mut serialized).unwrap();
```
making the result in-place modify `serialized` or not depending on forced variable type, could we add this as a test?

cpubot: Well, if you call deserialize_mut and don't annotate the return type as a mutable reference, it will just copy. 

What are you thinking we should add a test case for here?

kskalski: Right, I guess just that the code compiles, I think it's not totally obvious. We can spare testing that modifying a copy won't modify the original. ;-)

Maybe again nitpicking on names, but now I find the fact that `deserialize_mut` returns a plain object, not a mut reference, as a bit surprising (`_mut` previously referred to both input and output, now only to input). It has some logic, just highlighting the issue, which should be resolved when the test has series of supported cases.

kskalski:
```rust
let deserialized_copy: StructZeroCopy = deserialize(&serialized).unwrap();
let deserialized_ref: &StructZeroCopy = deserialize(&serialized).unwrap();
let deserialized_copy: StructZeroCopy = deserialize_mut(&mut serialized).unwrap();
let deserialized_mut_ref: &mut StructZeroCopy = deserialize_mut(&mut serialized).unwrap();
```
are all valid code, right?

cpubot: Yes, all those examples are valid.

And yes, it can definitely be a bit surprising that `deserialize_mut` can return an owned type, but this is more of an issue of lack of expressivity in the type system. For `deserialize_mut` in particular, we don't want to force the return type to be `&mut T::Dst`, because that wouldn't support partially borrowed types. E.g., the following
```rs
struct PartialZeroCopyMut<'a> {
    integer: u64,
    bytes: &'a mut [u8; 8],
}
```
would not satisfy the type signature.

kskalski:
LGTM

wincode/src/serde.rs
```diff
@@ -227,6 +227,44 @@ where
     T::deserialize(src)
 }
 
+/// Deserialize a mutable reference to a type from the given bytes.
+///
+/// This allows mutating the serialized data in place.
+///
+/// Only available for types that are [`ZeroCopy`](crate::ZeroCopy).
```
febo: This is not the case anymore, right?

cpubot: Yeah, I guess that's right generally. Though being able to actually get a mutable reference out of the function _would_ in fact require a type comprised of `ZeroCopy` members, so perhaps we make this comment more specific?

wincode/src/serde.rs
febo: The `from_bytes` and `from_bytes_mut` look great – thanks!

wincode/src/serde.rs
febo: Yes, that is a good idea. Maybe we can have an example of a "PartialZeroCopy" struct as well.

cpubot: Done: f0034acae7d4976ef24a8022ad04ec618dd5f29b

febo (approved):
Looks great!

bw-solana (approved):
LGTM

PR #63, opened 2025-12-20
Add enum benchmarks for static size optimizations

tanmay4l:
Add benchmarks for enum static size optimizations

Benchmarks unit enums and same-sized variant enums to show compile-time size calculation advantages. Includes baseline mixed-sized enum for comparison.
 

cc @cpubot

### DIFF
--- a/wincode/benches/benchmarks.rs
+++ b/wincode/benches/benchmarks.rs
@@ -603,75 +603,86 @@ macro_rules! bench_vec_enum {
                 let data: Vec<$type> = $data_gen(size);
                 let data_size = serialized_size(&data).unwrap();
 
-                group.bench_with_input(
-                    BenchmarkId::new("wincode/serialize_into", size),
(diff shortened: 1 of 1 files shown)


tanmay4l:
> Nice -- would be interesting to add `Vec<_>` impls for these too

added Vec<_> impls  too

tanmay4l:
@cpubot  hey ran into a clippy issue with wrapping_mul so just simplified the data gen to use `i as u64` instead also removed the #[repr(C)] since it was adding unnecessary padding

tanmay4l:
> > @cpubot hey ran into a clippy issue with wrapping_mul so just simplified the data gen to use `i as u64` instead also removed the #[repr(C)] since it was adding unnecessary padding
> 
> Yeah good call -- `#[repr(C)]` has no benefit from wincode's perspective on enums

done

PR #66, opened 2025-12-28
Add `assert_zero_copy` attribute macro

febo:
### Problem

`wincode` support zero-copy memory manipulation, which indicates a type can be safely be converted from an arbitrary byte sequence without copies. This is determines during the `SchemaRead` and `SchemaWrite` derivation, but there is no explicit feedback when a type fails to satisfy the zero-copy requirements.

### Solution

This PR adds a <del>`derive(ZeroCopy) #[assert_zero_copy]`</del> `#[wincode(assert_zero_copy)]` attribute that generates compile-time asserts for `ZeroCopy` implementation, providing feedback of any violations of the `ZeroCopy` requirements.

#### Examples

Here are a few examples of the asserts.

* missing `repr` attribute:
image

* missing `SchemaRead`:
image

* type with padding:
image

* non-zero-copy fields:
image

### DIFF
--- a/wincode-derive/src/common.rs
+++ b/wincode-derive/src/common.rs
@@ -545,7 +545,7 @@ pub(crate) struct SchemaArgs {
     /// Otherwise, the enum discriminants will be encoded using the default encoding (`u32`).
     #[darling(default)]
     pub(crate) tag_encoding: Option<Type>,
-    /// Indicates whether to assert that the type is zero-copy ore not.
+    /// Indicates whether to assert that the type is zero-copy or not.
     ///
     /// If specified, compile-time asserts will be generated to ensure the type meets zero-copy requirements.
     #[darling(default)]


wincode-derive/src/lib.rs
```diff
@@ -31,3 +32,13 @@ pub fn derive_schema_read(input: TokenStream) -> TokenStream {
         Err(e) => e.write_errors().into(),
     }
 }
+
+/// Assert that a struct is zero-copy deserializable.
+#[proc_macro_derive(ZeroCopy, attributes(wincode))]
```
febo: Not sure whether this should be named `SchemaZeroCopy` or not to be more consistent with `SchemaRead` and `SchemaWrite`. Or perhaps a name to reflect that it does only asserts.

febo: Renamed to `assert_zero_copy`.

febo:
> What do you think?

This looks good and it indeed makes more sense since it only adds compile-time asserts. Updated on https://github.com/anza-xyz/wincode/pull/66/commits/f412d8b0aae77fdec9d34500e55b70053c812114.

febo:
Having another look at this and I am inclined to think that  it is better to revert the macro back to a `derive` macro and move the `ZeroCopy` derivation to it.

It feels a bit more natural to write:
 ```rust
 #[derive(SchemaRead, SchemaWrite, ZeroCopy)]
 pub struct MyZeroCopyStruct { ... }
```
than:
 ```rust
#[assert_zero_copy]
 pub struct MyZeroCopyStruct { ... }
```

It is a bit unexpected that deriving `SchemaRead` also derives `ZeroCopy` under "limited" conditions. Not only from the fact that there is no feedback when `ZeroCopy` cannot be derived, which this PR offers, but because the user did not explicitly asked to do that. My impression is that in general we should have *one derive = one public trait* implementation. Unless deriving `ZeroCopy` is essential for `SchemaRead`, which I am not sure but seems that it is not required.

I don't feel super strong about it, more like a preference, so either way works for me. 😊

sam0x17 (approved):
LGTM, nice separation of macro implementation and unit-testable inner shell. Personally I usually like to do these inner shells like:

```rust
use syn::{Result, TokenStream as TokenStream2};

pub fn some_macro_impl(attr: impl Into<TokenStream2>, item: impl Into<TokenStream2>) -> Result<TokenStream2> {
    ...
}
```

Then you can very easily unit test this, and the fallibility makes error handling natural and easy. Also a neat trick is to use the impl Into<TokenStream2> so you can pass a variety of types, such as `TokenStream`, `TokenStream2`, and many `syn` types, directly making tests more ergonomic.

But what is written now is fine and better than most macros in most crates already :)

febo:
> This is what I mean above with the "derive asymmetry" -- users must be aware of the third `ZeroCopy` derive and be sure to include it to ensure downstream consumers can use it in a borrowed context.

This argument convinced me. 😀

febo:
> LGTM, nice separation of macro implementation and unit-testable inner shell. Personally I usually like to do these inner shells like:
> 
> ```rust
> use syn::{Result, TokenStream as TokenStream2};
> 
> pub fn some_macro_impl(attr: impl Into<TokenStream2>, item: impl Into<TokenStream2>) -> Result<TokenStream2> {
>     ...
> }
> ```
> 
> Then you can very easily unit test this, and the fallibility makes error handling natural and easy. Also a neat trick is to use the impl Into so you can pass a variety of types, such as `TokenStream`, `TokenStream2`, and many `syn` types, directly making tests more ergonomic.
> 
> But what is written now is fine and better than most macros in most crates already :)

Nice! I left as it is since it follows the current pattern. We could address this in a separate PR to update all macros.

febo:
Made some tweaks since I realised that we don't need to enforce `SchemaWrite` for zero-copy.

wincode-derive/src/assert_zero_copy.rs
```diff
@@ -0,0 +1,77 @@
...
+            "The struct representation is not eligible for zero-copy. Consider using \
+             #[repr(transparent)] or #[repr(C)] on the struct.",
+        ));
+    }
+
+    Ok(quote! {
+        const _: () = {
+            // Assert the struct implements `SchemaRead`.
+            const _assert_schema_read_impl: fn() = || {
+                fn assert_schema_read_impl<'de, T: ::wincode::SchemaRead<'de>>() {}
```
cpubot: It would be nice to use [`get_crate_name`](https://github.com/anza-xyz/wincode/blob/9f0ffa346d95c31b94486b7bfea724b73330c42f/wincode-derive/src/common.rs#L479) rather than hardcoding `::wincode` here so we can use it internally if needed. You can assign it to a variable at the top of the function here and use it in quoted sections

febo: Done in https://github.com/anza-xyz/wincode/pull/66/commits/d37f659b2c42231f5c4f40e63666f65936e6bbda

cpubot (dismissed):
Looks good to me! Thanks for adding this.

Another route to consider:

We could make `assert_zero_copy` a property of [`SchemaArgs`](https://github.com/anza-xyz/wincode/blob/e3c5164c1e3556ae33d678ea90b8d9a9602385d4/wincode-derive/src/common.rs#L516-L548). This would allow it to be used without an explicit import of `assert_zero_copy`, similar to how we expose options like `from`, `no_suppress_unused`, `tag_encoding`, etc (https://docs.rs/wincode/latest/wincode/#derive-attributes).

E.g.,
```rs
#[repr(C)]
#[wincode(assert_zero_copy)]
struct MyStruct {
    a: u8
}
```

This might make it feel more coherent with the rest of the derive attributes.

If we want to go this route, we'd update the `SchemaRead` derive to generate the const assertions (could just be a function call imported from `assert_zero_copy.rs`, with slight modifications) and output them adjacent to the trait implementation.

My preference leans slightly in this direction, just because it coheres more with existing derive attributes, but don't feel strongly enough to block. Let me know what you think.

febo:
> My preference leans slightly in this direction, just because it coheres more with existing derive attributes, but don't feel strongly enough to block. Let me know what you think.

Agree – made the changes in https://github.com/anza-xyz/wincode/pull/66/commits/23dee127a7ffd736fc514e2bc3f8ae5a68dbe595

wincode-derive/src/common.rs
```diff
@@ -545,6 +545,11 @@ pub(crate) struct SchemaArgs {
     /// Otherwise, the enum discriminants will be encoded using the default encoding (`u32`).
     #[darling(default)]
     pub(crate) tag_encoding: Option<Type>,
+    /// Indicates whether to assert that the type is zero-copy ore not.
```
cpubot:
```suggestion
    /// Indicates whether to assert that the type is zero-copy or not.
```

cpubot (commented):
Looks good! Just a typo

PR #67, opened 2026-01-05
Add global configuration

cpubot:
## Summary
This PR adds a global configuration trait and builder to wincode. This scheme allows tuning various constant knobs as well as customizing various defaults like the default length encoding for sequences, all available at compile time so that we do not introduce any additional runtime overhead. 

This yields significant ergonomic improvements (for tuning things like global preallocation limits and length encodings), as well as important optimization parameters for targets like onchain programs. What's more, it paves the way for additional configuration, like variable length integer encodings and customizable default enum discriminant encodings. This will allow wincode to interoperate seamlessly with folks coming from Bincode 2.0, where the `standard` configuration uses varint encodings, as well other serialization libraries like borsch where sequence lengths are encoded as `u32`s and enum discriminants as `u8`.

## Reviewing

Aside from the configuration definition itself and slight tweaks to `SeqLen`, _most_ of the changes in this PR are mechanical -- updating various implementations of `SchemaRead` and `SchemaWrite` to accept a `Config` parameter, and associated adjustments in macros. Commits should be individually reviewable and are logically atomic, so reviewing by commit may be ideal.

## Implementation
A `config` module is introduced, which contains `Config` and `ConfigCore` traits, and config-dependent serialize, deserialize, and zero-copy traits and functions. 

### `Config` / `ConfigCore`

The following configuration parameters are supported:
- Preallocation size limit (`usize`) - sets the preallocation size limit for `deserialize*` calls.
- Zero-copy alignment check (`bool`) - enables / disables the pointer alignment check in zero-copy paths.
- Length encoding (`SeqLen`) - sets the default length encoding for sequences (which are still overridable in individual cases by using the `containers` module).

Note that integer encoding and enum discriminant encodings are not included here, as I'm reserving those for follow up PRs to keep this one smaller.

### `Config` vs `ConfigCore`

It may seem odd that we have *two* configuration traits. The reason for this is specifically to support configuration parameters that may themselves depend on configuration. The `SeqLen` trait depends on `ConfigCore`, so that it has access to things like `PREALLOCATION_SIZE_LIMIT` (and in the future the integer encoding). Without this split, there is a circular dependency, which overflows the compiler (`Config` -> `SeqLen<Config>` -> `Config`, etc). The general rule of thumb for implementations is that they should always prefer `ConfigCore`, unless they need access to config-dependent parameters (e.g., `SeqLen`).

### Config dependent serialize/deserialize/zero-copy

The `config` module also exposes functions and traits that mirror those exported from the crate root. Rather than adding a config parameter to crate root traits and functions, this makes migration from previous versions significantly simpler, and yields a simpler default experience for those that want to use the defaults.

For example, this still works as it does today:
```rs
let bytes = wincode::serialize(&data);
let val = wincode::deserialize(&bytes);
```

And to use the config-aware versions:
```rs
let bytes = wincode::config::serialize(&data, config);
let val = wincode::config::deserialize(&bytes, config);
```

#### A note on config-aware zero-copy

It might be curious at first that the `ZeroCopy` trait now depends on a `Config`. The reason for this is that we can't generally make the assumption that zero-copy holds for all configurations. In particular, when varint encoding is introduced as a configuration parameter, integers are no longer zero-copy eligible.

### Config dependent `SchemaWrite` & `SchemaRead`

Naturally, `SchemaWrite` and `SchemaRead` now have a generic `ConfigCore` type parameter. This allows them to vary their implementations based on the configuration. For example, the `SchemaRead` zero-copy implementation has access to the `ZERO_COPY_ALIGN_CHECK` constant, allowing it vary whether it performs pointer alignment checks. The same is true for integer encodings and default sequence lengths.

### `containers` module

The `containers` module provides implementations of std library collection types with an additional length encoding parameter. Prior to this PR, the length encoding parameter had a default. This is reasonable when global length encoding configuration is not possible. Since this PR *does* provide global length encoding configuration, it's likely not a good idea to retain this default. In particular, the container types (e.g., `Vec<T, LengthEncoding>`) don't have access to `Config` in their declarations (and it almost certainly wouldn't make sense to include one), so whatever default we include there can differ from whatever is specified in `Config`, which I don't think is a good experience.

### `assert_zero_copy` derive attribute

Since `ZeroCopy` is config dependent, I've updated the `assert_zero_copy` attribute to take an optional `Path` value. This is a convenience for folks that want to assert the zero-copy nature of their types relative to a certain configuration. If no path is provided, the `DefaultConfig` is assumed.
E.g.,
```rs
#[derive(SchemaRead)]
// Checks `ZeroCopy<DefaultConfig>`
#[wincode(assert_zero_copy)]
```
```rs
#[derive(SchemaRead)]
// Checks `ZeroCopy<MyConfig>`
#[wincode(assert_zero_copy(MyConfig)]
```

### DIFF
--- a/wincode/src/config/mod.rs
+++ b/wincode/src/config/mod.rs
@@ -1,7 +1,7 @@
 //! Global configuration for wincode.
 //!
 //! This module provides configuration types and structs for configuring wincode's behavior.
-//! See [`ConfigBuilder`] for more details on how to configure wincode.
+//! See [`Configuration`] for more details on how to configure wincode.
 //!
 //! Additionally, this module provides traits and functions that mirror the serialization,
 //! deserialization, and zero-copy traits and functions from the crate root, but with an
@@ -14,13 +14,13 @@ use {
 pub const DEFAULT_PREALLOCATION_SIZE_LIMIT: usize = 4 << 20; // 4 MiB
 pub const PREALLOCATION_SIZE_LIMIT_DISABLED: usize = usize::MAX;
 
-/// Builder to configure global settings.
+/// Compile-time configuration for runtime behavior.
 ///
 /// Defaults:
 /// - Zero-copy alignment check is enabled.
 /// - Preallocation size limit is 4 MiB.
 /// - Length encoding is [`BincodeLen`].
-pub struct ConfigBuilder<
+pub struct Configuration<
     const ZERO_COPY_ALIGN_CHECK: bool = true,
     const PREALLOCATION_SIZE_LIMIT: usize = DEFAULT_PREALLOCATION_SIZE_LIMIT,
     LengthEncoding = BincodeLen,
@@ -29,7 +29,7 @@ pub struct ConfigBuilder<
 }
 
 impl<const ZERO_COPY_ALIGN_CHECK: bool, const PREALLOCATION_SIZE_LIMIT: usize, LengthEncoding> Clone
-    for ConfigBuilder<ZERO_COPY_ALIGN_CHECK, PREALLOCATION_SIZE_LIMIT, LengthEncoding>
+    for Configuration<ZERO_COPY_ALIGN_CHECK, PREALLOCATION_SIZE_LIMIT, LengthEncoding>
 {
     fn clone(&self) -> Self {
         *self
@@ -37,7 +37,7 @@ impl<const ZERO_COPY_ALIGN_CHECK: bool, const PREALLOCATION_SIZE_LIMIT: usize, L
 }
 
 impl<const ZERO_COPY_ALIGN_CHECK: bool, const PREALLOCATION_SIZE_LIMIT: usize, LengthEncoding> Copy
-    for ConfigBuilder<ZERO_COPY_ALIGN_CHECK, PREALLOCATION_SIZE_LIMIT, LengthEncoding>
+    for Configuration<ZERO_COPY_ALIGN_CHECK, PREALLOCATION_SIZE_LIMIT, LengthEncoding>
 {
 }
 
@@ -45,12 +45,12 @@ const fn generate<
     const ZERO_COPY_ALIGN_CHECK: bool,
     const PREALLOCATION_SIZE_LIMIT: usize,
     LengthEncoding,
->() -> ConfigBuilder<ZERO_COPY_ALIGN_CHECK, PREALLOCATION_SIZE_LIMIT, LengthEncoding> {
-    ConfigBuilder { _l: PhantomData }
+>() -> Configuration<ZERO_COPY_ALIGN_CHECK, PREALLOCATION_SIZE_LIMIT, LengthEncoding> {
+    Configuration { _l: PhantomData }
 }
 
-impl ConfigBuilder {
-    /// Create a new config builder with the default settings.
+impl Configuration {
+    /// Create a new configuration with the default settings.
     ///
     /// Defaults:
     /// - Zero-copy alignment check is enabled.
@@ -61,10 +61,10 @@ impl ConfigBuilder {
     }
 }
 
-pub type DefaultConfig = ConfigBuilder;
+pub type DefaultConfig = Configuration;
 
 impl<const ZERO_COPY_ALIGN_CHECK: bool, const PREALLOCATION_SIZE_LIMIT: usize, LengthEncoding>
-    ConfigBuilder<ZERO_COPY_ALIGN_CHECK, PREALLOCATION_SIZE_LIMIT, LengthEncoding>
+    Configuration<ZERO_COPY_ALIGN_CHECK, PREALLOCATION_SIZE_LIMIT, LengthEncoding>
 {
     #[expect(clippy::new_without_default)]
     pub const fn new() -> Self {
@@ -79,9 +79,9 @@ impl<const ZERO_COPY_ALIGN_CHECK: bool, const PREALLOCATION_SIZE_LIMIT: usize, L
     /// [`containers`](crate::containers).
     pub const fn with_length_encoding<L>(
         self,
-    ) -> ConfigBuilder<ZERO_COPY_ALIGN_CHECK, PREALLOCATION_SIZE_LIMIT, L>
+    ) -> Configuration<ZERO_COPY_ALIGN_CHECK, PREALLOCATION_SIZE_LIMIT, L>
     where
-        ConfigBuilder<ZERO_COPY_ALIGN_CHECK, PREALLOCATION_SIZE_LIMIT, L>: Config,
+        Configuration<ZERO_COPY_ALIGN_CHECK, PREALLOCATION_SIZE_LIMIT, L>: Config,
     {
         generate()
     }
@@ -95,7 +95,7 @@ impl<const ZERO_COPY_ALIGN_CHECK: bool, const PREALLOCATION_SIZE_LIMIT: usize, L
     /// This is enabled by default.
     pub const fn enable_zero_copy_align_check(
         self,
-    ) -> ConfigBuilder<true, PREALLOCATION_SIZE_LIMIT, LengthEncoding> {
+    ) -> Configuration<true, PREALLOCATION_SIZE_LIMIT, LengthEncoding> {
         generate()
     }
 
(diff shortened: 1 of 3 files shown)


wincode-derive/src/schema_read.rs
```diff
@@ -204,59 +204,17 @@ fn impl_struct_extensions(args: &SchemaArgs, crate_name: &Path) -> Result<TokenS
...
-        let field_projection_type = if args.from.is_some() {
-            // If the user is defining a mapping type, we need the type system to resolve the
-            // projection destination.
-            quote! { <#ty as SchemaRead<'de>>::Dst }
-        } else {
-            // Otherwise we can use the type directly.
-            // This makes the generated type more scrutable.
-            quote! { #ty }
-        };
-        quote! {
```
cpubot, line 233: These were deprecated in `0.2.2`. This PR is a breaking change, so ripping it out

wincode/src/config/mod.rs
```diff
@@ -0,0 +1,200 @@
...
+{
+    fn clone(&self) -> Self {
+        *self
+    }
+}
+
+impl<const ZERO_COPY_ALIGN_CHECK: bool, const PREALLOCATION_SIZE_LIMIT: usize, LengthEncoding> Copy
+    for ConfigBuilder<ZERO_COPY_ALIGN_CHECK, PREALLOCATION_SIZE_LIMIT, LengthEncoding>
+{
+}
```
kskalski, line 42: it appears `#[derive(Clone, Copy)]` works fine for `ConfigBuilder`

wincode/src/serde.rs
```diff
@@ -36,43 +37,43 @@ use {
...
+pub trait Deserialize<'de>: SchemaRead<'de, DefaultConfig> {
+    /// Deserialize the input bytes into a new `Self::Dst`.
     #[inline(always)]
     fn deserialize(mut src: &'de [u8]) -> ReadResult<Self::Dst> {
-        <Self as SchemaRead<'de>>::get(&mut src)
+        Self::get(&mut src)
     }
 
-    /// Deserialize `bytes` into `target`.
+    /// Deserialize the input bytes into `dst`.
```
kskalski: I would mention `src` here, e.g.
```
/// Deserialize the input `src` bytes into `dst`.
```
or maybe
```
/// Deserialize the bytes from `src` into `dst`.
```

wincode/src/config/mod.rs
cpubot: The issue is that derive adds `Clone` `Copy` bounds to generic type parameters, which means `ConfigBuilder` would only satisfy `Clone` `Copy` if its generic bounds also satisfy `Clone` `Copy`, which isn't relevant since it only contains `PhantomData`.

wincode/src/serde.rs
cpubot: Updated

wincode/src/config/mod.rs
kskalski: Given that generic parameters and whole config are stateless, maybe that's not that bad to require them to be copyable... but surely we can relax it and make the usage more ergonomic (I suppose for the purpose of `LengthEncoding` being some custom implementation of `SeqLen` that won't have `Copy`).

I'm wondering though - do you foresee config to ever hold any state or will rather enforce it to be compile-only utility? I'm asking, because the function parameters `config: C` caught my eye - `Config` doesn't have any function, all its functionality is obtained by accessing consts through `C::*`. Maybe we don't need the parameters at all and the generic type `C` is enough?
But if config might contain state in the future, then I guess this API is more future-proof.

cpubot: > but surely we can relax it and make the usage more ergonomic (I suppose for the purpose of LengthEncoding being some custom implementation of SeqLen that won't have Copy).

`SeqLen` is necessarily stateless given the trait provides only static methods.

> I'm wondering though - do you foresee config to ever hold any state or will rather enforce it to be compile-only utility

It's purely a compile time construction, and at least right now I can't think of any functionality we'd need that would require state.

> I'm asking, because the function parameters config: C caught my eye - Config doesn't have any function, all its functionality is obtained by accessing consts through C::*. Maybe we don't need the parameters at all and the generic type C is enough?

One reason for this is ergonomics. It's a bit cumbersome to define the full config builder type given all the types and constants, e.g., `ConfigBuilder<true, {4 << 20}, BincodeLen>`. By allowing the methods to take a `config: C`, the compiler can infer the type from the builder instance, e.g., 
```rs
let config = ConfigBuilder::default().with_length_encoding::<FixInt<u32>>();
serialize(&data, config)
```

Additionally I think users should go through `ConfigBuilder`'s API, which requires users to use `unsafe` methods (and understand the safety implications) for things like `disable_zero_copy_align_check` -- accepting a `config: C` enforces this.

wincode/src/config/mod.rs
```diff
@@ -0,0 +1,200 @@
...
+pub const DEFAULT_PREALLOCATION_SIZE_LIMIT: usize = 4 << 20; // 4 MiB
+pub const PREALLOCATION_SIZE_LIMIT_DISABLED: usize = usize::MAX;
+
+/// Builder to configure global settings.
+///
+/// Defaults:
+/// - Zero-copy alignment check is enabled.
+/// - Preallocation size limit is 4 MiB.
+/// - Length encoding is [`BincodeLen`].
+pub struct ConfigBuilder<
```
kskalski: I would actually drop the "builder" part, it doesn't follow the typical builder pattern where you have "state under construction" and then `finalize` / `build` the complete state.

I think bincode's original design (https://docs.rs/bincode/2.0.1/bincode/config/struct.Configuration.html) actually makes more sense - starting with default and on each adjustment you have a complete and usable configuration so even if there is a "building" process, it's just updating one config into another.

wincode/src/config/mod.rs
kskalski: Ok, the improved type inference is a good reason. I guess this is the cost of `SchemaWrite` and `SchemaRead` being traits that we want to parametrize (not sure how viable such API would be, but ideally I would see some "serializer" object being created with config as parameter that would then be used to call serialize on the input).

wincode/src/config/mod.rs
cpubot: > I think bincode's original design (https://docs.rs/bincode/2.0.1/bincode/config/struct.Configuration.html) actually makes more sense - starting with default and on each adjustment you have a complete and usable configuration so even if there is a "building" process, it's just updating one config into another.

This is pretty much the exact design we have here. Their `Configuration` struct implements their [`Config`](https://docs.rs/bincode/2.0.1/bincode/config/trait.Config.html) trait, which exposes the configured types and constants as trait associated types and constants so that users don't have to introduce a ton of parameters at implementation sites (e.g., can just use `Config::Type`). I can change `Config` to `ConfigBuilder` if you prefer that.

kskalski: I simply meant to change `ConfigBuilder` to `Configuration`

cpubot: Ah yes, that's what I meant to write -- typo! Yes, I don't mind

cpubot: 5eae78c471476f110fd8b398fc613fbf6af3bf8d

wincode/src/config/mod.rs
```diff
@@ -0,0 +1,200 @@
...
+//! additional configuration parameter.
+use {
+    crate::len::{BincodeLen, SeqLen},
+    core::marker::PhantomData,
+};
+
+pub const DEFAULT_PREALLOCATION_SIZE_LIMIT: usize = 4 << 20; // 4 MiB
+pub const PREALLOCATION_SIZE_LIMIT_DISABLED: usize = usize::MAX;
+
+/// Builder to configure global settings.
```
kskalski: worth updating comments to for `builder`, also might use a summary like
```
Compile-time adjustable configuration to control runtime behavior 
```

wincode/src/config/mod.rs
```diff
@@ -0,0 +1,200 @@
...
+const fn generate<
+    const ZERO_COPY_ALIGN_CHECK: bool,
+    const PREALLOCATION_SIZE_LIMIT: usize,
+    LengthEncoding,
+>() -> Configuration<ZERO_COPY_ALIGN_CHECK, PREALLOCATION_SIZE_LIMIT, LengthEncoding> {
+    Configuration { _l: PhantomData }
+}
+
+impl Configuration {
+    /// Create a new config builder with the default settings.
```
kskalski: `Create a new configuration with the default settings`

wincode/src/config/mod.rs
cpubot: Good call. Updated

kskalski (approved):
LGTM

PR #69, opened 2026-01-10
test: add roundtrip/bincode equivalence fuzzer

0xalpharush:
Add a fuzzer that tests input are accepted/rejected and serialized in accordance with bincode v1.

cc @nbelenkov

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,3 +41,17 @@ jobs:
         run: cargo clippy --all-targets --features alloc -- -D warnings
       - name: tests
         run: cargo test --all-features --all-targets --no-fail-fast --workspace --exclude wincode-fuzz
+
+  fuzz:
+    name: fuzz smoke check (nightly)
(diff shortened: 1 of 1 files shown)


nbelenkov:
Nice, thank you for writing this up. How are you generating the corpus for this? If its via a bash script or smth, could be worth checking it into the repo as well, so would be easier to run later on

0xalpharush:
@nbelenkov I can open a follow up PR that makes creating a seed corpus convenient

0xalpharush:
Added the `wincode-fuzz` package as an exclude to miri and unittest runs.

I can add a smoke check that builds and runs the fuzzer but wanted permission before modifying CI further.

Code here https://github.com/0xalpharush/wincode/pull/1/commits/bd99f4d98147c383e352be88ce5ffd8d8cd5cd2e 
and example run
https://github.com/0xalpharush/wincode/actions/runs/21007744960/job/60394418920?pr=1

cpubot (approved):

PR #70, opened 2026-01-12
Add collection benchmarks for BTreeMap, BTreeSet, LinkedList, VecDeque

tanmay4l:
Adds benchmarks for collections that were implemented but not benchmarked.

- BTreeMap<u64, u64>
- BTreeSet<u64>
- LinkedList<u64>
- VecDeque<u64>

Uses a macro to avoid code duplication (same pattern as enum benchmarks)


tanmay4l:
@cpubot fixed the clippy

PR #73, opened 2026-01-15
Support wincode(skip) in derive for fields

kskalski:
Serde has a convenient `#[serde(skip)]` marker for fields that should be excluded from serialization and deserialization.
It would be good to have something similar in wincode such that it's easy to adapt existing structs for use with wincode.

Support `skip` attribute:
* do not read from reader, instead initalize the field in derived read implementaiton
* do not write to writer
* builder workings are unchanged and the field still needs to be initialized when using `struct_extensions`

Support modes:
* `#[wincode(skip(default)]` and (shortcut `#[wincode(skip(default)]`) initializes the field with `Default::default()` 
* `#[wincode(skip(default_val = expr)]` initializes the field injecting `expr` into the derived read implementation (expr can be a constant or any piece of code that returns proper value)

### DIFF
--- a/wincode/src/lib.rs
+++ b/wincode/src/lib.rs
@@ -241,7 +241,7 @@
 //! let serialized = wincode::serialize(&byte_ref).unwrap();
 //! let deserialized: ByteRef<'_> = wincode::deserialize(&serialized).unwrap();
 //! assert_eq!(byte_ref, deserialized);
-//! # }  
+//! # }
 //! ```
 //!
 //! ### struct newtype
@@ -486,6 +486,14 @@
 //! |Attribute|Type|Default|Description
 //! |---|---|---|---|
 //! |`with`|`Type`|`None`|Overrides the default `SchemaRead` or `SchemaWrite` implementation for the field.|
+//! |`skip`|`bool`\|`Expr`|`false`|Skips the field during serialization and deserialization (initializing with default value).|
+//!
+//! ### `skip`
+//!
+//! Allows omitting the field during serialization and deserialization. When type is initialized
+//! during deserialization, the field will be set to the default value. This is typically
+//! `Default::default()` (when using `#[wincode(skip)]` or `#[wincode(skip(default))]`), but can
+//! be overridden by specifying `#[wincode(skip(default_val = <value>))]`.
 //!
 //! ## Variant level (enum variants)
 //! |Attribute|Type|Default|Description


wincode/src/schema/mod.rs
```diff
@@ -1302,6 +1302,58 @@ mod tests {
...
+                b: test.b,
+                ..deserialized
+            };
+            prop_assert_eq!(reinitialized, test);
+        });
+
+        proptest!(proptest_cfg(), |(test: TestEnum)| {
+            let mut serialized = serialize(&test).unwrap();
+            let mut uninit_zeroed = MaybeUninit::<TestEnum>::zeroed();
+            TestEnum::deserialize_into(serialized.as_mut(), &mut uninit_zeroed).unwrap();
```
kskalski, line 1397: This fails miri with UB, apparently because of this impl detail
https://github.com/anza-xyz/wincode/blob/6feb75ff4c5b26275b0ee8bc4a12ea906d996a02/wincode-derive/src/schema_read.rs#L484

I suppose we can't create uninitialized enum and properly address each field's offset, however I wonder if we could re-use existing memory provided by the `read(reader: &mut impl Reader<'de>, dst: &mut MaybeUninit<Self::Dst>)` treating it as temporary sandbox.
The current pattern for reading enum seems to be along the lines of:
```rust
enum X {
  V(A, B)
}

fn read(reader: &mut impl Reader<'de>, dst: &mut MaybeUninit<Self::Dst>) {
  let uninit_field_a = MaybeUninit::<A>::uninit();
  let a = A::read(reader, &mut uninit_field_a)?;
  let uninit_field_b = MaybeUninit::<B>::uninit();
  let b = B::read(reader, &mut uninit_field_b)?;
  dst.write(V(a, b))
}
```

but maybe we could do something like:
```rust
fn read(reader: &mut impl Reader<'de>, dst: &mut MaybeUninit<Self::Dst>) {
  let uninit_field_a_mut_ref = .. dst raw pointer conversion ...;
  let a = A::read(reader, uninit_field_a_mut_ref)?;
  let uninit_field_b_mut_ref = .. dst raw pointer conversion ...;
  let b = B::read(reader, uninit_field_b_mut_ref)?;
  dst.write(V(a, b))
}
```

kskalski:
Ok, I got the implementation for `skip[(default)]`, `skip(default_val = X)` and `skip(default_with = |dst: &mut MaybeUninit<_>| { dst.write(X); })`
There is a bit of mess in iterating over fields in different with and without skipped fields, I will try to clean it up more, but any comments at this point will be appreciated.

kskalski:
I had similar thoughts - the function with such signature doesn't guarantee the value is initialized. `MaybeUninit::write` returns `mut &T`, but even if we could be sure that reference is in fact the same memory pointer, I guess forcing the function to eventually call `write` is not different than just using `default_val = f()` where `f()` returns a value.

In fact I think `default_val` taking arbitrary expression is powerful enough to refrain from adding `default_with` until we have a more clear use-case.

wincode-derive/src/schema_read.rs
```diff
@@ -55,12 +55,21 @@ fn impl_struct(
...
-                    reader,
-                    unsafe { &mut *(&raw mut (*dst_ptr).#ident).cast::<#hint>() }
-                )?;
-                #init_count
+            if let Some(mode) = &field.skip {
+                match mode {
+                    SkipMode::Default => {
+                        quote! { unsafe { (*dst_ptr).#ident = Default::default() }; }
+                    }
+                    SkipMode::DefaultVal(val) => quote! { unsafe { (*dst_ptr).#ident = #val }; },
```
cpubot: I believe this will treat the field as initialized and attempt to drop the value first before assigning. 

Probably should use
```rs
unsafe { (&raw mut (*dst_ptr).#ident).write(#val) };
```

kskalski: updated and also added the missing `init_count` statement when default is assigned

wincode-derive/src/common.rs
```diff
@@ -221,9 +251,12 @@ impl FieldsExt for Fields<Field> {
...
         // known at compile time.
-        let anon_idents = self.member_anon_ident_iter(None).collect::<Vec<_>>();
-        let zero_copy_idents = self.member_anon_ident_iter(Some("zc_")).collect::<Vec<_>>();
-        let is_zero_copy_eligible = repr.is_zero_copy_eligible();
+        let anon_idents = self.unskipped_anon_ident_iter(None).collect::<Vec<_>>();
+        let zero_copy_idents = self
+            .unskipped_anon_ident_iter(Some("zc_"))
+            .collect::<Vec<_>>();
+        let has_skipped_fields = anon_idents.len() != self.len();
+        let is_zero_copy_eligible = !has_skipped_fields && repr.is_zero_copy_eligible();
```
cpubot: I think the padding check would actually still prevent zero-copy being `true` in the skipped fields case without the `has_skipped_fields` check.
https://github.com/anza-xyz/wincode/blob/1307e7227686ecb23722384790283d9db2cfbe8b/wincode-derive/src/common.rs#L243

The size of the skipped fields wouldn't be represented in `serialized_size`, so that check should fail if skipped fields are not zero-sized. 

Additionally, potential edge case where user specifies `#[wincode(skip)]` on a zero-sized type -- that type _could_ still qualify for zero-copy if all the other conditions hold, and the `has_skipped_fields` check would preclude that qualification

wincode-derive/src/schema_read.rs
```diff
@@ -621,6 +639,7 @@ pub(crate) fn generate(input: DeriveInput) -> Result<TokenStream> {
     let zero_copy_impl = match &args.data {
         Data::Struct(fields)
             if repr.is_zero_copy_eligible()
+                && !has_skipped_struct_fields
```
cpubot: See comment above about whether we need to strictly prohibit this https://github.com/anza-xyz/wincode/pull/73/changes#r2755970184

wincode-derive/src/common.rs
kskalski: I'm a bit worried if the padding check could generate a false positive if the sizes "balance out", say one field has serialization size > `size_of()` by the amount of skipped field's `size_of()`, but I suppose this isn't possible when all fields are also zero-copy.

wincode/src/lib.rs
```diff
@@ -325,6 +314,7 @@
 //! |`struct_extensions`|`bool`|`false`|Generates placement initialization helpers on `SchemaRead` struct implementations|
 //! |`tag_encoding`|`Type`|`None`|Specifies the encoding/decoding schema to use for the variant discriminant. Only usable on enums.|
 //! |`assert_zero_copy`|`bool`\|`Path`|`false`|Generates compile-time asserts to ensure the type meets zero-copy requirements. Can specify a custom config path, will use the [`DefaultConfig`](config::DefaultConfig) if `bool` form is used.|
+//! |`skip`|`bool`\|`Expr`|`false`|Skips the field during serialization and deserialization (initializing with default value).|
```
cpubot: This should actually go in the `## Field level` heading

https://github.com/anza-xyz/wincode/blob/1307e7227686ecb23722384790283d9db2cfbe8b/wincode/src/lib.rs#L500-L503

This `## Top level` heading is for _struct_ level attributes (e.g., `tag_encoding`, `from`, `struct_extensions`, etc)

PR #74, opened 2026-01-15
Add integer encoding configuration

cpubot:
This PR adds support for configuring wincode's integer encoding. It does not yet add support for bincode's varint encoding, as I will do that in a follow up PR. This PR sets up integer encoding specific parameter support in `Configuration`, associated types in `Config`, `IntEncoding` trait, `ByteOrder` trait, and provides a `FixInt` implementation that matches the current behavior.

Care has been taken to ensure that zero-copy behavior remains sane and safe to use. `IntEncoding` provides `STATIC` and `ZERO_COPY` constants that can be used in the downstream `SchemaRead` / `SchemaWrite` implementations. Additional marker traits like `PlatformEndianness` are also provided, which allows constraining `ZeroCopy` implementations where platform endianness matches configuration's endianness.

### DIFF
--- a/wincode/src/config/serde.rs
+++ b/wincode/src/config/serde.rs
@@ -95,8 +95,8 @@ impl<T, C: Config> DeserializeOwned<C> for T where T: SchemaReadOwned<C> {}
 ///
 /// ```
 /// # #[cfg(feature = "alloc")] {
-/// # use wincode::{config::Configuration, len::FixInt};
-/// let config = Configuration::default().with_length_encoding::<FixInt<u32>>();
+/// # use wincode::{config::Configuration, len::FixIntLen};
+/// let config = Configuration::default().with_length_encoding::<FixIntLen<u32>>();
 /// let vec: Vec<u8> = vec![1, 2, 3];
 /// let bytes = wincode::config::serialize(&vec, config).unwrap();
 /// assert_eq!(vec.len(), u32::from_le_bytes(bytes[0..4].try_into().unwrap()) as usize);
@@ -134,8 +134,8 @@ where
 ///
 /// ```
 /// # #[cfg(feature = "alloc")] {
-/// # use wincode::{config::Configuration, len::FixInt};
-/// let config = Configuration::default().with_length_encoding::<FixInt<u32>>();
+/// # use wincode::{config::Configuration, len::FixIntLen};
+/// let config = Configuration::default().with_length_encoding::<FixIntLen<u32>>();
 /// let vec: Vec<u8> = vec![1, 2, 3];
 /// let bytes = wincode::config::serialize(&vec, config).unwrap();
 /// let deserialized: Vec<u8> = wincode::config::deserialize(&bytes, config).unwrap();
--- a/wincode/src/len.rs
+++ b/wincode/src/len.rs
@@ -52,15 +52,15 @@ pub trait SeqLen<C: ConfigCore> {
 
 /// Use the configuration's integer encoding for sequence length encoding.
 ///
-/// For example, if the configuration's integer encoding is `FixInt`, then `UseInt<u64>`
+/// For example, if the configuration's integer encoding is `FixInt`, then `UseIntLen<u64>`
 /// will use the fixed-width u64 encoding.
-/// If the configuration's integer encoding is `VarInt`, then `UseInt<u64>` will use
+/// If the configuration's integer encoding is `VarInt`, then `UseIntLen<u64>` will use
 /// the variable-width u64 encoding.
 ///
 /// This is bincode's default behavior.
-pub struct UseInt<T>(PhantomData<T>);
+pub struct UseIntLen<T>(PhantomData<T>);
 
-impl<T, C: ConfigCore> SeqLen<C> for UseInt<T>
+impl<T, C: ConfigCore> SeqLen<C> for UseIntLen<T>
 where
     T: SchemaWrite<C> + for<'de> SchemaRead<'de, C>,
     T::Src: TryFrom<usize>,
@@ -98,11 +98,11 @@ where
 /// Fixed-width integer length encoding.
 ///
 /// Integers respect the configured byte order.
-pub struct FixInt<T>(PhantomData<T>);
+pub struct FixIntLen<T>(PhantomData<T>);
 
 macro_rules! impl_fix_int {
     ($type:ty) => {
-        impl<C: ConfigCore> SeqLen<C> for FixInt<$type> {
+        impl<C: ConfigCore> SeqLen<C> for FixIntLen<$type> {
             #[inline(always)]
             #[allow(irrefutable_let_patterns)]
             fn read<'de>(reader: &mut impl Reader<'de>) -> ReadResult<usize> {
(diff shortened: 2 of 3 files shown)


wincode/src/config/mod.rs
```diff
@@ -79,13 +138,81 @@ impl<const ZERO_COPY_ALIGN_CHECK: bool, const PREALLOCATION_SIZE_LIMIT: usize, L
...
+    ) -> Configuration<
+        ZERO_COPY_ALIGN_CHECK,
+        PREALLOCATION_SIZE_LIMIT,
+        LengthEncoding,
+        LittleEndian,
+        IntEncoding,
+    > {
+        generate()
+    }
+
```
kskalski: should we add `with_target_native_endian` using https://doc.rust-lang.org/reference/conditional-compilation.html#target_endian ?

it would be a nice feature as it would ensure zero copy, but the downside is non-static format for the same code dependent on platform it's compiled on

wincode/src/config/mod.rs
```diff
@@ -79,13 +138,81 @@ impl<const ZERO_COPY_ALIGN_CHECK: bool, const PREALLOCATION_SIZE_LIMIT: usize, L
...
+        self,
+    ) -> Configuration<
+        ZERO_COPY_ALIGN_CHECK,
+        PREALLOCATION_SIZE_LIMIT,
+        LengthEncoding,
+        LittleEndian,
+        IntEncoding,
+    > {
+        generate()
+    }
```
kskalski, line 179: should we provide `with_target_platform_endian`? The downside is that the same code would use different format depending on how it's compiled, but could be convenient for users that just operate exchanging data on the same machine

wincode/src/schema/int_encoding.rs
```diff
@@ -0,0 +1,240 @@
...
+    /// Decode a `i128` value from the reader.
+    fn decode_i128<'de>(reader: &mut impl Reader<'de>) -> ReadResult<i128>;
+}
+
+/// Fixed width integer encoding.
+///
+/// For all integer types, will encode `to_<byte_order>_bytes` and decode
+/// using the corresponding `from_<byte_order>_bytes` method.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct FixInt;
```
kskalski, line 212: it's confusing that there is now int_encoding::FixInt and len::FixInt<T>, we should call one of them differently, possibly the len type should be like `SeqLenFixInt<T>` though more backwards compatible to name new type differently, maybe `FixIntEncoding`

wincode/src/config/mod.rs
```diff
@@ -79,13 +138,81 @@ impl<const ZERO_COPY_ALIGN_CHECK: bool, const PREALLOCATION_SIZE_LIMIT: usize, L
...
+    /// Use little-endian byte order.
+    ///
+    /// Default is [`LittleEndian`].
+    pub const fn with_little_endian(
+        self,
+    ) -> Configuration<
+        ZERO_COPY_ALIGN_CHECK,
+        PREALLOCATION_SIZE_LIMIT,
+        LengthEncoding,
+        LittleEndian,
```
kskalski, line 175: Ideally the current type's generic parameters should use distinct naming from external types that are instances of required traits, maybe add some suffix like "Param" or `T` to the generic parameters?

wincode/src/config/mod.rs
cpubot: Seems reasonable to me!

wincode/src/schema/int_encoding.rs
cpubot: Well, they are the same encoding, just for different parts of the process (general integer encoding vs length encoding specific override). Perhaps the `len` module could just define its implementation on `int_encoding::FixInt` and re-export it. What do you think?

wincode/src/config/mod.rs
cpubot: I'm not sure I understand the problem here

wincode/src/config/mod.rs
cpubot: 55e55ad286e71a89762afc384afd08dffb11d9e9

wincode/src/config/mod.rs
kskalski: just a slight readability nit - I'm used to distinguishing generic type parameters from actual concrete types such that `Type::foo` is clearly a trait call or specific function call. Here it would help to see which argument is being passed on from current type's signature and which is a concrete type:
```
    Configuration<
        ZERO_COPY_ALIGN_CHECK_PARAM,
        PREALLOCATION_SIZE_LIMIT_PARAM,
        LengthEncParam,
        ByteOrder,
        IntEncParam,
    >
```
Not a big improvement, I leave it up to you if you can think a better naming schema for type parameters could be of use here.

wincode/src/schema/int_encoding.rs
kskalski: `int_encoding::FixInt` is a universal specification of encoding any ints, while `len::FixInt<T>` is a specific selection to encode lengths as T using `int_encoding::FixInt`, so basically in len we have two knobs:
* range of integers allowed as len
* how to encode them (`IntEncoding` trait)
I suppose you can't get away from having a type for this by re-export. We could however have a type like
`SeqLenSpec<IntEncoding, IntDomain>` and maybe even make the provided impls be more generic (they would support `SeqLen` for any user-provided` IntEncoding`)

cpubot: `FixInt` in both cases is a zero-sized type, and I don't anticipate that changing. Given that using `FixInt` in both the `SeqLen` case and the `IntEncoding` case implies the same encoding semantics (fixed-width with configured byte-order), I think it's reasonable to use that same ZST marker for both `IntEncoding` and `FixInt` implementations.

> We could however have a type like SeqLenSpec<IntEncoding, IntDomain> and maybe even make the provided impls be more generic (they would support SeqLen for any user-provided IntEncoding)

This behavior is already supported using [`UseLen`](https://github.com/anza-xyz/wincode/blob/5dc7c1b8baf09327f6da5607423974d62e64d492/wincode/src/len.rs#L60). `UseLen` will use the configured `IntEncoding` for sequence lengths, and it is the [default](https://github.com/anza-xyz/wincode/blob/5dc7c1b8baf09327f6da5607423974d62e64d492/wincode/src/len.rs#L147-L148).

wincode/src/config/mod.rs
cpubot: Okay, I see what you're saying. I don't have a problem with the current implementation, going to leave as is.

wincode/src/schema/int_encoding.rs
cpubot: Actually, I'm mistaken -- we can't use the same type for both since `FixInt` in `len` takes a `T` (the integer type).

Maybe we call them `FixIntLen` and `FixIntEncoding`? Or we can leave it up to users to create their own aliases if they import both.

kskalski: Yap, that was may point - they are not just separately defined structs, but just different since one also contains the len's type, not just encoding.
`FixIntLen` sounds good.

cpubot: Done: 8931fd8d25b394e34f49b48435324ade160af891

PR #87, opened 2026-01-20
Remove broken `Writer` implementation for `[u8]`

tobias-nd:
`impl Writer for [u8]` is broken, it mangles the serialization of the passed data by restarting serialization at the beginning of the provided buffer for each item. This means, for example, that `&str` gets serialized by serializing a length as little-endian `u64`, and then overwriting that length field with the actual string.

The bug is hidden due to another bug that prevents actually using the `Writer` implementation on `[u8]` -- all methods receiving a `Writer` take `&mut impl Writer` instead of `&mut (impl Writer + ?Sized)` excluding the `Writer` implementation for `[u8]`, as `[u8]` is not `Sized`.

Search-replacing using `sed -i 's/mut impl Writer/mut (impl Writer+?Sized)/g' **.rs`, one can observe the bug using the following proof of concept:

```rust
use std::io;
use std::io::Write as _;
use wincode::SchemaWrite as _;

fn main() {
    let mut buffer = [0xbb; 8];
    <&str>::write(&mut buffer[..], &"aaa").unwrap();
    io::stdout().write_all(&buffer).unwrap();
}
```

Hexdump of output:
```
00000000  61 61 61 00 00 00 00 00  bb bb bb bb bb bb bb bb  |aaa.............|
00000010
```

### DIFF
--- a/wincode/src/io/slice.rs
+++ b/wincode/src/io/slice.rs
@@ -375,16 +375,6 @@ fn advance_slice_mut_checked<'a, T>(
     Ok(dst)
 }
 
-/// Get a slice of `len` bytes for writing returning an error if the input slice does not have
-/// at least `len` bytes remaining.
-#[inline]
-fn get_slice_mut_checked<T>(input: &mut [T], len: usize) -> WriteResult<&'_ mut [T]> {
-    let Some((dst, _)) = input.split_at_mut_checked(len) else {
-        return Err(write_size_limit(len));
-    };
-    Ok(dst)
-}
-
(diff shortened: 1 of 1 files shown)


kskalski:
Right, I wasn't able to use the added implementations for `[u8]` due to sizing constraint, so we came up with the plan to remove `&mut`.
One way or the other those impls are a trap now and should be gone.

kskalski (dismissed):

tobias-nd:
Fixed the tests, I hadn't run with `--all-features`.

PR #88, opened 2026-01-20
Relax `SchemaRead::read` and `SchemaWrite::write` parameter types

tobias-nd:
There's no reason to require a "`Sized`" implementor, i.e. one that could be moved to the stack.

This restriction has maybe hidden the issue fixed in #87, as the `Writer` implementation could not be tested.

### DIFF
--- a/wincode-derive/src/schema_read.rs
+++ b/wincode-derive/src/schema_read.rs
@@ -176,7 +176,7 @@ fn impl_struct(
 /// impl<'de> SchemaRead<'de> for Message {
 ///     type Dst = Message;
 ///
-///     fn read(reader: &mut impl Reader<'de>, dst: &mut MaybeUninit<Self::Dst>) -> ReadResult<()> {
+///     fn read(reader: &mut (impl Reader<'de> + ?Sized), dst: &mut MaybeUninit<Self::Dst>) -> ReadResult<()> {
 ///         // Some more complicated logic not capturable by the macro...
 ///         let mut body = MaybeUninit::<Body>::uninit();
 ///         // Project a mutable MaybeUninit<Header> from the MaybeUninit<Body>.
@@ -390,7 +390,7 @@ fn impl_struct_extensions(args: &SchemaArgs, crate_name: &Path) -> Result<TokenS
 
             /// Read a value from the reader into the maybe uninitialized field.
             #[inline]
-            #vis fn #read_field_ident <'de>(&mut self, reader: &mut impl Reader<'de>) -> ReadResult<&mut Self> {
+            #vis fn #read_field_ident <'de>(&mut self, reader: &mut (impl Reader<'de> + ?Sized)) -> ReadResult<&mut Self> {
                 // SAFETY:
                 // - `self.inner` is a valid reference to a `MaybeUninit<#builder_dst>`.
                 // - We return the field as `&mut MaybeUninit<#target>`, so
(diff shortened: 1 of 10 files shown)


kskalski:
Yeah, I think changing parameters into `impl Writer` and `impl Reader` is the ultimate and better solution.

tobias-nd:
> @kskalski and I had discussed updating and `SchemaWrite` and `SchemaRead` to take an `impl Writer` and `impl Reader<'de>` rather than `&mut impl Writer` and `&mut impl Reader<'de>`. Additionally we'd provide blanket implementations
> 
> * `Writer for &mut T where T: Writer`
> 
> * `Reader<'a> for &mut T where T: Reader<'a>`
> 
> 
> This may be a more general solution

This has the unfortunate side-effect that every time you need to reborrow a reader/writer, you add one more layer of pointer indirection (which is only hopefully optimized out).

I.e. if you then derive `SchemaWrite` for a struct, you're going to need a writer multiple times, so you need to borrow the passed in writer using `&mut writer`. In the end, if you're a couple of levels down, you're going to have a `&mut &mut &mut &mut &mut Vec<u8>`, which requires inlining to optimize out, otherwise you're going to have code bloat from monomorphizing `Writer` implementations for all these types and runtime cost from dereferencing pointers several times where only one dereference would be required before this change.

kskalski:
Clearly something to watch for and check the generated code with such change. Zach will have a better intuition out of hand, but personally I doubt this would be an issue:
* except for maybe when using Box/`dyn` the real type of passed value is known and `&mut &mut x` conversion to just plain pointer to `x` is obvious and usually necessary for compiler to do 
* `std::io::Write` and `std::io::Read` do have [forwarding implementations](https://doc.rust-lang.org/src/std/io/impls.rs.html#13-90), so this would fall into "typical" approach
* we already have a bunch of places that take `&mut` of things or operate on `&mut &mut` - I think the expected behavior is that compiler always simplifies that. `wincode` relies *a lot* on inlining, so checking the generated code is the ultimate necessity...

As for this PR, could you remind us what is the actual code snippet that it enables?

My goal originally was to use `wincode::serialize_into(mut_slice_with_proper_len, value)`

From a more general point of view - I wonder if we should have clearer distinction of APIs that move the pointer/cursor vs such that just store data at known offsets - I would argue that quite common use-case for wincode is operating on fully known sizes and positions, so for best performance we rely on compiler to prune any code that does accounting. 
For example: when serializing `(u64, u64)` with `FixInt` into `[0u8; 16]` we don't need to create a slice pointing to the array, serialize first value letting it update the slice pointer with +8 bytes, then serialize the second value into that updated slice, again having that operation also update the slice - we can instead just serialize into 0th and 8th offsets without any intermediate updates of a slice.

tobias-nd:
> As for this PR, could you remind us what is the actual code snippet that it enables?

Probably nothing, only hypothetical types that don't implement `Sized` but implement `Reader`/`Writer`. With the implementations in #87 gone, there are no actual types remaining. I guess I can close this.

tobias-nd:
> * except for maybe when using Box/`dyn` the real type of passed value is known and `&mut &mut x` conversion to just plain pointer to `x` is obvious and usually necessary for compiler to do
> 
> * `std::io::Write` and `std::io::Read` do have [forwarding implementations](https://doc.rust-lang.org/src/std/io/impls.rs.html#13-90), so this would fall into "typical" approach
> 
> * we already have a bunch of places that take `&mut` of things or operate on `&mut &mut` - I think the expected behavior is that compiler always simplifies that. `wincode` relies _a lot_ on inlining, so checking the generated code is the ultimate necessity...

You're right, one would need to look at the actual generated code to understand whether this is a problem for the compiler or not.

PR #92, opened 2026-01-21
ci: add release.yml

yihau:
#### Update release process

This PR introduces a new release workflow:

1. Create a release PR that bumps the crate version and updates the changelog. For example, to publish wincode v0.3.0, open a PR that updates Cargo.toml to 0.3.0 and adds the corresponding changelog.
2. After the PR is merged, manually trigger the release pipeline and select the target crate and commit hash.
3. The release will proceed after manual approval.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -60,7 +60,7 @@ jobs:
 
       - name: Cargo publish dry run
         run: |
-          cargo publish -p ${CRATE_NAME} --dry-run
+          cargo publish -p "${CRATE_NAME}" --dry-run
         env:
           CRATE_NAME: ${{ inputs.crate }}
 


.github/workflows/release.yml
```diff
@@ -0,0 +1,93 @@
...
+jobs:
+  check:
+    runs-on: ubuntu-latest
+    outputs:
+      tag: ${{ steps.meta.outputs.tag }}
+      ref: ${{ steps.meta.outputs.ref }}
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          ref: ${{ inputs.ref }}
```
nbelenkov, line 29: checkout action keeps the credential until the end of the whole action in the repo, so better to specify to discard it straight away as we don't use it again

```suggestion
          ref: ${{ inputs.ref }}
          persist-credentials: false
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,93 @@
...
+        id: app-token
+        with:
+          app-id: ${{ vars.APP_ID }}
+          private-key: ${{ secrets.PRIVATE_KEY }}
+
+      - name: Set git config
+        run: |
+          git config --global user.email "github-actions@github.com"
+          git config --global user.name "github-actions"
+          git config --global url."https://x-access-token:${{ steps.app-token.outputs.token }}@github.com/".insteadOf https://github.com/
```
nbelenkov: its a bad practice to pass the variables directly

```suggestion
          git config --global url."https://x-access-token:${STEPS_APP_TOKEN_OUTPUTS_TOKEN}@github.com/".insteadOf https://github.com/
        env:
          STEPS_APP_TOKEN_OUTPUTS_TOKEN: ${{ steps.app-token.outputs.token }}
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,93 @@
...
+
+      - uses: actions/checkout@v6
+        with:
+          token: ${{ steps.app-token.outputs.token }}
+          persist-credentials: false
+          ref: ${{ needs.check.outputs.ref }}
+
+      - run: |
+          tag="${{ needs.check.outputs.tag }}"
```
nbelenkov:
```suggestion
          tag="${NEEDS_CHECK_OUTPUTS_TAG}"
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,93 @@
...
+          git tag -a "$tag" -m "Release $tag"
+          git push --tags
```
nbelenkov, line 107:
```suggestion
          git push --tags
        env:
          NEEDS_CHECK_OUTPUTS_TAG: ${{ needs.check.outputs.tag }}

```

.github/workflows/release.yml
```diff
@@ -0,0 +1,93 @@
...
+
+          tag="${crate}@v${version}"
+          ref="$(git rev-parse HEAD)"
+
+          echo "tag=$tag" >> "$GITHUB_OUTPUT"
+          echo "ref=$ref" >> "$GITHUB_OUTPUT"
+
+      - name: Check tag does not exist
+        run: |
+          tag="${{ steps.meta.outputs.tag }}"
```
nbelenkov:
```suggestion
          tag="${STEPS_META_OUTPUTS_TAG}"
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,93 @@
...
+          if git ls-remote --exit-code --tags origin "refs/tags/$tag" >/dev/null 2>&1; then
+            echo "Tag already exists: $tag. Please bump the version first (or delete the existing tag) and retry."
+            echo "Tag exists: $tag" >> "$GITHUB_STEP_SUMMARY"
+            exit 1
+          fi
+          echo "Tag: ${{ steps.meta.outputs.tag }}" >> "$GITHUB_STEP_SUMMARY"
```
nbelenkov:
```suggestion
          echo "Tag: ${STEPS_META_OUTPUTS_TAG}" >> "$GITHUB_STEP_SUMMARY"
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,93 @@
...
+          echo "Ref: ${{ steps.meta.outputs.ref }}" >> "$GITHUB_STEP_SUMMARY"
```
nbelenkov:
```suggestion
          echo "Ref: ${STEPS_META_OUTPUTS_REF}" >> "$GITHUB_STEP_SUMMARY"
        env:
          STEPS_META_OUTPUTS_TAG: ${{ steps.meta.outputs.tag }}
          STEPS_META_OUTPUTS_REF: ${{ steps.meta.outputs.ref }}
```

nbelenkov:
We are currently just making a tag but not a full release, so at the end of the flow after publish, we need something like:

```
- name: Create GitHub Release
  run: |
    gh release create "${TAG}" \
      --title "${TAG}" \
      --generate-notes
  env:
    TAG: ${{ needs.check.outputs.tag }}
    GITHUB_TOKEN: ${{ steps.app-token.outputs.token }}

```

.github/workflows/release.yml
yihau: [6120b61](https://github.com/anza-xyz/wincode/pull/92/commits/6120b6195f3f462ac02629b7957ea72746d045f2)

.github/workflows/release.yml
yihau: [d30c6a7](https://github.com/anza-xyz/wincode/pull/92/commits/d30c6a750edfc7d77fa3a1ced3d970b5736b8630)

.github/workflows/release.yml
yihau: [d30c6a7](https://github.com/anza-xyz/wincode/pull/92/commits/d30c6a750edfc7d77fa3a1ced3d970b5736b8630)

.github/workflows/release.yml
yihau: https://github.com/anza-xyz/wincode/pull/92/commits/3ffce6209c305db4ab46daf61a0c7597dddff546

.github/workflows/release.yml
yihau: https://github.com/anza-xyz/wincode/pull/92/commits/3ffce6209c305db4ab46daf61a0c7597dddff546

.github/workflows/release.yml
yihau: https://github.com/anza-xyz/wincode/pull/92/commits/3ffce6209c305db4ab46daf61a0c7597dddff546

yihau:
> We are currently just making a tag but not a full release, so at the end of the flow after ... (omit)

yeah, the idea looks good! however, I guess we can send it as a draft first so that we have time to review the release.

[6dc921c](https://github.com/anza-xyz/wincode/pull/92/commits/6dc921cede9b687ceb21fd77e2883fefab81affc)

.github/workflows/release.yml
```diff
@@ -0,0 +1,116 @@
...
+        id: app-token
+        with:
+          app-id: ${{ vars.APP_ID }}
+          private-key: ${{ secrets.PRIVATE_KEY }}
+
+      - name: Set git config
+        run: |
+          git config --global user.email "github-actions@github.com"
+          git config --global user.name "github-actions"
+          git config --global url."https://x-access-token:${GIHTUB_TOKEN}@github.com/".insteadOf https://github.com/
```
nbelenkov:
```suggestion
          git config --global url."https://x-access-token:${GITHUB_TOKEN}@github.com/".insteadOf https://github.com/
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,116 @@
...
+
+      - name: Create Github Release
+        run: |
+          gh release create "${TAG}" --title "${TAG}" --generate-notes --draft
+        env:
+          TAG: ${{ needs.check.outputs.tag }}
+          GITHUB_TOKEN: ${{ steps.app-token.outputs.token }}
+
+      - name: Package crate
+        run: cargo package -p ruleset-playground-derive
```
nbelenkov:
```suggestion
        run: cargo package -p inputs.crate
```

nbelenkov:
We are basically there, just 2 things:

1. We don't currently build crates or run tests during release. I think it makes sense to add another job before publishing to check that everything builds + tests run and do provenance there
2. If we don't build, I would move provenance generation to just after checkout and before tag creation for logical flow, it also means that `.crate` file already exists and doesn't need to be generated again by `cargo package`

Update:
One more thing, is that we should probably do a full release not a draft one, as I am not sure what draft release gives us in the current case, as the crate has been published already

.github/workflows/release.yml
yihau: ah... good catch! thank you: [7e7446f](https://github.com/anza-xyz/wincode/pull/92/commits/7e7446fd68d1f8052838709d7b460c42bb6140f7)

yihau:
> We don't currently build crates or run tests during release. I think it makes sense to add another job before publishing to check that everything builds + tests run and do provenance there

okay. just added a publish dry run (931d925). let's address other ci steps in separate PRs. for the provenance step, I think maybe it's better to have it after approve 🤔 

> If we don't build, I would move provenance generation to just after checkout and before tag creation for logical flow, it also means that .crate file already exists and doesn't need to be generated again by cargo package

I moved it earlier: 0b6d59d


> Update:
One more thing, is that we should probably do a full release not a draft one, as I am not sure what draft release gives us in the current case, as the crate has been published already

okay!  d194886

.github/workflows/release.yml
```diff
@@ -0,0 +1,127 @@
...
+
+          tag="${crate}@v${version}"
+          ref="$(git rev-parse HEAD)"
+
+          echo "tag=$tag" >> "$GITHUB_OUTPUT"
+          echo "ref=$ref" >> "$GITHUB_OUTPUT"
+
+      - name: Check tag does not exist
+        run: |
+          if git ls-remote --exit-code --tags origin "refs/tags/$TAG" >/dev/null 2>&1; then
```
nbelenkov: small nit, I am pretty sure this line will error as `$TAG` is not defined, and only `$tag` is. We probably should just stick to either `$TAG` or `$tag`

nbelenkov:
Thanks for addressing all of the above, it all basically looks good to me

.github/workflows/release.yml
yihau: yeah, this PR has been changing a lot. I should do a final review myself. Thanks for the reminder! I've pushed 3 cleanup and 2 ux improvements

unify variables style:
- [daf2b27](https://github.com/anza-xyz/wincode/pull/92/commits/daf2b27280391392ccd66b9996ae902248f21cd8)
- [f3810ee](https://github.com/anza-xyz/wincode/pull/92/commits/f3810ee68f29f298baf8a5dc21b440a5b672a6bf)
- [3ad402e](https://github.com/anza-xyz/wincode/pull/92/commits/3ad402e5ed527ce50fe5af583330b4806e61cd8c)

add default input.ref: [80e4891](https://github.com/anza-xyz/wincode/pull/92/commits/80e48918ebd5b66f474bb9449c85d7ac505521d4)

select empty input.crate by default: [46d274f](https://github.com/anza-xyz/wincode/pull/92/commits/46d274f07e6df62126c80b0814104d276b4a7e5e)

.github/workflows/release.yml
```diff
@@ -0,0 +1,135 @@
...
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          ref: ${{ inputs.ref }}
+          persist-credentials: false
+
+      - name: Compute metadata
+        id: meta
+        run: |
+          crate="${{ inputs.crate }}"
```
nbelenkov: We don't need one anymore I think
```suggestion
```

yihau: let me do it rn 🥶

yihau: oh.. actually it's extracted. just need to remove the redundant line. thank you!

.github/workflows/release.yml
```diff
@@ -0,0 +1,135 @@
...
+          ref: ${{ needs.check.outputs.ref }}
+
+      - run: |
+          git tag -a "${TAG}" -m "Release ${TAG}"
+          git push --tags
+        env:
+          TAG: ${{ needs.check.outputs.tag }}
+
+      - name: Package crate
+        run: cargo package -p ${CRATE_NAME}
```
nbelenkov:
```suggestion
        run: cargo package -p "${CRATE_NAME}"
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,135 @@
...
+            echo "Tag already exists: ${TAG}. Please bump the version first (or delete the existing tag) and retry."
+            echo "Tag exists: ${TAG}" >> "${GITHUB_STEP_SUMMARY}"
+            exit 1
+          fi
+        env:
+          TAG: ${{ steps.meta.outputs.tag }}
+
+      - name: Cargo publish dry run
+        run: |
+          cargo publish -p ${CRATE_NAME} --dry-run
```
nbelenkov:
```suggestion
          cargo publish -p "${CRATE_NAME}" --dry-run
```

.github/workflows/release.yml
nbelenkov: So, up to you if we do an empty default crate, but if we do, we should have a check at the start that fails, if we have it as default, otherwise I am not 100% where it will actually break

```
 name: Validate crate input
  run: |
    if [ -z "${CRATE_NAME}" ]; then
      echo "No crate selected. Please choose a crate to release."
      exit 1
    fi
  env:
    CRATE_NAME: ${{ inputs.crate }}
```

nbelenkov: otherwise looks good to me

yihau: yeah, I think we're good because when we use the empty option, it will prompt us:

Screenshot 2026-01-27 at 20 05 27

nbelenkov: ahhh, its required, very good

nbelenkov:
Update the source to always point to a pinned version of `wincode-derive` makes more sense In my eyes and makes the setup easier

PR #94, opened 2026-01-22
fix:cache base_ptr in struct builder

dvansari65:
perf: cache base_ptr in struct builder to eliminate repeated as_mut_ptr() calls

Optimizes generated builder structs by caching the MaybeUninit buffer pointer,
reducing field access cost !

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -432,6 +432,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "paste"
+version = "1.0.15"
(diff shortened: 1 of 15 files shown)


tobias-nd:
> to eliminate repeated as_mut_ptr() calls

`as_mut_ptr()` [is just a cast](https://doc.rust-lang.org/1.92.0/src/core/mem/maybe_uninit.rs.html#643-646), and thus should be free.

Have you verified that this impacts performance?

PR #95, opened 2026-01-22
Adding security section to README

nbelenkov:


nbelenkov:
this can be merged, I don't think I have permission though @cpubot

PR #100, opened 2026-01-26
Add `uuid` support

### LINKED ISSUE #75: Uuid type support
Hello, is it possible to add support for uuid crate Uuid type?

cpubot:
Adds the following features:
- `uuid`
- `uuid-serde-compat`

`Uuid`, as defined in the `uuid` crate, is simply a newtype wrapper over `[u8; 16]`. So, by default, we treat it like a fixed sized array, which allows it to satisfy static and zero-copy.

There is a discrepancy between treating it like a fixed sized array and how `uuid`'s `serde` feature encodes a `Uuid`. `uuid`'s `serde` implementation calls `serializer.serialize_bytes`, which treats the bytes as a _slice_. This means it encodes an additional length field in the serialized payload. To accommodate users coming from a `serde` context (e.g., using `bincode` 1.x or `bincode` 2.x with serde compatibility enabled), I added a `uuid-serde-compat` feature that will include that additional length encoding.

Closes #75

### DIFF
--- a/wincode/src/schema/external/uuid.rs
+++ b/wincode/src/schema/external/uuid.rs
@@ -9,7 +9,7 @@ use {
         io::{Reader, Writer},
         schema::{SchemaRead, SchemaWrite},
     },
-    core::mem::MaybeUninit,
+    core::mem::{transmute, MaybeUninit},
     uuid::Uuid,
 };
 
@@ -34,7 +34,10 @@ unsafe impl<'de, C: Config> SchemaRead<'de, C> for Uuid {
         let bytes = *reader.fill_array::<{ size_of::<Uuid>() }>()?;
         // SAFETY: `fill_array` guarantees we get exactly `size_of::<Uuid>()` bytes.
         unsafe { reader.consume_unchecked(size_of::<Uuid>()) };
-        dst.write(Uuid::from_bytes(bytes));
+        // SAFETY: `Uuid` is a `#[repr(transparent)]` newtype over `uuid::Bytes` (`[u8; 16]`).
+        let dst =
+            unsafe { transmute::<&mut MaybeUninit<Uuid>, &mut MaybeUninit<uuid::Bytes>>(dst) };
+        dst.write(bytes);
         Ok(())
     }
 }


wincode/src/schema/external/uuid.rs
```diff
@@ -0,0 +1,142 @@
...
+
+#[cfg(feature = "uuid")]
+unsafe impl<'de, C: Config> SchemaRead<'de, C> for Uuid {
+    type Dst = Uuid;
+
+    // serde serializes byte slices as a length-prefixed array.
+    // As such, it must fall back to TypeMeta::Dynamic when `uuid-serde-compat`
+    // is enabled. Otherwise, we can enable static and zero-copy.
+    #[cfg(not(feature = "uuid-serde-compat"))]
+    const TYPE_META: TypeMeta = TypeMeta::Static {
```
kskalski, line 24: isn't it possible to provide Static meta with known size based on C's seq len encoding?

cpubot: `SeqLen` doesn't have a `TYPE_META` const, so unfortunately not. Though, I think it'd be reasonable and simple to implement. After which, we could adjust this to determine `Static` based on the length encoding

kskalski: I guess we would need to somehow detect if it uses FixInt...

cpubot: Yeah, I think this would be straight forward to implement. `FixInt` can always declare `TypeMeta::Static`, and `UseInt` can key in on the `STATIC` and `ZERO_COPY` constants of `IntEncoding` -- should be able to make it work

cpubot: That being said, this is a bit of a one-off case. I don't think it necessarily makes sense generally for `SeqLen` to have a `TypeMeta`. `SeqLen` is deliberately separated from `SchemaWrite` / `SchemaRead` due to additional validations like preallocation checks, and I can't think of any other case outside of this specific serde-compatibility case where it adds a length prefix to an array of bytes with a statically known size.

wincode/src/schema/external/uuid.rs
```diff
@@ -0,0 +1,142 @@
...
+
+    #[inline]
+    fn read(reader: &mut impl Reader<'de>, dst: &mut MaybeUninit<Self::Dst>) -> ReadResult<()> {
+        // serde serializes byte slices as a length-prefixed array.
+        #[cfg(feature = "uuid-serde-compat")]
+        let _len = C::LengthEncoding::read(reader)?;
+        let bytes = *reader.fill_array::<{ size_of::<Uuid>() }>()?;
+        // SAFETY: `fill_array` guarantees we get exactly `size_of::<Uuid>()` bytes.
+        unsafe { reader.consume_unchecked(size_of::<Uuid>()) };
+        dst.write(Uuid::from_bytes(bytes));
```
kskalski: hope this is inlined by the compiler (instead of populating array on stack and copying it into `dst`)

cpubot: Yeah, can't really get around this due to the inner type not being public -- would need to implement within the `uuid` crate to have direct access (which I did attempt, but was rejected due to nascence of `wincode`). My sense is that it'll likely be inlined. But, if we wanted to force it, we could transmute `Dst` to `[u8; 16]` and write directly. [It is `#[repr(transparent)]`](https://docs.rs/uuid/1.20.0/src/uuid/lib.rs.html#432), so that could technically be acceptable.

wincode/src/schema/external/uuid.rs
kskalski: I see, because of the recursive structure of encoding we can't utilize seq len's static size in 
```
struct A {
   x: u64,
   seq: &[u8]
}
```
such that x+seq's len are zero-copied, right? Len is always encoded with the actual payload, which is dynamic if len was necessary, so its static size won't allow the whole collection to be static anyway...

cpubot: > Len is always encoded with the actual payload, which is dynamic if len was necessary, so its static size won't allow the whole collection to be static anyway

Yeah, exactly. Even if it's the case that the `SeqLen` impl itself is technically static, it doesn't buy us anything in the general case because the length value is relative to the following sequence it describes. In other words, we generally need to read the length _first_ before we can proceed to decode the sequence.

wincode/src/schema/external/uuid.rs
kskalski: I would say if we take all the effort to add uuid dependency and create custom implementation and we have all this APIs taking uninit, then we should close the loop and make sure there are no unnecessary copies.

cpubot: Yep, works for me. b51fa203990ad786d16e1a1894ae14b2ade6ca81

PR #102, opened 2026-01-27
Make sure to fail serialization if deserialization would error

### LINKED ISSUE #90: Behavior difference from `bincode` in large arrays, roundtrip not guaranteed
When deserializing, `wincode` tries to make sure that allocated buffers are not too large, in order to prevent simple DoS by inputs pretending to be large. However, when serializing, `BincodeLen::write` does not check whether the serialized data conforms to size limitations. This means that serialization can succeed despite outputting data that will fail to be deserialized.

Proof of concept showing failed roundtrip of large `Vec<u8>` (4 MiB + 1 B):
```rust
fn main() {
    let large: Vec<u8> = vec![0; 4 * 1024 * 1024 + 1];
    let serialized = wincode::serialize(&large).unwrap();
    println!("{:?}", wincode::deserialize::<Vec<u8>>(&serialized));
}
```

```
Err(PreallocationSizeLimit { needed: 4194305, limit: 4194304 })
```

This is also a behavior differential from the `bincode` crate, which does not implement such limits by default:

```rust
use bincode::config::legacy;
use bincode::config::standard;

fn main() {
    let large: Vec<u8> = vec![0; 4 * 1024 * 1024 + 1];
    let serialized = wincode::serialize(&large).unwrap();
    println!("{:?}", wincode::deserialize::<Vec<u8>>(&serialized));
    println!("{:?}",
        bincode::decode_from_slice::<Vec<u8>, _>(&serialized, standard())
            .map(|_| ())
    );
    println!("{:?}", 
        bincode::decode_from_slice::<Vec<u8>, _>(&serialized, legacy())
            .map(|_| ())
    );
}
```

```
Err(PreallocationSizeLimit { needed: 4194305, limit: 4194304 })
Ok(())
Ok(())
```

### Resolution

It would be possible to check the written length for deserializability, in `BincodeLen::write`. Alternatively, one could accept larger inputs but not preallocate as much, for `bincode` compatibility.

tobias-nd:
This crate makes sure that while deserializing, no large buffers are allocated, returning an error instead. This error is now also returned while serializing.

Also document this incompatibility with bincode.

Fixes #90.

### DIFF
--- a/wincode/src/lib.rs
+++ b/wincode/src/lib.rs
@@ -162,6 +162,10 @@
 //!   default configuration, provided your [`SchemaWrite`] and [`SchemaRead`] schemas and
 //!   [`containers`] match the layout implied by your `serde` types.
 //! - Length encodings are pluggable via [`SeqLen`](len::SeqLen).
+//! - Unlike `bincode`, this crate will fail to serialize or deserialize large
+//!   dynamic data structures by default, but this can be configured. This is
+//!   done for security and performance, as it allows to preallocate these data
+//!   structures safely.
 //!
 //! # Zero-copy deserialization
 //!


tobias-nd:
I fixed formatting, the missing symmetry on the `VecDeque` impl and the failing CI tests.

PR #93, opened 2026-01-21
Mark IntEncoding unsafe

cpubot:
There are safety invariants that must be upheld for `STATIC` and `ZERO_COPY` constants. The `size_of_*` functions similarly have important invariants that must be upheld.

PR #86, opened 2026-01-20
Mark `SchemaRead` and `SchemaWrite` trait as `unsafe`

tobias-nd:
Users can still derive these traits without mentioning `unsafe`.

Fixes #79. Fixes #80. Fixes #81.

PR #105, opened 2026-01-28
chore: prepare wincode 0.2.6

yihau:
for releasing wincode 0.2.6


tobias-nd:
wincode has had breaking changes (e.g. #86, #93, #99), so it'd probably be better to release this as 0.3.0 instead.

yihau:
thank you @tobias-nd

PR #108, opened 2026-01-28
Implement TagEncoding trait

cpubot:
This PR adds support for setting a global enum discriminant encoding via global configuration.

Adds:
- `with_tag_encoding<T>` to config 
- `TagEncoding` trait

The immediate use-case for this feature is to allow anchor to swap borsh with wincode in the next major version.

[spec](https://github.com/near/borsh?tab=readme-ov-file#specification)

Which will effectively be
```rs
DefaultConfig::new()
    .with_tag_encoding::<u8>()
    .with_length_encoding::<FixInt<u32>();
```

## Implementation notes

There isn't necessarily a "totally clean" way to pull this off from an API perspective. In particular, enum implementations perform matching and writing of integer literals in their `SchemaRead` and `SchemaWrite` implementations, respectively.

E.g.,
https://github.com/anza-xyz/wincode/blob/93f3312e052e4fbcff693e335b12d42a200a44d1/wincode/src/schema/impls.rs#L716-L720
https://github.com/anza-xyz/wincode/blob/93f3312e052e4fbcff693e335b12d42a200a44d1/wincode/src/schema/impls.rs#L759-L768

There is no way to tell the compiler that a trait is only implemented by a closed set of integer types. As such, it's not possible to match the result of a generic `SchemaRead::get` against literal integers. And similarly it is not possible to call a generic `SchemaWrite::write` with integer literals. 

This leaves two possible solutions:
1. Normalize discriminants to single integer type (e.g., `u32`)
2. Make a sealed `TagEncoding` trait only implemented by a closed set of integer types, and provide an `enum` that enumerates each supported type that is attached to `Config`.

On the surface, `2` feels like it may be cleaner from a code generation perspective. We can `match` on a `const enum` in `enum` implementations, and provide implementations that read and write the exact integer type specified in the config. However, one quickly runs into a serious code duplication issue. For example,
```rs
unsafe impl<'de, T, E, C: Config> SchemaRead<'de, C> for Result<T, E>
where
    T: SchemaRead<'de, C>,
    E: SchemaRead<'de, C>,
{
    type Dst = Result<T::Dst, E::Dst>;

        C::TagEncoding {
            TagEncoding::U8 => {
                let disc = <u8 as SchemaRead<C>>::get(reader)?;
                match disc {
                    0 => dst.write(Result::Ok(T::get(reader)?)),
                    1 => dst.write(Result::Err(E::get(reader)?)),
                    _ => return Err(invalid_tag_encoding(disc as usize)),
                };
            }
            TagEncoding::U16 => {
                let disc = <u16 as SchemaRead<C>>::get(reader)?;
                match disc {
                    0 => dst.write(Result::Ok(T::get(reader)?)),
                    1 => dst.write(Result::Err(E::get(reader)?)),
                    _ => return Err(invalid_tag_encoding(disc as usize)),
                };
            }
            // ...
        }
        Ok(())
    }
}
```

This could be solved with a clever `macro_rules!` implementation. But this adds complexity for downstream consumers and makes defining custom enum implementations more cumbersome.

As such, I've decided on option `1`, which is to normalize enum discriminants to a `u32`. This may seem unfortunate at first, as it may appear that we introduce unnecessary widening when reading and writing discriminants for integer's of width < 32 bits. However, after testing, I have observed that the compiler is reliably able to elide the `u32` conversion entirely in the generated assembly. In particular, because `TagEncoding` is ultimately monomorphized into concrete integer targets, and _all_ implementations we provide (including via the derive macros) provide integer literals for both reading and writing, the result of `try_from`/`try_into` calls are known at compile time. In the generated assembly, discriminants are loaded at their original width and compared directly to immediates.

While not perfect (I would prefer if there was some way to declare to the compiler that a given trait is always "matchable"  against integer literals), I think this route is preferable from a usability standpoint.  And given that the compiler is able to elide `u32` normalization entirely, we aren't paying for extra widening.

### DIFF
--- a/wincode/src/schema/tag_encoding.rs
+++ b/wincode/src/schema/tag_encoding.rs
@@ -6,7 +6,7 @@ use {
         io::Writer,
         SchemaRead, SchemaWrite, WriteResult,
     },
-    std::any::type_name,
+    core::any::type_name,
 };
 
 /// Tag encoding trait.


kskalski:
I'm a bit hesitant if this feature is really necessary given how the encoding can be specified per type, but surely it gives a way to start off with a better uniform encoding. And I guess it's also nice for scenarios when not using `derive` feature.
Might be worth giving a bit more justification why/when this is going to be useful in PR description.

kskalski (approved):

PR #109, opened 2026-01-29
Feat: Add `SchemaRead` and `SchemaWrite` implementations of `NonZero` types - NonZeroU8, NonZeroU16 ..... from `std::num`

adpthegreat:
This PR adds `SchemaRead` and `SchemaWrite` implementations for `NonZero` Types from the standard library using a macro 

Ported from the bincode [impls](https://git.sr.ht/~stygianentity/bincode/tree/trunk/item/src/enc/impls.rs#L11)

### DIFF
--- a/wincode/src/schema/mod.rs
+++ b/wincode/src/schema/mod.rs
@@ -3124,7 +3124,7 @@ mod tests {
         let result: ReadResult<SystemTime> = deserialize(&bytes);
         assert!(result.is_err());
     }
-    
+
     #[test]
     fn test_nonzero_types() {
         proptest!(proptest_cfg(), |(
@@ -3198,11 +3198,17 @@ mod tests {
         // Test that deserializing a zero value fails
         let zero_bytes = serialize(&0u32).unwrap();
         let result: ReadResult<NonZeroU32> = deserialize(&zero_bytes);
-        assert!(result.is_err(), "Deserializing zero should fail for NonZeroU32");
-        
+        assert!(
+            result.is_err(),
+            "Deserializing zero should fail for NonZeroU32"
+        );
+
         let zero_bytes = serialize(&0u64).unwrap();
         let result: ReadResult<NonZeroU64> = deserialize(&zero_bytes);
-        assert!(result.is_err(), "Deserializing zero should fail for NonZeroU64");
+        assert!(
+            result.is_err(),
+            "Deserializing zero should fail for NonZeroU64"
+        );
     }
 
     #[test]


wincode/src/schema/impls.rs
```diff
@@ -4,6 +4,10 @@ use std::{
     collections::{HashMap, HashSet},
     hash::Hash,
     time::{SystemTime, UNIX_EPOCH},
+    num::{
```
cpubot: Should use `core::num`

wincode/src/schema/mod.rs
```diff
@@ -2926,6 +2901,55 @@ mod tests {
         let result: ReadResult<SystemTime> = deserialize(&bytes);
         assert!(result.is_err());
     }
+    
+    #[test]
+    #[cfg(feature = "std")]
```
cpubot: This `tests` module is already gated by `std`

wincode/src/schema/mod.rs
```diff
@@ -2926,6 +2901,55 @@ mod tests {
...
+                                NonZeroUsize, NonZeroI8, NonZeroI16, NonZeroI32, NonZeroI64, NonZeroI128, NonZeroIsize);
+            let bincode_deserialized: NonZeroTuple = bincode::deserialize(&bincode_serialized).unwrap();
+            let schema_deserialized: NonZeroTuple = deserialize(&schema_serialized).unwrap();
+            prop_assert_eq!(value, bincode_deserialized);
+            prop_assert_eq!(value, schema_deserialized);
+        });
+    }
+
+    #[test]
+    #[cfg(feature = "std")]
```
cpubot: This `tests` module is already gated by `std`

cpubot (changes requested):
Looks good. Just remove the extra `cfg(feature = "std")` in tests and use `core::num`

adpthegreat:
Done

adpthegreat:
Rebased

wincode/src/schema/mod.rs
```diff
@@ -3090,6 +3094,53 @@ mod tests {
...
+            let bincode_deserialized: NonZeroTuple = bincode::deserialize(&bincode_serialized).unwrap();
+            let schema_deserialized: NonZeroTuple = deserialize(&schema_serialized).unwrap();
+            prop_assert_eq!(value, bincode_deserialized);
+            prop_assert_eq!(value, schema_deserialized);
+        });
+    }
+
+    #[test]
+    fn test_nonzero_option() {
+        proptest!(proptest_cfg(), |(value: Option<NonZeroU32>)| {
```
cpubot: I don't think this test is actually useful -- more useful would be checking that deserializing a 0 value results in an error

wincode/src/schema/mod.rs
```diff
@@ -3090,6 +3094,53 @@ mod tests {
...
+            nz_u32 in (1u32..=u32::MAX).prop_map(|v| NonZeroU32::new(v).unwrap()),
+            nz_u64 in (1u64..=u64::MAX).prop_map(|v| NonZeroU64::new(v).unwrap()),
+            nz_u128 in (1u128..=u128::MAX).prop_map(|v| NonZeroU128::new(v).unwrap()),
+            nz_usize in (1usize..=usize::MAX).prop_map(|v| NonZeroUsize::new(v).unwrap()),
+            nz_i8 in (i8::MIN..=-1i8).prop_union(1i8..=i8::MAX).prop_map(|v| NonZeroI8::new(v).unwrap()),
+            nz_i16 in (i16::MIN..=-1i16).prop_union(1i16..=i16::MAX).prop_map(|v| NonZeroI16::new(v).unwrap()),
+            nz_i32 in (i32::MIN..=-1i32).prop_union(1i32..=i32::MAX).prop_map(|v| NonZeroI32::new(v).unwrap()),
+            nz_i64 in (i64::MIN..=-1i64).prop_union(1i64..=i64::MAX).prop_map(|v| NonZeroI64::new(v).unwrap()),
+            nz_i128 in (i128::MIN..=-1i128).prop_union(1i128..=i128::MAX).prop_map(|v| NonZeroI128::new(v).unwrap()),
+            nz_isize in (isize::MIN..=-1isize).prop_union(1isize..=isize::MAX).prop_map(|v| NonZeroIsize::new(v).unwrap()),
```
cpubot: Why not the following?

```suggestion
            nz_u8: NonZeroU8,
            nz_u16: NonZeroU16,
            nz_u32: NonZeroU32,
            nz_u64: NonZeroU64,
            nz_u128: NonZeroU128,
            nz_usize: NonZeroUsize,
            nz_i8: NonZeroI8,
            nz_i16: NonZeroI16,
            nz_i32: NonZeroI32,
            nz_i64: NonZeroI64,
            nz_i128: NonZeroI128,
            nz_isize: NonZeroIsize,
```

cpubot (changes requested):

wincode/src/schema/mod.rs
adpthegreat: i was omitting zero in the test cases, because i thought that deserializing it would cause an error (like the previous test you recommended), thats why the test range is all negative numbers and positive numbers for that specific numerical type (u32, u64), intentioanlly skipping zero etc, anyways i followed what you recommended and it worked, the prop test suite probably handles it under the hood

wincode/src/schema/mod.rs
adpthegreat: yes i agree, i did something similar in the refcell implementation

Issue #112, opened 2026-01-30
Derive SchemaWrite on empty struct fails.


marcfir:
Having a struct like this
```rust
    #[derive(Debug, SchemaWrite)]
    struct ComplexType {

    }
```

gives me an
````
error[E0308]: mismatched types
   --> src/lib.rs
    |
    |     #[derive(Debug, SchemaWrite, Default, PartialEq)]
    |                     ^^^^^^^^^^^ expected `TypeMeta`, found `Option<_>`
    |
    = note: expected enum `wincode::TypeMeta`
               found enum `std::option::Option<_>`
    = note: this error originates in the derive macro `SchemaWrite` (in Nightly builds, run with -Z macro-backtrace for more info)
```
Is this intended?

adpthegreat:
@marcfir  i have recreated your error locally and pushed a fix, it was a problem in the derive implementation for `SchemaWrite` on structs , fixed in #114

marcfir:
Wow this was quick. Thx.

adpthegreat:
@cpubot you can close this issue now

cpubot:
Closed via #114

PR #114, opened 2026-01-30
Fix: fixed zero sized struct error

adpthegreat:
This PR fixes issue #112

The issue was the macros expansion in the derive implementation for SchemaWrite on struct 

```
 return (quote! {Ok(0)}, quote! {Ok(())}, quote! {None});
```

instead of 

```
 return (quote! {Ok(0)}, quote! {Ok(())}, quote! {
            TypeMeta::Static {
                size: 0,
                zero_copy: true,
            }
        });
```

Returning `None` instead of a `Type_meta` was what caused the error, hence you can see "expected `TypeMeta`, found `Option<_>`"   ....... in the error 

TEST OUTPUT
```
     Running unittests src/lib.rs (target/debug/deps/wincode-8988e63017e17882)

running 1 test
test schema::tests::test_empty_struct ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 160 filtered out; finished in 0.00s

     Running unittests src/lib.rs (target/debug/deps/wincode_derive-2b64f00eb5f22432)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 4 filtered out; finished in 0.00s

```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -935,24 +935,12 @@ dependencies = [
  "solana-short-vec",
  "thiserror",
  "uuid",
- "wincode-derive 0.3.1 (registry+https://github.com/rust-lang/crates.io-index)",
+ "wincode-derive",
 ]
 
 [[package]]
 name = "wincode-derive"
 version = "0.3.1"
-dependencies = [
- "darling",
- "proc-macro2",
- "quote",
- "syn",
-]
-
-[[package]]
-name = "wincode-derive"
-version = "0.3.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d89833c7743abd1a831b7eb0fdfd7210bd8bedc297435e9426aad9b63d7c7f77"
 dependencies = [
  "darling",
  "proc-macro2",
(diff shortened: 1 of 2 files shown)


adpthegreat:
Alright will do that

adpthegreat:
Done

adpthegreat:
I've made the changes to `SchemaRead `, also CI test fails because it still points to the `wincode-derive` from crates.io without the fix, running it with `wincode-derive = { path = "../wincode-derive", optional = true }` locally would work and show the fix indeed works

````
  Running unittests src/lib.rs (target/debug/deps/wincode-8988e63017e17882)

running 1 test
test schema::tests::test_empty_struct ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 160 filtered out; finished in 0.00s

     Running unittests src/lib.rs (target/debug/deps/wincode_derive-2b64f00eb5f22432)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 4 filtered out; finished in 0.00s
```

PR #121, opened 2026-01-31
wincode: Add rust version

febo:
### Problem

We need to set a MSRV version on `wincode` crate.

### Solution

This PR uses the rust-version set on the workspace on the `wincode` crate. It includes fixes for clippy nits and downgrade of the `criterion` dev dependency.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,13 +24,13 @@ jobs:
         run: cargo +nightly miri test --all-features --lib --bins --no-fail-fast --workspace --exclude wincode-fuzz
 
   lint-test:
-    name: clippy & tests (stable)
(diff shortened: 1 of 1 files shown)


wincode/Cargo.toml
```diff
@@ -40,7 +41,7 @@ proptest-derive = "0.6.0"
 rand = { version = "0.8.5" }
 serde = { version = "1.0.227", features = ["derive", "rc"] }
 solana-short-vec = { version = "3.0.0" }
-criterion = { version = "0.8.1", features = ["html_reports"] }
+criterion = { version = "0.7.0", features = ["html_reports"] }
```
febo, line 44: This was needed since the latest version uses `1.86`.

cpubot (approved):

PR #124, opened 2026-01-31
Include SchemaRead/SchemaWrite bounds when deriving with generics

cpubot:
## Problem

Currently, users have to manually define bounds on struct / enum definitions with generic types when deriving `SchemaWrite` or `SchemaRead`.  E.g.,

```rs
struct Gen<T: SchemaWrite<Src = T> + for <'de> SchemaRead<'de, Dst = T>> {
    inner: T,
}
```
This is actually even worse now that `SchemaWrite` and `SchemaRead` depend on a generic `Config`. This forces the type to actually include a phantom config type.

This is a bad user experience, and also unnecessarily restricts the type such that any instance of the type must implement `SchemaRead` and `SchemaWrite`, when really that bound is only necessary when serializing / deserializing. 

## Solution

This updates `wincode-derive` to automatically generate the necessary `where` clauses on `SchemaRead` and `SchemaWrite` implementations, such that:
```rs
struct Gen<T> {
    inner: T,
}
```

Compiles and works as expected.

### DIFF
--- a/wincode/src/schema/mod.rs
+++ b/wincode/src/schema/mod.rs
@@ -3441,6 +3441,35 @@ mod tests {
     }
 
     #[test]
+    fn test_generic_struct_with_existing_bound() {
+        #[derive(
+            SchemaWrite,
+            SchemaRead,
+            serde::Serialize,
+            serde::Deserialize,
+            Debug,
+            PartialEq,
+            Eq,
+            proptest_derive::Arbitrary,
+        )]
+        #[wincode(internal)]
+        #[repr(transparent)]
+        struct GenT<T: Copy> {
+            inner: T,
+        }
+
+        proptest!(proptest_cfg(), |(value: GenT<u64>)| {
+            let serialized = serialize(&value).unwrap();
+            let bincode_serialized = bincode::serialize(&value).unwrap();
+            prop_assert_eq!(&serialized, &bincode_serialized);
+            let deserialized: GenT<u64> = deserialize(&serialized).unwrap();
+            let bincode_deserialized: GenT<u64> = bincode::deserialize(&bincode_serialized).unwrap();
+            prop_assert_eq!(&deserialized, &bincode_deserialized);
+            prop_assert_eq!(value, deserialized);
+        });
+    }
+
+    #[test]
     fn test_generic_enum() {
         #[derive(
             SchemaWrite,


wincode/src/schema/mod.rs
```diff
@@ -3319,4 +3319,166 @@ mod tests {
...
+            SchemaRead,
+            serde::Serialize,
+            serde::Deserialize,
+            Debug,
+            PartialEq,
+            Eq,
+            proptest_derive::Arbitrary,
+        )]
+        #[wincode(internal)]
+        struct GenT<T, U> {
```
kskalski, line 3377: we should also test case where generic param is constrained on something, say

```suggestion
        struct GenT<T: Display, U> {
```

I do wonder what wincode does for `?Sized` type params, though maybe that's not related to this PR

wincode-derive/src/schema_read.rs
```diff
@@ -590,9 +590,31 @@ fn append_config(generics: &mut Generics) {
...
+            bounded_ty: parse_quote!(#ident),
+            colon_token: parse_quote![:],
+            bounds,
+        }));
+    }
+    if predicates.is_empty() {
+        return;
+    }
+    let where_clause = generics.make_where_clause();
+    where_clause.predicates.extend(predicates);
```
kskalski, line 611: I suppose this won't work if the types are already constrained on `SchemaRead`... not sure if it would be easy to detect and skip auto-added constraints, probably fine to add it another time

cpubot: Yeah, serde has a `bound` attribute for this https://serde.rs/attr-bound.html. Perhaps something we could add in the future.

wincode/src/schema/mod.rs
cpubot: It errors in the `?Sized` case
```
error[E0277]: the size for values of type `T` cannot be known at compilation time
    --> wincode/src/schema/mod.rs:3447:13
     |
3447 |             SchemaRead,
     |             ^^^^^^^^^^ doesn't have a size known at compile-time
...
3457 |         struct GenT<T: ?Sized> {
     |                     - this type parameter needs to be `Sized`
     |
note: required because it appears within the type `schema::tests::test_generic_struct_with_existing_bound::GenT<T>`
    --> wincode/src/schema/mod.rs:3457:16
     |
3457 |         struct GenT<T: ?Sized> {
     |                ^^^^
note: required by a bound in `schema::SchemaRead::Dst`
    --> wincode/src/schema/mod.rs:150:5
     |
150  |     type Dst;
     |     ^^^^^^^^^ required by this bound in `SchemaRead::Dst`
     = note: this error originates in the derive macro `SchemaRead` (in Nightly builds, run with -Z macro-backtrace for more info)
help: consider removing the `?Sized` bound to make the type parameter `Sized`
     |
3457 -         struct GenT<T: ?Sized> {
3457 +         struct GenT<T> {
     |
help: consider relaxing the implicit `Sized` restriction
     |
150  |     type Dst: ?Sized;
     |             ++++++++
```

cpubot: Added a test that includes existing bounds on `T` 92f4f73a7f188971ee247484fa35b782bcd3a9a9

PR #140, opened 2026-02-03
Test/update fuzzer

0xalpharush:
- Test features introduced in https://github.com/anza-xyz/wincode/releases/tag/wincode%40v0.3.0
- Add build.rs to create seed corpus
- Test resource limiting behavior

### DIFF
--- a/fuzz/fuzz_targets/roundtrip.rs
+++ b/fuzz/fuzz_targets/roundtrip.rs
@@ -2,7 +2,12 @@
 
 use {
     libfuzzer_sys::fuzz_target,
-    std::net::{IpAddr, Ipv4Addr, Ipv6Addr},
+    std::{
+        collections::{BTreeMap, BTreeSet, HashMap, HashSet, LinkedList, VecDeque},
+        net::{IpAddr, Ipv4Addr, Ipv6Addr},
+        time::{Duration, SystemTime},
+    },
+    uuid::Uuid,
     wincode::config::{Configuration, DEFAULT_PREALLOCATION_SIZE_LIMIT},
     wincode_derive::{SchemaRead, SchemaWrite},
 };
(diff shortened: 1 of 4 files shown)


0xalpharush:
Will need to pull in https://github.com/anza-xyz/wincode/pull/141 and https://github.com/anza-xyz/wincode/pull/142 as adding the types that were recently implemented to the fuzzer [finds](https://github.com/0xalpharush/wincode/actions/runs/21655194981/job/62428439640) these issues.

fuzz/fuzz_targets/roundtrip.rs
```diff
@@ -133,213 +56,361 @@ impl NanAwareEq for FloatTypes {
...
-        let bincode_result: Result<$type, _> = bincode::deserialize($bytes);
-
-        match (wincode_result, bincode_result) {
-            (Ok(wincode_val), Ok(bincode_val)) => {
+    };
+    (unordered, $label:expr, $wincode_result:expr, $bincode_result:expr, $reser_w:expr) => {
+        match ($wincode_result, $bincode_result) {
+            (Ok(wv), Ok(bv)) => {
+                assert_eq!(wv, bv, "Deserialization mismatch for {}", $label);
+                let ws = ($reser_w)(&wv);
```
nbelenkov, line 101: In the ordered version, we check that the object reserializes back to the expected value, but here we check that it is the length but not the actul bytes

fuzz/fuzz_targets/roundtrip.rs
```diff
@@ -133,213 +56,361 @@ impl NanAwareEq for FloatTypes {
...
+macro_rules! test_config {
+    ($bytes:expr, $type:ty, unordered, $wc:expr, v1($opts:expr)) => {{
+        use bincode_1::Options;
+        let config = $wc;
+        let opts = $opts;
+        let wr: Result<$type, _> = wincode::config::deserialize($bytes, config);
+        if is_preallocation_error(&wr) {
+            let limited = opts.with_limit(DEFAULT_PREALLOCATION_SIZE_LIMIT as u64);
+            let br: Result<$type, _> = limited.deserialize_from($bytes);
+            assert!(br.is_err(),
```
nbelenkov, line 131: Technically, we only check that bincode rejects, not that the rejection reasons are equivalent, but I think it should be okay. So you could have a case where wincode hits the preallocation limit on an input where bincode would reject for a different reason

fuzz/fuzz_targets/roundtrip.rs
0xalpharush: We don't have `assert_eq!(ws, bs, "Re-serialization mismatch for {}", $label);` bc the order is unstable. Or do I misunderstand?

nbelenkov: yes that we can't do, I agree, but we could decode the re-serialized bytes and compare that way, so:

```
              let wv2 = ($deser_w)(&ws);                                                                                                                                                                                                           
              assert_eq!(wv, wv2, "Re-deserialization mismatch for {}", $label);                                                                                                                                                                   
```

Not sure if this is over the top though

PR #142, opened 2026-02-04
add length check for uuid

publicqi:
if length is not verified, a malformed input can lead to silent corruption here.


wincode/src/schema/external/uuid.rs
```diff
@@ -30,7 +30,10 @@ unsafe impl<'de, C: Config> SchemaRead<'de, C> for Uuid {
     fn read(reader: &mut impl Reader<'de>, dst: &mut MaybeUninit<Self::Dst>) -> ReadResult<()> {
         // serde serializes byte slices as a length-prefixed array.
         #[cfg(feature = "uuid-serde-compat")]
-        let _len = C::LengthEncoding::read(reader)?;
+        let len = C::LengthEncoding::read(reader)?;
+        if len != size_of::<Uuid>() {
+            return Err(crate::error::invalid_value("Uuid: invalid length prefix"));
+        }
```
cpubot: Should gate the check by the feature as well
```suggestion
        {
            let len = C::LengthEncoding::read(reader)?;
            if len != size_of::<Uuid>() {
                return Err(crate::error::invalid_value("Uuid: invalid length prefix"));
            }
        }
```

PR #152, opened 2026-02-05
feat!: remove &mut from parameters passing Reader and Writer impls

kskalski:
### Summary
Improve experience of using APIs receiving `wincode::io::Reader` and `wincode::io::Writer` by removing necessity to provide mutable reference to the type implementing those traits.
This way we no longer force mutable borrows for slices (which do implement `Reader` and `Writer`) or custom implementations of reader/writer before they are passed to the APIs.

### Background
Current API works on `&mut` since often we call several functions in sequence and this way the object is automatically "reborrowed" for each call. As a consequence the caller gets access to updated state of the reader/writer after each and after all calls, since ownership remains at the call site. This is for example used for reading or writing to slice, the slice is updated to new position after the operations.

The downside is that caller is unable to easily just pass the owned reader / writer (even well-sized slices) to the read/write functions. Instead a temporary value needs to be created and its `&mut` obtained.

##### Reading
Previously this is required pattern for using API taking a `&mut impl Reader`:
```rust
let bytes = vec![1, 2, 3, 4];
let mut bytes_slice = bytes.as_slice();
assert_eq!(deserialize_from::<u32>(&mut bytes_slice).unwrap(), 0x4030201);
assert!(bytes_slice.is_empty());
```

Now this is also supported (thanks to provided `Reader` blanket implementation for `&mut R where R: Reader`), but since functions take `impl Reader` it's also possible to do:
```rust
assert_eq!(deserialize_from::<u32>(bytes.as_slice()).unwrap(), 0x4030201);
assert!(!bytes.is_empty()); // slice update after read bytes is discarded
```

##### Writing
Similarly writing to a well-sized slice required mutably borrowing it when passing to API taking `&mut impl Writer`:
```rust
let mut bytes = vec![1, 2, 3, 4];
let mut bytes_slice = bytes.as_mut_slice();
serialize_into::<u32>(&mut bytes_slice, &0x5060708).unwrap();
assert!(bytes_slice.is_empty());
assert_eq!(bytes, [8, 7, 6, 5]);
```

Now this is also supported:
```rust
serialize_into::<u32>(bytes.as_mut_slice(), &0x6070809).unwrap();
assert_eq!(bytes, [9, 8, 7, 6]);
```

#### Breaking
`SchemaRead` and `ScheamWrite` traits are updated to take `impl Reader` / `impl Writer` instead of `&mut` to them.

One consequence is that implementations that previously performed several calls of `read` / `write` could just pass the parameter holding `reader` / `writer` to each of them, since that parameter was already known to be `&mut`. Now only passing owned object is possible, to call several functions expecting `Reader` / `Writer`, the `&mut` of the parameter needs to be used:
```rust
fn read(mut reader: impl Reader) -> ReadResult<()> {
   let x = u32::read(&mut reader)?;
   let y = u64::read(&mut reader)?;
   Ok(())
}
```

### DIFF
--- a/wincode/src/io/mod.rs
+++ b/wincode/src/io/mod.rs
@@ -225,46 +225,57 @@ impl<'a, R: Reader<'a>> Reader<'a> for &mut R {
     where
         Self: 'b;
 
+    #[inline(always)]
     fn fill_buf(&mut self, n_bytes: usize) -> ReadResult<&[u8]> {
         (*self).fill_buf(n_bytes)
     }
 
+    #[inline(always)]
     fn fill_exact(&mut self, n_bytes: usize) -> ReadResult<&[u8]> {
         (*self).fill_exact(n_bytes)
     }
 
+    #[inline(always)]
     fn fill_array<const N: usize>(&mut self) -> ReadResult<&[u8; N]> {
         (*self).fill_array()
     }
 
+    #[inline(always)]
     fn borrow_exact(&mut self, len: usize) -> ReadResult<&'a [u8]> {
         (*self).borrow_exact(len)
     }
 
+    #[inline(always)]
     fn borrow_exact_mut(&mut self, len: usize) -> ReadResult<&'a mut [u8]> {
         (*self).borrow_exact_mut(len)
     }
 
+    #[inline(always)]
     unsafe fn consume_unchecked(&mut self, amt: usize) {
         (*self).consume_unchecked(amt)
     }
 
+    #[inline(always)]
     fn consume(&mut self, amt: usize) -> ReadResult<()> {
         (*self).consume(amt)
     }
 
+    #[inline(always)]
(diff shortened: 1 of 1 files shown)


wincode/src/io/mod.rs
```diff
@@ -219,6 +219,69 @@ pub trait Reader<'a> {
     }
 }
 
+impl<'a, R: Reader<'a>> Reader<'a> for &mut R {
+    type Trusted<'b>
+        = R::Trusted<'b>
+    where
+        Self: 'b,
+        R: 'b;
```
cpubot: I believe we can drop the `R:'b`

```suggestion
        Self: 'b;
```

kskalski: hm, I think it didn't compile, let me check again

kskalski: ok, it does :) a0de7d3

kskalski:
I'm rebasing as there are some new impls on master

PR #156, opened 2026-02-05
fix: rand version bump `random` export

cpubot:
dependabot bumped the `rand` crate, but didn't update a doctest using it. Added additional CI step to ensure doctests are exercised -- apparently `--all-targets` doesn't run doc tests.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,6 +41,8 @@ jobs:
         run: cargo clippy --all-targets --features alloc -- -D warnings
       - name: tests
         run: cargo test --all-features --all-targets --no-fail-fast --workspace
+      - name: test doc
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -41,6 +41,8 @@ jobs:
         run: cargo clippy --all-targets --features alloc -- -D warnings
       - name: tests
         run: cargo test --all-features --all-targets --no-fail-fast --workspace
+      - name: test doc
+        run: cargo test --all-features --doc --no-fail-fast --workspace
```
steviez, line 45: Confirmed that we can't combine these:
```
cargo test --all-features --all-targets --no-fail-fast --workspace --doc
error: Can't mix --doc with other target selecting options
```
And these run sequentially in same env so not we're recompiling from start to run the extra step

PR #150, opened 2026-02-05
derive: `SchemaRead` ref constraints

cpubot:
Prior to the 0.3.0 release, `ZeroCopy` was implemented unconditionally for integer types on little endian targets. This meant that taking references to integer types or zero-copy structs containing integer types in a `SchemaRead` derived struct/enum trivially type-checked.

E.g.,
```rs
#[derive(SchemaWrite, SchemaRead)]
#[wincode(internal)]
enum ZeroRef<'a> {
    Unit,
    Ref(&'a StructZeroCopy),
}
```
Where `StructZeroCopy` contains integer types, (among other byte array types).

Since 0.3.0, `ZeroCopy` depends on the `Config` used for deserialization. And, more specifically, `SchemaRead` is only implemented on references `&T` where `T: ZeroCopy<Config>`. 
https://github.com/anza-xyz/wincode/blob/cc7074c2990615c335fe6580cfb7a36a50ad7a2f/wincode/src/schema/impls.rs#L1455-L1457

As such, the above example does not trivially type-check unless additional constraints are added to the `SchemaRead` implementation. In particular, the following constraint is needed:
```rs
where &'de StructZeroCopy: SchemaRead<'de, Config>
```

More generally, types containing references deriving `SchemaRead` must ensure that those references implement `SchemaRead<'de, Config>`. It would also be sufficient to constrain `T: ZeroCopy` (as it implies the former), but this is more complicated from a code generation perspective, specifically for types with generic lifetimes.

This PR implements a straight-forward solution. It iterates over fields whose types contain any lifetimes, and adds additional constraints to the where-clause that require those types to implement `SchemaRead<'de, Config>`.

Closes #145

PR #157, opened 2026-02-05
Bump wincode-derive to 0.4.0

cpubot:
This version includes the following bug fixes and features for `wincode-derive`:

- https://github.com/anza-xyz/wincode/pull/73
- https://github.com/anza-xyz/wincode/pull/114
- https://github.com/anza-xyz/wincode/pull/124
- https://github.com/anza-xyz/wincode/pull/146
- https://github.com/anza-xyz/wincode/pull/150
- https://github.com/anza-xyz/wincode/pull/152

[Full commit diff](https://github.com/anza-xyz/wincode/compare/wincode@v0.3.1...wincode-derive-0.3.2)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -827,7 +827,7 @@ dependencies = [
 
 [[package]]
 name = "wincode-derive"
-version = "0.3.1"
+version = "0.4.0"
 dependencies = [
  "darling",
  "proc-macro2",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,6 +23,3 @@ rust-version = "1.84.1"
 [workspace.dependencies]
 quote = "1.0.44"
 proc-macro2 = "1.0.106"
-
-[patch.crates-io]
-wincode-derive = { path = "wincode-derive" }
--- a/wincode-derive/Cargo.toml
+++ b/wincode-derive/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "wincode-derive"
-version = "0.3.1"
+version = "0.4.0"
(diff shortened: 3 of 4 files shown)


steviez:
Is #124 a breaking change or will pre-existing bounds continue to compile ? Thinking about semver and wanting to make sure this should be `0.3.2` and not `0.4.0`

steviez (commented):
Refreshed myself [here](https://semver.org/), a few notes:
- Since we are in `0.x.y`, [no guarantee of stability for minor](https://semver.org/#spec-item-4)
- If there is an API change, I believe the proper thing to do here is to bump the minor. It doesn't definitively state though, j[ust this tip here](https://semver.org/#how-should-i-deal-with-revisions-in-the-0yz-initial-development-phase) when dealing with `0.x.y` development

Looking at [who is currently using wincode](https://crates.io/crates/wincode/reverse_dependencies), it is:
- Anza owned repo's
- `gptman` (https://github.com/rust-disk-partition-management/gptman/pull/144)
- `coreutils` (https://github.com/uutils/coreutils/pull/10134)

> That being said, if we wanted to be very strict about this, a 0.4.0 release is justifiable.

Given that this crate already has some adoption outside of Anza and may see more, I think we should do "the right thing" and jump to `0.4.0`. We got a lot of flack (rightfully so) for our abuse of semver in the SDK, so we should probably continue to be on our best behavior moving forward 😅

steviez (approved):
Can you please update the PR title as well to reflect the update ? Otherwise, LGTM

kskalski (approved):

Issue #89, opened 2026-01-20
Unsoundness via `Pod<T>` type


tobias-nd:
Instantiating `Pod<T>` with an type that is not zero copy can result in undefined behaviour by safe code. 

While it is documented that `Pod<T>` should only be used with zero copy types, this is actually not enforced anywhere. Therefore safe code can cause undefined behaviour by deserializing non zero copy types through `Pod<T>`.

This example creates a char which is not an Unicode scalar value which will cause a segmentation fault in the rust standard library when trying to print it.

```rust
let c = Pod::<char>::deserialize(&[0xff; 4]).unwrap();
println!("{:?}", c);
```

cpubot:
Do you have a proposed solution? The derive macro could conceivably detect `Pod` usage in the `with` attribute and require something like `unsafe(with = "Pod<_>")`. But for direct usage like in the example, it's not clear what can be done other than providing clear documentation on `Pod`.

tobias-nd:
I don't have a good solution.

Your proposal with the derive macro sounds like a good first step.

> But for direct usage like in the example, it's not clear what can be done other than providing clear documentation on `Pod`.

If there's no way to provide a safe abstraction, one possible solution can be to remove the `Pod` type altogether.

One could give the user the option to create their own `Pod<T>` types, like with a macro
```rust
wincode::assert_pod! {
    unsafe struct PodAddress(Address);
}
```
which would create a type `PodAddress` that can replace `Pod<Address>`. Since the user is creating the type, they can be forced to write `unsafe`.

Alternatively, one could create a `Pod<T>` type that doesn't deserialize to `T`, but to `Pod<T>` and only provide unsafe getters/setters.

cpubot:
More context: `Pod` is sort of a shim for solana-sdk types that don't yet have wincode support. My feeling generally is that `Pod` likely should be removed / deprecated given that there are safe workarounds, and that unsoundness that could be pushed into user code rather than explicitly supported by the library. 

Given the compile-time metadata that was added in `0.2`, `Pod` is unnecessary for optimization that was necessary in `0.1`. As such, the only purpose it currently serves is as a shim for mapped memcpy/zerocopy types that don't have explicit wincode support. Given this, perhaps we gate the `Pod` struct with a feature, `unsafe_pod` (or something similar), so that the type is only present when the user explicitly opts into the feature?

PR #165, opened 2026-02-09
Deprecate the `Pod<_>` type

tobias-nd:
Create a replacement `pod_wrapper!` which can wrap types to allow for memcpy de-/serialization.

Once the deprecated `Pod<_>` type is removed, #89 will be fixed.

### DIFF
--- a/wincode/src/schema/containers.rs
+++ b/wincode/src/schema/containers.rs
@@ -1,6 +1,6 @@
 //! This module provides specialized implementations of standard library collection types that
 //! provide control over the length encoding (see [`SeqLen`](crate::len::SeqLen)), as well
-//! as special case opt-in raw-copy overrides (see [`Pod`]).
+//! as special case opt-in raw-copy overrides (see [`pod_wrapper!`]).
 //!
 //! # Examples
 //! Raw byte vec with solana short vec length encoding:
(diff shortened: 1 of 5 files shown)


tobias-nd:
~Note that this is a breaking change!~

tobias-nd:
> The types in the linked code are defined in `solana-sdk`, so the macro as defined in this crate will not actually compile in that context.

As I understand, the macro `pod_wrapper!` can be used in the same context as the old `Pod<_>` type. You just create the wrapper type locally.

I converted the code you quoted to `pod_wrapper!`: https://github.com/tobias-nd/agave/commit/d8d823323e84eab9fe012a4827a9397815d25479. (The code also used wincode 0.2 before, so I didn't convert the entire `impl<C: Config> SchemaRead<C> for VersionedMsg` to 0.4). It was only adding
```rs
wincode::pod_wrapper! {
    unsafe struct PodAddress(Address);
    unsafe struct PodHash(Hash);
    unsafe struct PodSignature(Signature);
}
```
and replacing `Pod<Address>` with `PodAddress`, etc. `cargo check` in the entry crate works.

> As [suggested in #89](https://github.com/anza-xyz/wincode/issues/89#issuecomment-3786110128), one potential remediation would be to gate the `Pod` type by something like an `unsafe_pod` feature.

Gating the `Pod` type with something like an `unsafe_pod` feature will make the entire crate graph inherit the `unsafe_pod` feature, which means you'll be able to see the `Pod` type even if you haven't opted into yourself, only one of your dependencies did.

> Or perhaps we can have some notion of an `unsafe_with` attribute where `Pod` is only accepted if that attribute is used (i.e., using pod in `with` triggers a compiler error).

This would still allow you to call the `<Pod<!> as SchemaRead<C>>::get(reader)` directly, unfortunately.

tobias-nd:
> To give users an opportunity to upgrade without breaking, let's deprecate the `Pod` struct and point them to the `pod_wrapper` macro.

Implemented. The `Pod<_>` type is now deprecated, and could be removed in 0.5.

wincode/src/schema/containers.rs
```diff
@@ -195,15 +200,76 @@ pub struct Arc<T: ?Sized, Len>(PhantomData<T>, PhantomData<Len>);
...
+pub use pod_wrapper;
+
+/// Indicates that the type is represented by raw bytes and does not have any invalid bit patterns.
+///
+/// Prefer [`pod_wrapper!`] instead.
+#[deprecated(
+    since = "0.4.6",
+    note = "This unsound type has been replaced by the `pod_wrapper!` macro."
+)]
 pub struct Pod<T: Copy + 'static>(PhantomData<T>);
```
LucasSte, line 262: Deprecations are supposed to be minor releases, not patches. The deprecation warning should have come with `0.5.0`, not `0.4.6`.

According to [semver](https://semver.org/#how-should-i-handle-deprecating-functionality):

> When you deprecate part of your public API, you should do two things: (1) update your documentation to let users know about the change, (2) issue a new minor release with the deprecation in place.

This is causing warnings in Agave, as it is pulling version 0.4.6, since 0.4.5 is specified in its toml file.

cpubot: > Deprecations are supposed to be minor releases, not patches. The deprecation warning should have come with `0.5.0`, not `0.4.6`.

By Rust convention patch releases are minor versions in `0.x` crates.

> This is causing warnings in Agave, as it is pulling version 0.4.6, since 0.4.5 is specified in its toml file.

No worries. I can fix these.

PR #167, opened 2026-02-09
Expose struct_extensions as a proc-macro

cpubot:
Currently, `struct_extensions` is only available on types that use the `SchemaRead` derive macro. `struct_extensions` can be useful on types where the type itself does not use the `SchemaRead` derive macro (i.e., it uses a custom `SchemaRead` implementation).

Consider the following example in `solana-sdk`: https://github.com/anza-xyz/solana-sdk/blob/80b834a441a508c3d26f0b768a2787d135261b7b/transaction/src/versioned/mod.rs#L309-L319.

This code is not drop-safe. If the second `read` call to `VersionedMessage` fails, the previously deserialized `Vec<Signature>` will not be dropped.

This is a perfect case for `struct_extensions`, but the type uses a custom `SchemaRead` implementation, so it cannot be used.

This PR refactors the existing `struct_extensions` helper to its own module, and additionally exposes it as a proc-macro attribute, so it can be used without the derive macros.

### DIFF
--- a/wincode-derive/src/lib.rs
+++ b/wincode-derive/src/lib.rs
@@ -12,6 +12,7 @@ mod assert_zero_copy;
 mod common;
 mod schema_read;
 mod schema_write;
+mod struct_extensions;
 
 /// Implement `SchemaWrite` for a struct or enum.
 #[proc_macro_derive(SchemaWrite, attributes(wincode))]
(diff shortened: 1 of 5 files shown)


wincode-derive/src/lib.rs
```diff
@@ -32,3 +33,50 @@ pub fn derive_schema_read(input: TokenStream) -> TokenStream {
...
+/// struct Message {
+///     body: Body,
+/// }
+///
+/// unsafe impl<'de, C: Config> SchemaRead<'de, C> for Message {
+///     type Dst = Message;
+///
+///     fn read(mut reader: impl Reader<'de>, dst: &mut MaybeUninit<Self::Dst>) -> ReadResult<()> {
+///         let msg = MessageUninitBuilder::<C>::from_maybe_uninit_mut(&mut dst);
+///         // ...
```
kskalski, line 69: This example could be expanded a bit to showcase the utility of builder a bit more (previously it was a private impl documentation, now it is a pub exposed macro, so it would be good to have good description when it's useful or point to similar doc for `struct_extensions` attribute).

PR #169, opened 2026-02-10
feat: Expose UninitBuilder as derive macro; deprecate struct_extensions

cpubot:
See https://github.com/anza-xyz/wincode/pull/167 for context on previous implementation.

Currently, `struct_extensions` is only available on types that use the `SchemaRead` derive macro. `struct_extensions` can be useful on types where the type itself does not use the `SchemaRead` derive macro (i.e., it uses a custom `SchemaRead` implementation).

Consider the following example in `solana-sdk`: https://github.com/anza-xyz/solana-sdk/blob/80b834a441a508c3d26f0b768a2787d135261b7b/transaction/src/versioned/mod.rs#L309-L319.

This code is not drop-safe. If the second `read` call to `VersionedMessage` fails, the previously deserialized `Vec<Signature>` will not be dropped.

This is a perfect case for `struct_extensions`, but the type uses a custom `SchemaRead` implementation, so it cannot be used.

This PR refactors the existing `struct_extensions` helper to its own module, and additionally exposes it as its own derive macro `UninitBuilder`, so it can be used without deriving `SchemaRead`.

## Update

Previous impl exposed `struct_extensions` as a proc macro, which has the downside that we can't hook into other derive attributes like `#[wincode(with)]`. This led me to implement https://github.com/anza-xyz/wincode/pull/168, and got me thinking that perhaps we should actually just expose this functionality as a derive macro. In particular, the case described by #168 is solved by being able to hook into `#[wincode(with)]`. This avoids the potential footgun where `SchemaWrite` is defined using attributes like `#[wincode(with)]` and later in the custom `SchemaRead` impl, one forgets to call the corresponding `read_<field>_with`. In other words, this gives us symmetry between builder methods and any attributions made on the struct definition, which seems like a better overall solution to these problems.

With this change, I feel like we should probably deprecate the `struct_extensions` attribute, as it can be confusing to have two ways to do the same thing. Additionally, it may be worth coming up with a better name for struct extensions in context of a derive macro. `struct_extensions` feels reasonable as a feature of `SchemaRead`, but feels a bit opaque as a standalone feature. Perhaps we update to something like `#[derive(UninitBuilder)]` and deprecate the `struct_extensions` attribute.

### DIFF
--- a/wincode/src/schema/mod.rs
+++ b/wincode/src/schema/mod.rs
@@ -1239,6 +1239,33 @@ mod tests {
     }
 
     #[test]
+    #[allow(deprecated)]
+    fn test_struct_extensions_sanity() {
+        #[derive(SchemaWrite, SchemaRead, Debug, PartialEq, Eq, proptest_derive::Arbitrary)]
+        #[wincode(internal, struct_extensions)]
+        struct Test {
+            a: Vec<u8>,
+            b: [u8; 32],
+            c: u64,
+        }
+
+        proptest!(proptest_cfg(), |(test: Test)| {
+            let serialized = serialize(&test).unwrap();
+            let mut uninit = MaybeUninit::<Test>::uninit();
+            let mut reader = serialized.as_slice();
+            let mut builder = TestUninitBuilder::<DefaultConfig>::from_maybe_uninit_mut(&mut uninit);
+            builder
+                .read_a(&mut reader)?
+                .read_b(&mut reader)?
+                .write_c(test.c);
+            prop_assert!(builder.is_init());
+            builder.finish();
+            let init = unsafe { uninit.assume_init() };
+            prop_assert_eq!(test, init);
+        });
+    }
+
+    #[test]
     fn test_uninit_builder_with_container() {
         #[derive(UninitBuilder, Debug, PartialEq, Eq, proptest_derive::Arbitrary)]
         #[wincode(internal)]


wincode-derive/src/lib.rs
```diff
@@ -32,3 +33,50 @@ pub fn derive_schema_read(input: TokenStream) -> TokenStream {
...
+/// struct Message {
+///     body: Body,
+/// }
+///
+/// unsafe impl<'de, C: Config> SchemaRead<'de, C> for Message {
+///     type Dst = Message;
+///
+///     fn read(mut reader: impl Reader<'de>, dst: &mut MaybeUninit<Self::Dst>) -> ReadResult<()> {
+///         let msg = MessageUninitBuilder::<C>::from_maybe_uninit_mut(&mut dst);
+///         // ...
```
kskalski: same comment as in other PR - we should move the expanded documentation, justification and examples for using the builder here

wincode-derive/src/lib.rs
```diff
@@ -32,3 +33,50 @@ pub fn derive_schema_read(input: TokenStream) -> TokenStream {
...
+///     }
+/// }
+/// ```
+///
+/// We cannot do this for enums, given the lack of facilities for placement initialization.
+#[proc_macro_derive(StructExtensions, attributes(wincode))]
```
kskalski: I'm definitely in favor of `UninitBuilder`

wincode-derive/src/lib.rs
cpubot: I'm hesitant to put a complicated example here in the derive macro crate as it cannot be type-checked or actually executed as a test. I'll put a simple case here for ease of maintainability and update lib.rs in the main crate to use the derive macro.

wincode-derive/src/lib.rs
```diff
@@ -32,3 +33,41 @@ pub fn derive_schema_read(input: TokenStream) -> TokenStream {
...
 }
+
+/// Include placement initialization helpers for structs.
+///
+/// This adds some convenience methods to structs that can avoid a lot of boilerplate when
+/// implementing custom `SchemaRead` implementations. In particular, provide methods that
+/// deal with projecting subfields of structs into `MaybeUninit`s. Without this,
+/// users have to write a litany of `&mut *(&raw mut (*dst_ptr).field).cast()` to
+/// access MaybeUninit struct fields.
+///
```
kskalski, line 46: I thought the main motivation for builder was to make partial initialization safe during errors / drops, I'm surprised it's not mentioned here

wincode/src/schema/mod.rs
```diff
@@ -1237,9 +1239,9 @@ mod tests {
     }
 
     #[test]
-    fn test_struct_extensions_with_container() {
```
kskalski, line 1240: let's keep one test with allowed deprecation for struct_extensions, to at least make sure it compiles

wincode-derive/src/lib.rs
cpubot: Updated

wincode/src/schema/mod.rs
cpubot: Done 7504c4d24490c595b37c9070b33a8e4b76a12200

wincode/src/schema/mod.rs
```diff
@@ -1238,6 +1238,32 @@ mod tests {
         });
     }
 
+    #[test]
```
kskalski, line 1241: should add allow deprecated here I guess

cpubot: done

kskalski (dismissed):

PR #177, opened 2026-02-11
Add `by_ref` function to `Reader` and `Writer` traits

### LINKED ISSUE #170: Recursive type overflow
Hello, consider this type:
```
pub enum Value {
    Usize(usize)
    List(Vec<Value>),
}
```

In older versions this was working with serialize/deserialize , now it gives

```
error[E0275]: overflow evaluating the requirement `TrustedSliceWriter<'_>: Writer`
  |
  = help: consider increasing the recursion limit by adding a `#![recursion_limit = "256"]` attribute to your crate
  = note: required for `&mut TrustedSliceWriter<'_>` to implement `Writer`
  = note: 128 redundant requirements hidden
  = note: required for `&mut &mut &mut &mut &mut &mut &mut &mut &mut &mut &mut ...` to implement `Writer`
```

Might it be expected behavior?

cpubot:
Fixes #170

## Problem

Given the following recursive type:
```rs
pub enum Value {
    Usize(usize)
    List(Vec<Value>),
}
```
Serializing or deserialization yields the following error:
```
error[E0275]: overflow evaluating the requirement `TrustedSliceWriter<'_>: Writer`
  |
  = help: consider increasing the recursion limit by adding a `#![recursion_limit = "256"]` attribute to your crate
  = note: required for `&mut TrustedSliceWriter<'_>` to implement `Writer`
  = note: 128 redundant requirements hidden
  = note: required for `&mut &mut &mut &mut &mut &mut &mut &mut &mut &mut &mut ...` to implement `Writer`
```

The issue is that the signatures of `SchemaRead::read` and `SchemaWrite::write` rely on `impl Reader` / `impl Writer`, where that instance may already be of `&mut impl Reader` or `&mut impl Writer`, given the blanket implementation for mutable references to `Reader` and `Writer`.

https://github.com/anza-xyz/wincode/blob/0086e1c3b623cba13af7457756478fc187c0cca8/wincode/src/io/mod.rs#L222
https://github.com/anza-xyz/wincode/blob/0086e1c3b623cba13af7457756478fc187c0cca8/wincode/src/io/mod.rs#L417

On a recursive type, where implementations may pass `&mut reader` or `&mut writer` to nested member types to avoid moving the parameter, an unbounded mutable borrow chain is created, resulting in the above error.

## Solution

This PR introduces a `by_ref` method to `Reader` and `Writer`, which can be used in place of creating mutable borrows on `impl Reader` or `impl Writer`.

> Similar to [`std::io::Read::by_ref`](https://doc.rust-lang.org/std/io/trait.Read.html#method.by_ref)

Importantly, in the `by_ref` implementation for the blanket `&mut Reader` and `&mut Writer` implementations, we stop any recursion from occuring:
https://github.com/anza-xyz/wincode/blob/ddee7720eb981ad784b7c55994ff33d98ac5eb4f/wincode/src/io/mod.rs#L260-L263

By using `by_ref` at all callsites rather than `&mut reader` or `&mut writer`, the issue is resolved.

### Potential pitfalls

We don't have any way to enforce that users use `by_ref` instead of `&mut`. One possible way of enforcing this would be to remove blanket `&mut` implementations and introduce wrapper types `ReaderRef` and `WriterRef`, as discussed [here](https://github.com/anza-xyz/wincode/issues/170#issuecomment-3887431830). I am hesitant to remove those blanket implementations, as it is likely to break downstream code relying on it. I think this proposal is a reasonable middle ground -- all wincode supported implementations and derive macros use `by_ref`, which should cover the vast majority of use-cases. These functions provide good documentation to push users towards using `by_ref`. We also provide a default implementation for `by_ref` in the `Reader` and `Writer` traits, which should work in virtually all cases, so it is unlikely to break people downstream.

### DIFF
--- a/wincode/codegen/tuple.rs
+++ b/wincode/codegen/tuple.rs
@@ -35,7 +35,7 @@ pub fn generate(arity: usize, mut out: impl Write) -> Result<()> {
         let write_impl = params
             .iter()
             .zip(&idxs)
-            .map(|(ident, i)| quote!( <#ident as crate::SchemaWrite<Cfg>>::write(&mut writer, &value.#i)?; ))
+            .map(|(ident, i)| quote!( <#ident as crate::SchemaWrite<Cfg>>::write(writer.by_ref(), &value.#i)?; ))
             .collect::<Vec<_>>();
 
         let read_impl = params
@@ -50,7 +50,7 @@ pub fn generate(arity: usize, mut out: impl Write) -> Result<()> {
                 };
                 quote! {
                     <#ident as crate::SchemaRead<'de, Cfg>>::read(
-                        &mut reader,
+                        reader.by_ref(),
                         unsafe { &mut *(&raw mut (*dst_ptr).#index).cast() }
                     )?;
                     #init_count
@@ -158,6 +158,7 @@ pub fn generate(arity: usize, mut out: impl Write) -> Result<()> {
                     dst: &mut core::mem::MaybeUninit<Self::Dst>
                 ) -> crate::ReadResult<()>
                 {
+                    use crate::io::Reader;
                     let dst_ptr = dst.as_mut_ptr();
                     struct DropGuard<#(#params),*> {
                         init_count: u8,
--- a/wincode-derive/src/lib.rs
+++ b/wincode-derive/src/lib.rs
@@ -60,7 +60,7 @@ pub fn derive_schema_read(input: TokenStream) -> TokenStream {
 ///         // Deserializes a `Vec<u8>` into the `payload` slot of `Message` and marks the field
 ///         // as initialized. If the subsequent `read_bytes` call fails, the `payload` field will
 ///         // be dropped.
-///         msg_builder.read_payload(&mut reader)?;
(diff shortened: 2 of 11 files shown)


wincode/codegen/tuple.rs
```diff
@@ -158,6 +158,7 @@ pub fn generate(arity: usize, mut out: impl Write) -> Result<()> {
                     dst: &mut core::mem::MaybeUninit<Self::Dst>
                 ) -> crate::ReadResult<()>
                 {
+                    use crate::io::Reader;
```
kskalski, line 161: is this import necessary now?

kskalski:
Cool, yes, this looks like a proper solution and it might also be better for ensuring intermediate borrows do not have perf impact (as so far we realied on inlines to work on all `&mut R` calls).

Just noticed one weird import in the code.

wincode/codegen/tuple.rs
cpubot: It is necessary because the `read` impl now uses `Reader::by_ref`, which needs the trait in scope

PR #182, opened 2026-02-12
Switch to Rust 2024 edition

kskalski:


febo:
Same comment here as #181, bumping the requirement will prevent programs of using the latest wincode version.

febo:
Same [comment](https://github.com/anza-xyz/wincode/pull/181#issuecomment-4016715341) applies here, should be fine to switch to 2024 edition.

PR #186, opened 2026-02-12
chore: pin nightly version to avoid flaky build errors

cpubot:
It seems the latest version of nightly broke miri. Pinning our nightly version for the miri suite to a known working version to avoid inconsistent builds.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,15 +13,15 @@ jobs:
       - uses: actions/checkout@v6
       - uses: dtolnay/rust-toolchain@master
         with:
-          toolchain: nightly-2026-02-01
-          components: rustfmt, miri
+          toolchain: nightly-2024-11-23
+          components: rustfmt, miri, rust-src
       - uses: Swatinem/rust-cache@v2
         with:
           cache-on-failure: true
       - name: rustfmt (check)
-        run: cargo +nightly-2026-02-01 fmt --all -- --check
+        run: cargo +nightly-2024-11-23 fmt --all -- --check
       - name: miri (check)
-        run: cargo +nightly-2026-02-01 miri test --all-features --lib --bins --no-fail-fast --workspace --exclude wincode-fuzz
+        run: cargo +nightly-2024-11-23 miri test --all-features --lib --bins --no-fail-fast --workspace --exclude wincode-fuzz
 
   lint-test:
     name: clippy & tests (MSRV)


levsha (dismissed):

.github/workflows/ci.yml
```diff
@@ -13,15 +13,15 @@ jobs:
       - uses: actions/checkout@v6
       - uses: dtolnay/rust-toolchain@master
         with:
-          toolchain: nightly
+          toolchain: nightly-2026-02-01
```
steviez: I'm not positive that it matters, but should we choose a nightly build that aligns with MSRV ?
https://github.com/anza-xyz/wincode/blob/0daa955b825042836e33a6e552a086ee48c7a3b9/Cargo.toml#L21
https://github.com/anza-xyz/wincode/blob/0daa955b825042836e33a6e552a086ee48c7a3b9/rust-toolchain.toml#L1-L2

For Agave, we have been following the convention of using the final corresponding (1.84 in this case) nightly build

kskalski: Using latest nightly is a double-edged sword - we get earlier validation for newest toolchain, so we can fix issues as they arise, but we also get upstream breakages like recent, since it's ultimately most-unstable toolchain you can get. It's even more painful when it happens for unrelated PRs.

Thus I agree - using nightly aligned with target toolchain is better. I wonder though how should be relate workspace's toolchain to MSRV, does it make sense to use different versions for them (e.g. https://github.com/anza-xyz/wincode/pull/181)

steviez: >  so we can fix issues as they arise, but we also get upstream breakages like recent, since it's ultimately most-unstable toolchain you can get. It's even more painful when it happens for unrelated PRs.

Yeah, I think this statement is captures the key sentiment for pinning. Pinning gives us stability and lets us choose when we want to deal with breakages. On the other hand, using nightly could mean unrelated failures in critical times (ie if we ever have an urgent patch).

So, I think pinning is the correct move and the only question in my mind is whether we should pin nightly to match MSRV. I'm leaning towards aligning them

cpubot: Ok -- pinned to nightly as of the date of the 1.84.1 release

.github/workflows/ci.yml
```diff
@@ -13,15 +13,15 @@ jobs:
       - uses: actions/checkout@v6
       - uses: dtolnay/rust-toolchain@master
         with:
-          toolchain: nightly
-          components: rustfmt, miri
+          toolchain: nightly-2025-01-30
```
kskalski: Typically we use the last nightly that uses target stable release name (which is more or less the code that was forked as a stable release of that toolchain).
That would be `2024-11-22`

```
$ cargo +nightly-2024-11-22 test
info: syncing channel updates for 'nightly-2024-11-22-x86_64-unknown-linux-gnu'
info: latest update on 2024-11-22, rust version 1.84.0-nightly (b19329a37 2024-11-21)
info: downloading component 'cargo'
^C
$ cargo +nightly-2024-11-23 test
info: syncing channel updates for 'nightly-2024-11-23-x86_64-unknown-linux-gnu'
info: latest update on 2024-11-23, rust version 1.85.0-nightly (a47555110 2024-11-22)
info: downloading component 'cargo'
```

cpubot: Setting to `2024-11-23` to avoid:
```
error: rustc 1.84.0-nightly is not supported by the following packages:
  wincode@0.4.4 requires rustc 1.84.1
  wincode@0.4.4 requires rustc 1.84.1
  wincode@0.4.4 requires rustc 1.84.1
  wincode@0.4.4 requires rustc 1.84.1
  wincode-derive@0.4.2 requires rustc 1.84.1
  wincode-derive@0.4.2 requires rustc 1.84.1
```

steviez: Hmm, well that's annoying. I guess nightly advanced to a new minor before the `1.84.1` patch release came out on the stable `1.84.0` tag. It looks like we set that min version in https://github.com/anza-xyz/wincode/pull/111; it is not clear to me that `1.84.1` is required over `1.84.0` there.

I guess using one one-day-newer nightly probably isn't a huge deal; going to think on it a bit more before giving approval / curious to hear others thoughts

kskalski (approved):
I think it's fine. 
We should keep in mind why we are using nightly here at all:
* rustfmt fmt due to settings like https://github.com/rust-lang/rustfmt/issues/4991 (still not stable, but unlikely to diverge anytime, maybe it will become stable soon)
* miri test - newer versions of toolchain might generate different code and miri might hit some code paths differently, 1 day is unlikely a big deal here. On the other hand I can imagine newer toolchain might contain more precise miri test / UB detection, not sure how likely this could be, but I suppose it doesn't outweight the previous point

PR #187, opened 2026-02-13
perf: manual decode for utf-8 char deserialization using take_array

kskalski:
* Avoid calling `fill_exact` during `char` deserialization such that `Reader` implementation isn't required to always provide borrowed slice of requested (2-4 byte) size.
* Use `take_array` of varying lengths depending on first byte check
* Manually compose `code_point` from bytes for each 2-4 len (based on https://github.com/anza-xyz/wincode/pull/33)
* Add separate ReadError for invliad UTF-8 code point value (avoids necessity to recreate buf to generate the error)
* Add benchmark for char deserialization

This removes the only call to `fill_exact` in production code outside of default / forwarding implementations of `Reader`.

#### Benchmark comparison
Checks decoding 10_000 chars from a random String
* master: 72us
* pr-33: 31us
* this PR: 27us
* bincode: 118us

### DIFF
--- a/wincode/src/schema/impls.rs
+++ b/wincode/src/schema/impls.rs
@@ -349,74 +349,56 @@ unsafe impl<'de, C: ConfigCore> SchemaRead<'de, C> for char {
 
     #[inline]
     fn read(mut reader: impl Reader<'de>, dst: &mut MaybeUninit<Self::Dst>) -> ReadResult<()> {
-        let b0 = *reader.peek()?;
-
-        let len = match b0 {
-            0x00..=0x7F => 1,
-            0xC2..=0xDF => 2,
-            0xE0..=0xEF => 3,
-            0xF0..=0xF4 => 4,
-            _ => return Err(invalid_char_lead(b0)),
-        };
-
-        if len == 1 {
-            unsafe { reader.consume_unchecked(1) };
-            dst.write(b0 as char);
-            return Ok(());
-        }
-
-        let buf = reader.fill_exact(len)?;
+        use crate::error::ReadError;
 
         // We re-validate with from_utf8 only on error path to get proper Utf8Error.
-        #[inline]
         #[cold]
-        fn utf8_error(buf: &[u8]) -> crate::error::ReadError {
+        fn utf8_error(buf: &[u8]) -> ReadError {
             invalid_utf8_encoding(core::str::from_utf8(buf).unwrap_err())
         }
-
-        // Manual UTF-8 decoder for 2x speedup by avoiding intermediate str allocation
-        let code_point = match len {
-            2 => {
-                let b1 = buf[1];
+        let b0 = *reader.peek()?;
+        let code_point = match b0 {
+            0x00..=0x7F => {
+                unsafe { reader.consume_unchecked(1) };
+                dst.write(b0 as char);
+                return Ok(());
+            }
+            0xC2..=0xDF => {
+                let [b0, b1] = reader.take_array()?;
                 // Validate continuation byte (must be 10xxxxxx)
                 if (b1 & 0xC0) != 0x80 {
-                    return Err(utf8_error(buf));
+                    return Err(utf8_error(&[b0, b1]));
                 }
                 ((b0 & 0x1F) as u32) << 6 | ((b1 & 0x3F) as u32)
             }
-            3 => {
-                let b1 = buf[1];
-                let b2 = buf[2];
+            0xE0..=0xEF => {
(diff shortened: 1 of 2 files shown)


wincode/src/schema/impls.rs
```diff
@@ -356,27 +356,29 @@ unsafe impl<'de, C: ConfigCore> SchemaRead<'de, C> for char {
...
+                unsafe { reader.consume_unchecked(1) };
+                dst.write(b0 as char);
+                return Ok(());
+            }
             0xC2..=0xDF => 2,
             0xE0..=0xEF => 3,
             0xF0..=0xF4 => 4,
             _ => return Err(invalid_char_lead(b0)),
         };
+        let mut buf = [0u8; 4];
```
kskalski: I think we could use `[MaybeUninit<u8>; 4]` and `assume_init_ref` with rust >=1.93 (https://doc.rust-lang.org/stable/core/primitive.slice.html#method.assume_init_ref), but I doubt for 4-byte array it would make any difference

kskalski:
The practical problem is that even when operating on buffers, satisfying `fill_exact` and `fill_array` may be difficult and force the implementation to do `memmove` or worse do allocations. This is the case for `agave` reader that operates on chunks of larger buffer reading them in parallel - generally when you are reading a series of values you get to the end of the current chunk, say you are at position `pos: chunk.len() - 3` and when you are asked for contiguous slice of memory of size 4, you can't just point to `chunk[pos..pos + 4]`, since it goes out of the current chunk. Even if you have the rest of the data already in other chunk, you need to do extra defragmentation:
* move `chunk[pos..]` to `chunk[0..chunk.len() - pos]`
* move `chunk.len() - pos` data from next chunk into current one
* the management of positions / `consume` after such op become more complicated (data is now in different part of buffer than usually, etc.)

I would argue that `copy_to_slice` is an "always cheaper" version of that, especially when user is going to copy out of returned slice anyway. Instead, they should copy from the position where the data is already in to destination memory. `copy_to_slice` is efficient for buffered reader, since it can easily avoid the unnecessary or short IOs, the crucial part is that it can populate target slice in 2 steps: first from the existing available data, then maybe after doing IO or switching to next chunk, the rest (possibly repeating the process several times).

It's fine to have `fill_buf(n)` API when expected return slice can be `<n`. However I think we should avoid APIs (or at least uses of it) where expectation is `>=n`.

Arguably the de-fragmentation cost mentioned above is not that bad if the expectation for returned slice can be bounded to `n <= predictable_max`. So fortunately getting rid of `fill_exact` is not a blocker. In case of this PR I think populating small piece of uninitialized stack is better than forcing reader to give access to contiguous memory (even in your reference impl we can avoid unnecessary memmove this way). I will do some comparisons of the assembly involved in real case, so we also know perf implications for current out of mem readers.

wincode/src/schema/impls.rs
cpubot: It's still not totally clear to me why we want to eliminate usage of `fill_buf` / `fill_exact`, but if we do go this route, perhaps
```rs
let mut buf = MaybeUninit::<[u8; 4]>::uninit();
// Safety: len is at most 4, so slice is up to buf.len() and casting [u8] to uninit is safe
let uninit_slice = unsafe {
    core::slice::from_raw_parts_mut(buf.as_mut_ptr().cast::<MaybeUninit<u8>>(), len)
};
reader.copy_into_slice(uninit_slice)?;

let buf = unsafe { core::slice::from_raw_parts(buf.as_ptr().cast::<u8>(), len) };
```

kskalski: I will post a bit more context on that in a different place - in short the issue is that chunked buffered reader can't guarantee >1 byte sized contiguous slices are always available.

For this PR I was also thinking of building on top of https://github.com/anza-xyz/wincode/pull/33 and use `take_array` in each branch separately, but I can also compare with the version above.

cpubot: Sounds good. Would definitely like to hear more about the background context on this

kskalski: See https://github.com/anza-xyz/wincode/pull/53#issuecomment-3918126497

cpubot (approved):
Nice!

wincode/src/schema/impls.rs
kskalski: Ok, I think we got what I wanted in a win-win fashion - current state of the PR gets rid of `fill_exact` and achieves better performance than master and even slightly better than patch proposed in https://github.com/anza-xyz/wincode/pull/33 (see benchamark numbers from PR description)

PR #188, opened 2026-02-13
feat: add take_array replacing all usages of fill_array

kskalski:
All uses of `Reader::fill_array` boil down to
```rust
let bytes: [u8; N] = *reader.fill_array();
unsafe { reader.consume_unchecked(N) };
```
which suggests this API isn't in its proper shape.
Additionally it forces implementation to always have `N` consecutive memory bytes possible to borrow, which makes it difficult to implement buffered reader.

`take_array`, which reads directly into stack created array and returns it, is a direct replacement for all uses of `fill_array` in `wincode` repository. Right now it's just a wrapper over `fill_array` and `consume_unchecked`.

Changes:
* add `take_array`
* replace `fill_array + consume_unchecked` with `take_array`

### DIFF
--- a/wincode/src/io/mod.rs
+++ b/wincode/src/io/mod.rs
@@ -77,6 +77,16 @@ pub trait Reader<'a> {
         Ok(unsafe { &*src.as_ptr().cast::<[u8; N]>() })
     }
 
+    /// Return exactly `N` bytes as `[u8; N]` and advance by `N`.
+    ///
+    /// Errors if fewer than `N` bytes are available.
+    #[inline]
+    fn take_array<const N: usize>(&mut self) -> ReadResult<[u8; N]> {
+        let arr = *self.fill_array::<N>()?;
+        unsafe { self.consume_unchecked(N) }
+        Ok(arr)
+    }
+
     /// Zero-copy: return a borrowed slice of exactly `len` bytes and advance by `len`.
     ///
     /// The returned slice is tied to `'a`. Prefer [`Reader::fill_exact`] unless you truly need zero-copy.
@@ -275,6 +285,11 @@ impl<'a, R: Reader<'a> + ?Sized> Reader<'a> for &mut R {
     #[inline(always)]
     fn fill_array<const N: usize>(&mut self) -> ReadResult<&[u8; N]> {
         (*self).fill_array()
+    }
+
+    #[inline(always)]
+    fn take_array<const N: usize>(&mut self) -> ReadResult<[u8; N]> {
+        (*self).take_array()
     }
 
     #[inline(always)]
--- a/wincode/src/len.rs
+++ b/wincode/src/len.rs
@@ -258,13 +258,11 @@ macro_rules! impl_fix_int {
             #[inline(always)]
             #[allow(irrefutable_let_patterns)]
             fn read<'de>(mut reader: impl Reader<'de>) -> ReadResult<usize> {
-                let bytes = reader.fill_array::<{ size_of::<$type>() }>()?;
+                let bytes = reader.take_array::<{ size_of::<$type>() }>()?;
                 let len = match C::ByteOrder::ENDIAN {
-                    Endian::Big => <$type>::from_be_bytes(*bytes),
-                    Endian::Little => <$type>::from_le_bytes(*bytes),
+                    Endian::Big => <$type>::from_be_bytes(bytes),
+                    Endian::Little => <$type>::from_le_bytes(bytes),
                 };
-                // SAFETY: `fill_array` ensures we read exactly `size_of::<$type>()` bytes.
-                unsafe { reader.consume_unchecked(size_of::<$type>()) };
                 let Ok(len) = usize::try_from(len) else {
                     return Err(pointer_sized_decode_error());
                 };
--- a/wincode/src/schema/external/uuid.rs
+++ b/wincode/src/schema/external/uuid.rs
@@ -36,9 +36,7 @@ unsafe impl<'de, C: Config> SchemaRead<'de, C> for Uuid {
                 return Err(crate::error::invalid_value("Uuid: invalid length prefix"));
             }
         }
-        let bytes = *reader.fill_array::<{ size_of::<Uuid>() }>()?;
-        // SAFETY: `fill_array` guarantees we get exactly `size_of::<Uuid>()` bytes.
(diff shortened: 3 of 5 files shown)


wincode/src/io/mod.rs
```diff
@@ -70,11 +70,30 @@ pub trait Reader<'a> {
     /// Return exactly `N` bytes as `&[u8; N]` without advancing.
     ///
     /// Errors if fewer than `N` bytes are available.
-    fn fill_array<const N: usize>(&mut self) -> ReadResult<&[u8; N]> {
-        let src = self.fill_exact(N)?;
+    #[deprecated = "use `read_to_array`, `copy_to_array` or `fill_buf` instead"]
+    fn fill_array<const N: usize>(&mut self) -> ReadResult<[u8; N]> {
```
cpubot: All `fill_*` methods return slices, so I don't think we should change that convention. Additionally, this is a breaking change, so if we were to deprecate, we should add deprecation notice before making any signature or functionality changes.

Also, I don't see any harm in keeping this as is. I wouldn't deprecate.

wincode/src/io/mod.rs
```diff
@@ -70,11 +70,30 @@ pub trait Reader<'a> {
...
         // SAFETY:
         // - `fill_exact` ensures we read N bytes.
-        Ok(unsafe { &*src.as_ptr().cast::<[u8; N]>() })
+        Ok(unsafe { arr.assume_init() })
+    }
+
+    /// Return exactly `N` bytes as `[u8; N]` and advance by `N`.
+    ///
+    /// Errors if fewer than `N` bytes are available.
+    fn read_to_array<const N: usize>(&mut self) -> ReadResult<[u8; N]> {
```
cpubot: As far as naming, we don't have a precedent within reader methods for `read_*`. But, we do have a convention with `SchemaRead`, where `read` takes a `dst: &mut MaybeUninit<Self::Dst>`.

Perhaps we call this `get_array`, which resembles `SchemaRead::get`

wincode/src/io/mod.rs
```diff
@@ -70,11 +70,30 @@ pub trait Reader<'a> {
...
+        self.copy_into_slice(slice)?;
         // SAFETY:
         // - `fill_exact` ensures we read N bytes.
-        Ok(unsafe { &*src.as_ptr().cast::<[u8; N]>() })
+        Ok(unsafe { arr.assume_init() })
+    }
+
+    /// Return exactly `N` bytes as `[u8; N]` and advance by `N`.
+    ///
+    /// Errors if fewer than `N` bytes are available.
```
cpubot, line 82: Should definitely be inlined

wincode/src/io/mod.rs
```diff
@@ -70,11 +70,30 @@ pub trait Reader<'a> {
...
+    fn read_to_array<const N: usize>(&mut self) -> ReadResult<[u8; N]> {
+        let mut arr: MaybeUninit<[u8; N]> = MaybeUninit::uninit();
+        // SAFETY: `[MaybeUninit<u8>]` has same layout as `[u8]`
+        self.copy_into_array(&mut arr)?;
+        // SAFETY:
+        // - `fill_exact` ensures we read N bytes.
+        Ok(unsafe { arr.assume_init() })
```
cpubot: Could just be:
```rs
let ar = *self.fill_array::<N>()?;
unsafe { self.consume_unchecked(N) }
Ok(ar)
```

I think we should be hesitant to add additional `copy_nonoverlapping` invocations, as it's less optimizable than a `ptr::read` or a dereference, especially for small values

wincode/src/io/mod.rs
```diff
@@ -203,14 +222,10 @@ pub trait Reader<'a> {
...
-        // SAFETY:
-        // - `fill_array` must do the appropriate bounds checking.
-        unsafe {
-            ptr::copy_nonoverlapping(src, dst.as_mut_ptr(), 1);
-            self.consume_unchecked(N);
-        }
-        Ok(())
+        let ptr = dst.as_mut_ptr() as *mut MaybeUninit<u8>;
+        // SAFETY: ptr points to N bytes memory from dst
+        let slice = unsafe { std::slice::from_raw_parts_mut(ptr, N) };
```
cpubot: I don't think this should necessarily forward to `copy_into_slice`. `copy_into_slice` takes a runtime-lengthed slice, so we're essentially erasing the compile time length information and bolting on a runtime length. Compiler is probably able to figure this out given the inlining, but I don't see any reason to change unless we have good reason to.

wincode/src/io/mod.rs
kskalski: Sorry, I meant to keep the signature and body unchanged. Reverted. Also removed deprecation, I still think we should do it, but clearly it is a wider discussion and should be done separately (maybe also deprecating `borrow_exact`).

wincode/src/io/mod.rs
kskalski: done

wincode/src/io/mod.rs
kskalski: I don't like `get_array`, since it implies no state change (or at least no position change) in the reader, while we do actually consume the bytes.

Renamed to `take_array`.

As for the other names - I suppose reader's `copy_into` should be then renamed to `read_` to match signature / usage in `SchemaRead` (though personally I would do the opposite and rename `SchemaRead`'s `read` to `copy_into`), but this is clearly also a point for a separate discussion.

wincode/src/io/mod.rs
kskalski: ok, changed as suggested for now, since `fill_array` isn't deprecated

wincode/src/io/mod.rs
kskalski: Reverted.

Actually I think the default `io::Reader` implementations here are not conservative enough given `fill_buf`'s documentation
```
    /// This is _not_ required to return exactly `n_bytes`, it is required to return _up to_ `n_bytes`.
    /// Use [`Reader::fill_exact`] if you need exactly `n_bytes`.
```

in particular `fill_exact` does assume it will just get exactly `n_bytes` out of `fill_buf` and thus all the impls that rely on `fill_exact`.

Seems like most of those should be moved into slice-specific impl that can assume full contiguous memory for the whole read range, while the default here should only rely on `copy_into_slice` instead of calling `fill_exact`.

PR #193, opened 2026-02-17
test: add external types from solana-sdk/agave

0xalpharush:
Open to suggestions on where these should live/how best to take a dep.

### DIFF
--- a/fuzz/Cargo.lock
+++ b/fuzz/Cargo.lock
@@ -3,10 +3,542 @@
 version = 4
 
 [[package]]
+name = "adler2"
+version = "2.0.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


nbelenkov:
I think it makes more sense to pull published crates and pin to a version vs pulling from repo directly

0xalpharush:
I used a commit because some of the changes were very recent, and I figured it made sense to fuzz them prior to their release (and we can easily bump to pull in new impl's/ changes). 

I can remove any unpublished types and change the dep to crates.io

nbelenkov:
ahhh that makes sense, sounds good then

0xalpharush:
Updated to use latest crates.io release

cpubot (approved):

PR #195, opened 2026-02-17
chore: add name to config and remove cron

0xalpharush:
Pulled out from https://github.com/anza-xyz/wincode/pull/193

Addresses breaking change on our end that is causing failures https://github.com/anza-xyz/wincode/actions/runs/22120331764/job/63938583510


cpubot (approved):

nbelenkov:
lgtm as well

PR #196, opened 2026-02-18
feat(io): add reader implementation for BufRead

kskalski:
Introduces `BufReadAdapter<R>` in `wincode::io::std_read` module, a low-overhead wrapper over any `std::io::BufRead` source (e.g. `BufReader<R>`) implementing `wincode::io::Reader`.

Mix of consuming and scoped borrowing operations supported by checking consume state on each operation and using one of:
* Fast path — when the reader’s internal buffer already holds enough bytes, `peek_array` and `take_scoped` return a subslice directly with no allocation, `copy_into_slice` uses `read` directly into the `dst` buffer
* Aux-buffer fallback — when a contiguous slice larger than one currently available from internal reader's `fill_buf` is requested, bytes are assembled into an `aux_buf: Vec<u8>`. Subsequent operations can re-serve or extend from this buffer, but switch back to using internal reader once the bytes in `aux_buf` are consumed.

### DIFF
--- a/wincode/src/io/mod.rs
+++ b/wincode/src/io/mod.rs
@@ -563,6 +563,8 @@ impl<W: Writer + ?Sized> Writer for &mut W {
 mod cursor;
 pub mod slice;
 #[cfg(feature = "std")]
+pub mod std_read;
+#[cfg(feature = "std")]
 pub mod std_write;
 #[cfg(feature = "alloc")]
 mod vec;
--- a/wincode/src/io/std_read.rs
+++ b/wincode/src/io/std_read.rs
@@ -1,0 +1,428 @@
+use {
+    crate::io::{ReadResult, Reader, read_size_limit, slice::SliceUnchecked},
+    std::{
+        io::{BufRead, BufReader, ErrorKind, Read},
+        mem::{MaybeUninit, transmute},
+        num::NonZeroUsize,
+    },
+};
+
+/// [`Reader`] implementation over any [`BufRead`] source.
+///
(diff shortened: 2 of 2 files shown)


kskalski:
I certainly agree that removing far-look-ahead functions is desirable and this PR I think demonstrated the reasons well. :)

I will take a look at the benches too - thanks.

Some quick thoughts:
- `peek_byte` seems as useful API to me and is trivial to implement without cost for `BufRead` (it isn't for `Read` though, so maybe a good question is if it's going to be common to use `Read` without buffering - e.g. de-compressors sometimes do implement `Read`, but not `BufRead`)
- we don't really need `consume` even with `peek_byte`, `take_byte` seems fine to consume peeked byte
- a big complexity addition and need for extra branching on each call in this PR is for handling `as_trusted_for`, since it needs to delay `inner.consume` - partially it is a consequence of wrapping `BufRead` as black-box - when owning and managing the buffer manually, it would be possible to bump the internal cursor and still return the slice reader for bytes preceding the position marker. Having a more efficient `as_trusted_for` would be a big win - we use it often for large and small ranges
- if branching inherently added for `as_trusted_for` due to contract of `BufRead` is the reason for big perf hit, then I think it's going to be better to expose our own buffered reader instead of wrapping `BufRead` - need to take a closer look at benchmarks
- `peek_array` and `take_scoped` are certainly APIs I would cut - the former isn't used at all and for the latter we know many ways to get rid of

kskalski:
Wrapping `BufRead` constrains a lot how the buffer can be checked for optimized trusted reader or other functions, so I think current `std_read` is good enough and I will look into possible optimizations in the specific buffered reader implementation instead.

PR #197, opened 2026-02-18
test: add cursor test case deserializing nested zero-copy payload

kskalski:
Add test that checks that `as_trusted_reader` readers not supporting zero-copy can still properly deserialize zero-copy inner payload (that schemas do not call `borrow_exact` on any step of the deserializaton chain).

I noticed that `SchemaRead::read` implementation may call `borrow_exact` without compile errors if it operates on lifetimes carefully, this will result in runtime error only when passed reader doesn't support zero-copy. We should have at least some coverage to ensure our `read` impls do not call `borrow_exact` unnecessarily. The easiest way to do that is to use some reader that do not support zero-copy and read content that (might, in theory) use calls to `borrow_exact`.

### DIFF
--- a/wincode/src/io/cursor.rs
+++ b/wincode/src/io/cursor.rs
@@ -691,5 +691,28 @@ mod tests {
             // Remaining bytes are untouched
             prop_assert_eq!(&cursor.inner[bytes.len()..], &vec![1; bytes.len()]);
         }
+
+        #[cfg(feature = "derive")]
+        #[test]
+        fn cursor_read_items_with_inner_zero_copy(bytes in proptest::collection::vec(any::<u8>(), 64)) {
+            use crate::{config::DefaultConfig, SchemaRead};
+
+            // Test reader not supporting zero-copy, but used to read items that contain nested
+            // zero-copy content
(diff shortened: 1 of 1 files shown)


kskalski:
A side question: could we somehow detect if given type requires borrow-style zero-copy deserialization and forbid its use with readers that do not support zero-copy?
E.g. this test case compiles, but fails at runtime:
```rust
        #[test]
        fn cursor_read_items_with_inner_zero_copy(bytes in proptest::collection::vec(any::<u8>(), 64)) {
            use crate::{config::DefaultConfig, SchemaRead};

            // Test reader not supporting zero-copy, but used to read items that contain nested
            // zero-copy content
            #[derive(crate::SchemaRead)]
            #[wincode(internal)]
            struct NonZeroCopyWrapper<'a> {
                zero_copy_content: &'a [u8; 8],
            }

            let mut cursor = Cursor::new(&bytes);
            let mut dst = MaybeUninit::uninit();
            <[NonZeroCopyWrapper<'_>; 8] as SchemaRead<DefaultConfig>>::read(&mut cursor, &mut dst)
                .unwrap();
        }
```

PR #200, opened 2026-02-20
feat: implement schemas for bytes::Bytes

kskalski:
`bytes` crate is fairly popular, especially when working on zero-copy networking code. Its `Bytes` and `BytesMut` are clone-able wrappers over contiguous memory slices:
* `Bytes` is a read-only reference counted analog for `Box<[u8]>` with slicing support
* `BytesMut` is similar to `Vec<u8>`

It's useful to be able to deserialize them directly (it's used in agave/ledger).

This PR:
* adds optional feature `bytes`
* adds `SchemaRead` and `SchemaWrite` implementation for `Bytes` and `BytesMut`

### DIFF
--- a/wincode/src/schema/external/bytes.rs
+++ b/wincode/src/schema/external/bytes.rs
@@ -4,6 +4,7 @@ use {
         io::{Reader, Writer},
         ReadResult, SchemaRead, SchemaWrite, WriteResult,
     },
+    alloc::boxed::Box,
     bytes::{Bytes, BytesMut},
     core::mem::MaybeUninit,
 };


kskalski:
It might be useful to implement reader / writer for them too, though alternative of using `bytes.as_ref()` and/or converting from/to `Vec` seems feasible, so I leave it for future consideration.

wincode/Cargo.toml
```diff
@@ -26,8 +26,10 @@ solana-short-vec = ["dep:solana-short-vec"]
 derive = ["dep:wincode-derive"]
 uuid = ["dep:uuid"]
 uuid-serde-compat = ["uuid"]
+bytes = ["dep:bytes"]
 
 [dependencies]
+bytes = { version = "1.11.1", optional = true, default-features = false }
```
cpubot: Without `std`, it appears `bytes` still [requires at least `alloc`](https://docs.rs/bytes/1.11.1/src/bytes/lib.rs.html#74).

wincode/Cargo.toml
```diff
@@ -26,8 +26,10 @@ solana-short-vec = ["dep:solana-short-vec"]
 derive = ["dep:wincode-derive"]
 uuid = ["dep:uuid"]
 uuid-serde-compat = ["uuid"]
+bytes = ["dep:bytes"]
```
cpubot: Without `std`, `bytes` still [requires at least `alloc`](https://docs.rs/bytes/1.11.1/src/bytes/lib.rs.html#74). So we may need to also require `alloc` when `bytes` is enabled
e.g.,

```toml
bytes = ["alloc", "dep:bytes"]
```

cpubot: There's also an `extra-platforms` feature https://github.com/tokio-rs/bytes/blob/00cc5ff2bd861f7b62603a268ec477c3cead5f1e/README.md?plain=1#L30-L43

Perhaps we also add something like the following, thought probably not strictly necessary 
```toml
bytes-extra-platforms = ["alloc", "dep:bytes", "bytes/extra-platforms"]
```

wincode/src/schema/external/bytes.rs
```diff
@@ -0,0 +1,92 @@
+use {
+    crate::{
+        config::Config,
+        io::{Reader, Writer},
+        ReadResult, SchemaRead, SchemaWrite, WriteResult,
+    },
+    bytes::{Bytes, BytesMut},
+    std::mem::MaybeUninit,
```
cpubot: Can avoid `std` here

```rs
    core::mem::MaybeUninit,
```

wincode/src/schema/external/bytes.rs
```diff
@@ -0,0 +1,92 @@
...
+};
+
+unsafe impl<'de, C: Config> SchemaRead<'de, C> for Bytes {
+    type Dst = Self;
+
+    fn read(reader: impl Reader<'de>, dst: &mut MaybeUninit<Self::Dst>) -> ReadResult<()> {
+        // Box<[u8]> is closest to native representation of Bytes and avoid data copy or allocs.
+        let boxed_slice = <Box<[u8]> as SchemaRead<'de, C>>::get(reader)?;
```
cpubot, line 17: Should also use `alloc::boxed::Box` to avoid `std` requirement

kskalski: looks like `Box` is resolved based on the present features, so I think it works automatically

wincode/src/schema/external/bytes.rs
kskalski: done

wincode/Cargo.toml
kskalski: ok, added `alloc` to `bytes` and the `bytes-extra-platforms` feature, though I'm not sure if it wouldn't just work for the user using bytes as their dependency with that feature enabled...

wincode/src/schema/external/bytes.rs
cpubot: We still need to import `Box` from `alloc` otherwise compilation will fail when the `std` feature is not enabled. `Box` is only globally imported from `std`'s prelude
```
cargo test --features alloc,bytes
    Blocking waiting for file lock on build directory
   Compiling wincode v0.4.4 (/home/zbrown/code/wincode/wincode)
   Compiling thiserror v2.0.18
error[E0412]: cannot find type `Box` in this scope
  --> wincode/src/schema/external/bytes.rs:16:28
   |
16 |         let boxed_slice = <Box<[u8]> as SchemaRead<'de, C>>::get(reader)?;
   |                            ^^^ not found in this scope
   |
help: consider importing one of these structs
   |
1  + use crate::containers::Box;
   |
1  + use alloc::boxed::Box;
```

kskalski: weird, that command compiles for me just fine, but I guess it doesn't hurt to have an import, added in 95f4da7

cpubot: I think probably due to the fact that we have "std" as the default feature set. If you remove that and try compiling with the bytes feature, you should see the error.

May be worth us introducing a cargo powerset check to CI in the future to catch these as we add more features

kskalski:
Right, confirmed it failed with
```
wincode/wincode$ cargo build --lib --no-default-features --features=alloc,bytes
```
and compiles now

PR #202, opened 2026-02-22
refactor: replace matches of TypeMeta series with summing helper

kskalski:
Generated code and some implementations of Schema can be simplified by using const helper function on `TypeMeta` that joins several types. The "join" there assumes a "concatenated" serialization, which
* sums the static sizes
* sets zero copy to conjunction of `zero_copy` of static types
* turns into dynamic type if any of the specified types is dynamic

### DIFF
--- a/wincode/src/schema/mod.rs
+++ b/wincode/src/schema/mod.rs
@@ -106,9 +106,16 @@ impl TypeMeta {
         }
     }
 
-    /// Returns type overriding static type's `zero_copy` to `false` (if `keep_zero_copy == false`)
+    /// Returns this [`TypeMeta`] instance with `zero_copy` masked by `keep_zero_copy`.
     ///
-    /// Note: type is never upgraded to `zero_copy` if it wasn't such before.
+    /// For `TypeMeta::Static`, this preserves `size` and computes:
+    /// `zero_copy = zero_copy && keep_zero_copy`.
+    ///
+    /// For `TypeMeta::Dynamic`, this is a no-op.
+    ///
+    /// This method never upgrades a type to zero-copy.
+    /// - `keep_zero_copy(true)` leaves the flag unchanged.
+    /// - `keep_zero_copy(false)` clears the flag.
     pub const fn keep_zero_copy(self, keep_zero_copy: bool) -> Self {
         match self {
             Self::Static { size, zero_copy } => TypeMeta::Static {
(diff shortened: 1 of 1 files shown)


wincode/src/schema/mod.rs
```diff
@@ -105,6 +105,47 @@ impl TypeMeta {
             _ => panic!("Type is not static"),
         }
     }
+
+    /// Returns type overriding static type's `zero_copy` to `false` (if `keep_zero_copy == false`)
+    ///
+    /// Note: type is never upgraded to `zero_copy` if it wasn't such before.
+    pub const fn with_zero_copy(self, keep_zero_copy: bool) -> Self {
```
cpubot: `with_zero_copy` could be misleading from a naming perspective. The comment does a good job documenting behavior, but perhaps we could make the name more clear that it doesn't upgrade.

Perhaps something like `maintain_zero_copy`?

kskalski: I tried to preserve the typical naming for such transformations using `with_` prefix, but yeah, with the directional logic it might be misleading. Maybe just "keep_zero_copy"

wincode/src/schema/mod.rs
```diff
@@ -105,6 +105,47 @@ impl TypeMeta {
...
+        }
+    }
+
+    /// Aggregate `types` assuming they are placed in memory and serialized together
+    ///
+    /// Returns `Static` type if all types are static:
+    /// * with `size` equal to sum of individual sizes and
+    /// * `zero_copy` set to true if all types are zero-copy.
+    ///
+    /// Returns `Dynamic` type if any of the `types` is dynamic.
```
cpubot: Perhaps we could add some doctests here, e.g.,
```rs
    /// ```
    /// use wincode::TypeMeta;
    ///
    /// let types = [
    ///     TypeMeta::Static { size: 1, zero_copy: true },
    ///     TypeMeta::Static { size: 2, zero_copy: true },
    ///     TypeMeta::Dynamic,
    ///     TypeMeta::Static { size: 3, zero_copy: true },
    /// ];
    /// assert_eq!(TypeMeta::join_types(types), TypeMeta::Dynamic);
    /// ```
    ///
    /// ```
    /// use wincode::TypeMeta;
    ///
    /// let types = [
    ///     TypeMeta::Static { size: 1, zero_copy: true },
    ///     TypeMeta::Static { size: 2, zero_copy: true },
    ///     TypeMeta::Static { size: 3, zero_copy: true },
    /// ];
    /// assert_eq!(TypeMeta::join_types(types), TypeMeta::Static { size: 6, zero_copy: true });
    /// ```
    ///
    /// ```
    /// use wincode::TypeMeta;
    ///
    /// let types = [
    ///     TypeMeta::Static { size: 1, zero_copy: true },
    ///     TypeMeta::Static { size: 2, zero_copy: false },
    ///     TypeMeta::Static { size: 3, zero_copy: true },
    /// ];
    /// assert_eq!(TypeMeta::join_types(types), TypeMeta::Static { size: 6, zero_copy: false });
    /// ```
```

wincode/src/schema/mod.rs
```diff
@@ -105,6 +105,47 @@ impl TypeMeta {
...
+        match self {
+            Self::Static { size, zero_copy } => TypeMeta::Static {
+                size,
+                zero_copy: zero_copy && keep_zero_copy,
+            },
+            Self::Dynamic => Self::Dynamic,
+        }
+    }
+
+    /// Aggregate `types` assuming they are placed in memory and serialized together
```
cpubot: Convention in the codebase is to finish lines with a period.

Additionally, could clean up the wording, e.g.,

```rs
/// Combines multiple constituent [`TypeMeta`] values into one aggregate.
///
/// Intended for composite types whose constituents are serialized sequentially.
///
/// Semantics:
/// - If any input is `Dynamic`, returns `Dynamic`.
/// - Otherwise returns `Static` with:
///   - `size = sum of all constituent sizes`
///   - `zero_copy = logical AND of all constituent zero_copy flags`
///
/// Notes:
/// - This function does **not** validate layout/padding; it only combines metadata.
/// - For `N = 0`, the result is `TypeMeta::Static { size: 0, zero_copy: true }`.
/// - The caller must ensure the summed size is meaningful for the target type.
```

wincode/src/schema/mod.rs
```diff
@@ -105,6 +105,47 @@ impl TypeMeta {
...
+    ///
+    /// Returns `Static` type if all types are static:
+    /// * with `size` equal to sum of individual sizes and
```
cpubot: Convention in the codebase is to use `-` for lists

wincode/src/schema/mod.rs
```diff
@@ -105,6 +105,47 @@ impl TypeMeta {
             _ => panic!("Type is not static"),
         }
     }
+
+    /// Returns type overriding static type's `zero_copy` to `false` (if `keep_zero_copy == false`)
```
cpubot: Suggesting more descriptive semantics

```rs
/// Returns this [`TypeMeta`] instance with `zero_copy` masked by `keep_zero_copy`.
///
/// For `TypeMeta::Static`, this preserves `size` and computes:
/// `zero_copy = zero_copy && keep_zero_copy`.
///
/// For `TypeMeta::Dynamic`, this is a no-op.
///
/// This method never upgrades a type to zero-copy.
/// - `keep_zero_copy(true)` leaves the flag unchanged.
/// - `keep_zero_copy(false)` clears the flag.
```

kskalski: Make sense: https://github.com/anza-xyz/wincode/pull/202/commits/a619154fbbf402388dfff4e00ada8f872a1bc66a

wincode/src/schema/mod.rs
kskalski: ack

wincode/src/schema/mod.rs
kskalski: Sounds good: https://github.com/anza-xyz/wincode/pull/202/commits/a619154fbbf402388dfff4e00ada8f872a1bc66a

wincode/src/schema/mod.rs
kskalski: done

cpubot (approved):
Looks great

PR #205, opened 2026-02-25
example: proposal for split api

kskalski:
Simplified example on how SchemaRead and Reader could be split into borrowing and non-borrowing APIs.
Includes chunking iterators that return sub-readers of appropriate type that can implement bound checks skipping.

The idea is:
* `trait ReaderNonBorrowing` expose `read_x` (byte, array + fill_buf / copy_into style API)
* `trait ReaderBorrowing<'de>: ReaderNonBorrowing`, adds  `borrow_bytes`
* `trait SchemaDecodeBorrowing<'de>` expose `decode_borrowed(ReaderBorrowing)`
* `trait SchemaDecode` expose a different entry point - `decode(ReaderNonBorrowing)`
* deserializing `&'a T` is done through `SchemaDecodeBorrowing` by blanket impl constrained on `ZeroCopy` as in wincode, plus any special impls for `&str`, `&[T]`, etc.
* even though `SchemaDecodeBorrowing` can be trivially implemented for any `SchmeaDecode` type, we skip having blanket `impl<'a, T: SchemaDecode> SchemaDecodeBorrowing<'a> for T` or making `SchemaDecode` inherit from `SchemaDecodeBorrowing`, since that will conflict with blanket impl of `SchemaDecodeBorrowing` for `&'de T` - the impact can be minimized by having derive automatically generate concrete implementations and manual impl is very simple (such impl is necessary when non-borrowing type is used in borrowing type annotated with derive (derive cannot say which field implements which trait, so it will blindly always emit `decode_borrowed` calls) 
* we need separate derive support for `SchemaDecode` and `SchemaDecodeBorrowing` - they will generate different trait implementations, each of them will implement and emit calls of `decode` vs `decode_borrowed` respectively

This API ensures that decoding types with references can only be done with borrowing reader. It also simplifies signature and implementations necessary for non-borrowing reader.

### DIFF
--- a/wincode/examples/split_api.rs
+++ b/wincode/examples/split_api.rs
@@ -48,7 +48,7 @@ impl<'de, T: ReaderBorrowing<'de>> ReaderBorrowing<'de> for &mut T {
     }
 }
 
-pub trait SchemaDecode: for<'a> SchemaDecodeBorrowing<'a>
+pub trait SchemaDecode
 where
     Self: Sized,
 {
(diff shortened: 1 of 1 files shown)


kskalski:
Rght, I agree with the summary and the conclusion.

PR #245, opened 2026-03-18
feat(containers): add function to decode Vec with given body len

kskalski:
It's often useful to decode just vector's body with known length (which in hand-tuned serialization formats may be stored in other fields or implied from previously decoded data).

Add function `wincode::containers::Vec::<T, SeqLen>::decode_body_with_len::<ConfigCore>(reader, len)` that does:
* prealloc check based on `SeqLen<ConfigCore>`
* allocate vector with enough capacity
* copy into properly sized slice of space capacity
* set len of allocated vec

### DIFF
--- a/wincode/src/schema/containers.rs
+++ b/wincode/src/schema/containers.rs
@@ -87,6 +87,69 @@ use {
 #[cfg(feature = "alloc")]
 pub struct Vec<T, Len>(PhantomData<Len>, PhantomData<T>);
 
+#[cfg(feature = "alloc")]
+impl<'de, T, Len> Vec<T, Len> {
+    /// Decode exactly `len` items of `T` from `reader` into a freshly allocated [`Vec`](vec::Vec).
+    ///
+    /// Errors if the prealloc check for `len` fails, fewer than `len` items are available,
+    /// or any item fails to decode.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// # #[cfg(feature = "alloc")] {
+    /// # use wincode::containers;
+    /// # use wincode::config::DefaultConfig;
+    /// # use wincode::len::BincodeLen;
+    /// # type C = DefaultConfig;
+    /// let data = [1u8, 2, 3, 4];
+    /// let serialized = wincode::serialize(&data).unwrap();
+    ///
+    /// let decoded = containers::Vec::<u8, BincodeLen>::decode_body_with_len::<C>(
+    ///     &serialized[..],
+    ///     data.len(),
+    /// )
+    /// .unwrap();
+    ///
+    /// assert_eq!(decoded.as_slice(), data.as_slice());
+    /// # }
+    /// ```
+    ///
+    /// ```
+    /// # #[cfg(feature = "alloc")] {
+    /// # use wincode::containers;
+    /// # use wincode::config::DefaultConfig;
+    /// # use wincode::len::BincodeLen;
+    /// # type C = DefaultConfig;
+    /// let data = [1u8, 2, 3, 4];
+    /// let serialized = wincode::serialize(&data).unwrap();
+    ///
+    /// // Requesting more elements than were serialized returns an error.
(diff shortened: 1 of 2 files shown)


wincode/src/schema/containers.rs
```diff
@@ -87,6 +87,69 @@ use {
...
+    pub fn decode_body_with_len<C: ConfigCore>(
+        reader: impl Reader<'de>,
+        len: usize,
+    ) -> ReadResult<vec::Vec<T::Dst>>
+    where
+        Len: SeqLen<C>,
+        T: SchemaRead<'de, C>,
+    {
+        Len::prealloc_check::<T>(len)?;
+        let mut vec: vec::Vec<T::Dst> = vec::Vec::with_capacity(len);
```
kskalski, line 145: maybe a better idea would be to provide Box<[T]> decoding instead - Vec should be trivial to create from box and with `Box::<[T]>::new_uninit_slice` we get exactly the requested capacity

wincode/src/schema/containers.rs
```diff
@@ -87,6 +87,69 @@ use {
...
+    /// ```
+    pub fn decode_body_with_len<C: ConfigCore>(
+        reader: impl Reader<'de>,
+        len: usize,
+    ) -> ReadResult<vec::Vec<T::Dst>>
+    where
+        Len: SeqLen<C>,
+        T: SchemaRead<'de, C>,
+    {
+        Len::prealloc_check::<T>(len)?;
```
davidkleymann: Shouldn't this use `Len::read_prealloc_check::<T::Dst>(reader.by_ref())?` ?

kskalski: No, `read_prealloc_check` reads the length from the reader and the point of this API is to use `len` provided in fn parameter instead of reading it. We still want to validate the length against the limit defined by `Len: SeqLen<C>`, the call here does exactly (and only) that.

davidkleymann: Ah sorry, I got confused I meant `Len::prealloc_check::<T::Dst>(reader.by_ref())?`

kskalski: Oh, it should be `Len::prealloc_check::<T::Dst>(len)?`, right.

kskalski:
Calling the function is a bit verbose (https://github.com/anza-xyz/solana-sdk/pull/630), since `Vec<T, Len>` doesn't include config in its signature and thus it needs to be passed as extra generic argument on the function, even though typically `C` will be a `Config` and `Len` would be taken from `C::LengthEncoding`.

It feels natural to put such function inside `Vec`, but possibly it should be free floating or in some extra module.

kskalski:
My point of view is:
* having direct API is just simpler and easier to comprehend and use
* this function alone eliminates all uses of `take_scoped` from `solana-sdk`
* as for prealloc check - I can see arguments both ways:
  * as you pointed out earlier lack of pre-alloc check is the one against using helper functions like `decode_into_slice_t`,
  * but in fact it is easy for user to just add the check manually before decoding - they might have already did the validation of the len, maybe manually or by reading it earlier with `LengthEncoding::read_prealloc_check` - I'm leaning towards removing this part
  * without necessity to have len, maybe the API can be simplified further, requiring user to specify the len type is a pain
  * however, we could also provide a `SeqLen` type that skips prealloc check, this way the API would be flexible to support both modes, e.g. `Vec::<T, NoAllocCheckLen>::decode_body_with_len::<C>(..)`
* the `Ctx` proposal could be a generalization of this kinds of APIs, simply calling into the relevant direct function - it is a powerful mechanism, but I think it adds a burden of learning how to use it and given we just need to fix 1 single use-case, it feels like using sledgehammer to crack a nut
* the issue of where to place the new function remains - having it in `containers::Vec` is far from perfect

kskalski:
Better API for this will probably involve defining some trait, e.g. `containers::DecodeSequenceBody` that we could implement for various containers.
In a way this would get really close to the `Ctx` proposal and given that https://github.com/anza-xyz/wincode/pull/220 now uses a separate dedicated trait for decoding with context seems better to just go with that.

PR #290, opened 2026-04-06
feat(io): add advance function to Reader

kskalski:
Handling legacy binary formats occasionally require skipping over a number of bytes when reading serialized data.
Two such cases the came up recently are:
* trailing bytes encountered deserializing `solana_ledger::BitVec` (https://github.com/anza-xyz/agave/pull/11624/)
* delegations in `EpochStakes` (https://github.com/anza-xyz/agave/pull/11758/)

This PR adds `advance` function, which can be used to discard a number of input bytes without inspecting them. By default it's implemented with `take_scoped()`, which limits it to readers supporting such call - `std::io::Read` implementations, which do not support `take_scoped` have a more expensive dedicated `advance` implementation that iterate copying input buffer or using `consume` in `BufReader`.

### DIFF
--- a/wincode/src/io/std_read.rs
+++ b/wincode/src/io/std_read.rs
@@ -79,17 +79,18 @@ impl<R: Read + ?Sized> Reader<'_> for BufReader<R> {
 
     fn advance(&mut self, mut n_bytes: usize) -> ReadResult<()> {
         use io::BufRead as _;
-        loop {
-            let buffered = self.buffer().len();
-            if let Some(remaining) = n_bytes.checked_sub(buffered) {
-                self.consume(buffered);
+        while n_bytes != 0 {
+            let buf_len = loop {
+                match self.fill_buf() {
+                    Ok(buf) => break buf.len(),
+                    Err(e) if e.kind() == io::ErrorKind::Interrupted => continue,
+                    Err(e) => return Err(e.into()),
+                }
+            };
+
+            if let Some(remaining) = n_bytes.checked_sub(buf_len) {
+                self.consume(buf_len);
                 n_bytes = remaining;
-                if n_bytes == 0 {
-                    break;
-                }
-                if self.fill_buf()?.is_empty() {
-                    return Err(read_size_limit(n_bytes));
-                }
             } else {
                 self.consume(n_bytes);
                 break;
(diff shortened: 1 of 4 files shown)


wincode/src/io/std_read.rs
```diff
@@ -60,13 +60,43 @@ impl<R: Read + ?Sized> Reader<'_> for ReadAdapter<R> {
...
+        use io::BufRead as _;
+        loop {
+            let buffered = self.buffer().len();
+            if let Some(remaining) = n_bytes.checked_sub(buffered) {
+                self.consume(buffered);
+                n_bytes = remaining;
+                if n_bytes == 0 {
+                    break;
+                }
+                if self.fill_buf()?.is_empty() {
```
cpubot: Do we need to account for `ErrorKind::Interrupted`?

E.g.,
```rs
fn advance(&mut self, mut n_bytes: usize) -> ReadResult<()> {
    use io::BufRead as _;
    while n_bytes != 0 {
        let buffered = loop {
            match self.fill_buf() {
                Ok(buf) => break buf,
                Err(e) if e.kind() == io::ErrorKind::Interrupted => continue,
                Err(e) => return Err(e.into()),
            }
        };

        if buffered.is_empty() {
            return Err(read_size_limit(n_bytes));
        }

        let take = n_bytes.min(buffered.len());
        self.consume(take);
        n_bytes -= take;
    }
    Ok(())
}
```

kskalski: right, good to handle interruped similarly as standard library... I help the checked sub code so we can avoid allowing arithmetic side effects

wincode/src/io/std_read.rs
```diff
@@ -79,17 +79,18 @@ impl<R: Read + ?Sized> Reader<'_> for BufReader<R> {
...
+        while n_bytes != 0 {
+            let buf_len = loop {
+                match self.fill_buf() {
+                    Ok(buf) => break buf.len(),
+                    Err(e) if e.kind() == io::ErrorKind::Interrupted => continue,
+                    Err(e) => return Err(e.into()),
                 }
+            };
+
+            if let Some(remaining) = n_bytes.checked_sub(buf_len) {
```
0jovi0, line 92: Doesn't this need a buf_len == 0 guard? When the stream is exhausted, fill_buf() returns Ok(&[]), so buf_len == 0 → checked_sub(0) yields Some(n_bytes) unchanged → consume(0) is a no-op → while n_bytes != 0 spins forever.

kskalski: thanks, added test to catch it and fixed the code

kskalski:
In a way it's a convenience API, but having it as part of the trait allows for more direct and efficient implementation.
If we encountered >1 cases where it isn't needed, but would allow optimal way to handle them, it could be the case for other users.
If that's not convincing, maybe we need to gate by a finding more evidence this is useful.

kskalski:
Counter proposal is to handle the use-case at schema level by providing "discard" adapters https://github.com/anza-xyz/wincode/pull/359

PR #293, opened 2026-04-09
feat: make usize/isize zero-copy on 64-bit platform

kskalski:
Similarly to other primitive types, also checks for `target_endian` to only make types zero-copy of little-endian target platform.

### DIFF
--- a/wincode/src/schema/impls.rs
+++ b/wincode/src/schema/impls.rs
@@ -193,6 +193,11 @@ macro_rules! impl_pointer_width {
                 const TYPE_META: TypeMeta = if C::IntEncoding::STATIC {
                     TypeMeta::Static {
                         size: size_of::<$target>(),
+                        #[cfg(all(target_pointer_width = "64", target_endian = "big"))]
+                        zero_copy: matches!(C::ByteOrder::ENDIAN, Endian::Big),
+                        #[cfg(all(target_pointer_width = "64", target_endian = "little"))]
+                        zero_copy: matches!(C::ByteOrder::ENDIAN, Endian::Little),
+                        #[cfg(not(target_pointer_width = "64"))]
                         zero_copy: false,
                     }
                 } else {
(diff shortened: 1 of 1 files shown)


wincode/src/schema/impls.rs
```diff
@@ -193,6 +193,11 @@ macro_rules! impl_pointer_width {
                 const TYPE_META: TypeMeta = if C::IntEncoding::STATIC {
                     TypeMeta::Static {
                         size: size_of::<$target>(),
+                        #[cfg(all(target_pointer_width = "64", target_endian = "big"))]
+                        zero_copy: matches!(C::ByteOrder::ENDIAN, Endian::Big),
+                        #[cfg(all(target_pointer_width = "64", target_endian = "little"))]
+                        zero_copy: matches!(C::ByteOrder::ENDIAN, Endian::Little),
+                        #[cfg(not(target_pointer_width = "64"))]
```
cpubot: I think we could actually just do
```rs
#[cfg(target_pointer_width = "64")]
zero_copy: C::IntEncoding::ZERO_COPY,
#[cfg(not(target_pointer_width = "64"))]
zero_copy: false,
```

`C::IntEncoding::ZERO_COPY` already factors in platform endianness

wincode/src/schema/impls.rs
cpubot, line 189: We'd also want

```rs
#[cfg(target_pointer_width = "64")]
unsafe impl<C: ConfigCore> ZeroCopy<C> for $type where C::IntEncoding: ZeroCopy<C> {}
```

wincode/src/schema/impls.rs
kskalski: ah, I was using float's impl as reference, but that's the one that isn't following `IntEncoding` :)
updated

wincode/src/schema/impls.rs
kskalski: yap, added

cpubot (approved):

PR #312, opened 2026-05-13
fix(io): ensure Vec<u8> and Cursor<Vec<u8>> don't expose uninit mem

### LINKED ISSUE #306: Memory exposure in serialized output
Wincode can leak contents of uninitialized memory into the output when writing into a Vec. This can be used to bypass ASLR or leak sensitive information from memory.

## Steps to reproduce

You can reproduce the issues by adding the following tests to the codebase and running `cargo +nightly miri test --release`

wincode/src/io/cursor.rs:

```rust
#[cfg(all(test, miri, feature = "alloc"))]
mod miri_ub_tests {
    use {super::*, alloc::vec::Vec};

    #[test]
    fn cursor_vec_write_after_gap_leaves_uninitialized_bytes() {
        let mut cursor = Cursor::new(Vec::with_capacity(8));
        cursor.set_position(4);

        // Issue: the safe `Cursor<Vec<u8>>` writer grows the vector to the new
        // cursor position after writing, but it does not initialize the gap
        // between the old length and the write position. Reading the gap from
        // the returned safe `Vec<u8>` is therefore an uninitialized read in Miri.
        cursor.write(&[1]).unwrap();
        let vec = cursor.into_inner();
        let _uninitialized_gap_byte = vec[0];
    }

    #[test]
    fn cursor_mut_vec_write_after_gap_leaves_uninitialized_bytes() {
        let mut vec = Vec::with_capacity(8);
        let mut cursor = Cursor::new(&mut vec);
        cursor.set_position(4);

        // Issue: the same gap-initialization bug is reachable through the safe
        // `Cursor<&mut Vec<u8>>` writer; after the cursor is dropped, safe code
        // can read uninitialized bytes from the now-extended vector.
        cursor.write(&[1]).unwrap();
        drop(cursor);
        let _uninitialized_gap_byte = vec[0];
    }
}
```

wincode/src/io/vec.rs:

```rust
#[cfg(all(test, miri, feature = "std"))]
mod miri_ub_tests {
    use {
        super::*,
        crate::serialize_into,
        std::time::{Duration, UNIX_EPOCH},
    };

    #[test]
    fn vec_trusted_writer_error_leaves_reserved_tail_uninitialized() {
        let before_epoch = UNIX_EPOCH.checked_sub(Duration::from_secs(1)).unwrap();
        let value = (0xAAu8, before_epoch);
        let mut vec = Vec::new();

        // Issue: `Vec<u8>::as_trusted_for` sets the vector length before the
        // trusted window has been initialized. Safe serialization of a static
        // composite type can reserve such a window, write an earlier field, and
        // then return an error from a later field. The safe caller receives an
        // `Err`, but the vector length now covers uninitialized bytes.
        assert!(serialize_into(&mut vec, &value).is_err());
        let _uninitialized_reserved_byte = vec[1];
    }
}
```

cpubot:
Closes #306

### DIFF
--- a/wincode/src/io/cursor.rs
+++ b/wincode/src/io/cursor.rs
@@ -374,18 +374,18 @@ pub(super) mod vec {
         // SAFETY: by calling `as_trusted_for`, caller guarantees they
         // will fully initialize `n_bytes` of memory and will not write
         // beyond the bounds of the slice.
-        Ok(unsafe { CursorVecUnchecked::new(inner, pos) })
+        Ok(unsafe { VecPosUnchecked::new(inner, pos) })
     }
 }
 
 #[cfg(feature = "alloc")]
-struct CursorVecUnchecked<'a> {
+struct VecPosUnchecked<'a> {
     inner: &'a mut Vec<u8>,
     pos: &'a mut usize,
 }
 
 #[cfg(feature = "alloc")]
-impl<'a> CursorVecUnchecked<'a> {
+impl<'a> VecPosUnchecked<'a> {
     /// # Safety
     ///
     /// The caller must ensure that `*pos` is within `inner`'s capacity and
@@ -398,7 +398,7 @@ impl<'a> CursorVecUnchecked<'a> {
 }
 
 #[cfg(feature = "alloc")]
-impl<'a> Writer for CursorVecUnchecked<'a> {
+impl<'a> Writer for VecPosUnchecked<'a> {
     #[inline]
     fn write(&mut self, src: &[u8]) -> WriteResult<()> {
         // SAFETY:
--- a/wincode/src/io/cursor.rs
+++ b/wincode/src/io/cursor.rs
@@ -290,6 +290,35 @@ mod vec {
         Ok(())
     }
 
+    /// Zero-fill the gap between `inner.len()` and `pos`, if `pos` is past the
+    /// current initialized length.
+    ///
+    /// # Safety
+    ///
+    /// The caller must ensure that `inner.capacity() >= pos`.
+    ///
+    /// If `pos > inner.len()`, the range `inner.len()..pos` must be spare capacity
+    /// owned by `inner` and valid to write as `MaybeUninit<u8>`. This function
+    /// initializes that entire gap with zero bytes and then sets `inner.len()` to
+    /// `pos`.
+    ///
+    /// Callers should normally establish this by calling `maybe_grow(inner, pos, n)`
+    /// with any `n` such that `pos + n` is checked and fits in the vector capacity.
+    #[inline]
+    unsafe fn zero_fill_gap(inner: &mut Vec<u8>, pos: usize) {
+        if let Some(init_gap) = pos.checked_sub(inner.len()) {
+            let spare = inner.spare_capacity_mut();
+            debug_assert!(spare.len() >= init_gap);
+
+            unsafe {
+                spare
+                    .get_unchecked_mut(..init_gap)
+                    .fill(MaybeUninit::new(0));
+                inner.set_len(pos);
+            }
+        }
+    }
+
     /// Add `len` to the cursor's position and update the length of the vector if necessary.
     ///
     /// # SAFETY:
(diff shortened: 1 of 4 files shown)


wincode/src/io/std_write.rs
kskalski: we can use a new `zero_fill_gap` helper here and `maybe_grow` above

wincode/src/io/cursor.rs
```diff
@@ -290,6 +290,35 @@ mod vec {
...
+    ///
+    /// If `pos > inner.len()`, the range `inner.len()..pos` must be spare capacity
+    /// owned by `inner` and valid to write as `MaybeUninit<u8>`. This function
+    /// initializes that entire gap with zero bytes and then sets `inner.len()` to
+    /// `pos`.
+    ///
+    /// Callers should normally establish this by calling `maybe_grow(inner, pos, n)`
+    /// with any `n` such that `pos + n` is checked and fits in the vector capacity.
+    #[inline]
+    unsafe fn zero_fill_gap(inner: &mut Vec<u8>, pos: usize) {
```
kskalski, line 307: seems like zero_fill_gap is always called after maybe_grow (including the code that duplicates similar impl in cursor that I marked), we should probably just consolidate those into a single fn

wincode/src/io/cursor.rs
```diff
@@ -311,43 +340,90 @@ mod vec {
...
         }
+
+        if next_pos > cur_len {
+            // SAFETY: any gap before the trusted window was initialized before
+            // constructing this writer, and this call just initialized the bytes
+            // from the previous cursor position through `next_pos`.
+            unsafe { self.inner.set_len(next_pos) }
+        }
+
+        *self.pos = next_pos;
```
kskalski: does calling `vec::add_len(self.inner, self.pos, src.len())` here instead (probably marking it inline too) degrade perf?

wincode/src/io/cursor.rs
```diff
@@ -311,43 +340,90 @@ mod vec {
...
         // SAFETY: by calling `as_trusted_for`, caller guarantees they
         // will fully initialize `n_bytes` of memory and will not write
         // beyond the bounds of the slice.
-        Ok(unsafe { SliceMutUnchecked::new(buf) })
+        Ok(unsafe { CursorVecUnchecked::new(inner, pos) })
     }
+}
+
+#[cfg(feature = "alloc")]
+struct CursorVecUnchecked<'a> {
```
kskalski: the struct doesn't hold a `Cursor` type, let's rename it to something like VecPosUnchecked, which will also disambiguate some name duplication with std_write CursorVecUnchecked (which does hold an actual `Cursor` type)

wincode/src/io/cursor.rs
kskalski, line 260: This sub-mod doesn't reference `Cursor` type from this module (or std::io) at all and is not specific to it. Probably for a separate PR, but seems like it would better fit in vec.rs

wincode/src/io/cursor.rs
cpubot: That should work without issue ddc7f4f07db7f38814888aee1a9b79bb5cf2af47

wincode/src/io/cursor.rs
cpubot: Done https://github.com/anza-xyz/wincode/pull/312/changes/c60773cc259351e08c48ecca3dbff0fe9d1117b4

wincode/src/io/std_write.rs
cpubot: Done https://github.com/anza-xyz/wincode/pull/312/changes/c60773cc259351e08c48ecca3dbff0fe9d1117b4

wincode/src/io/cursor.rs
cpubot: Done https://github.com/anza-xyz/wincode/pull/312/commits/142cc7bb06d604dd6ff03944dab4c677ed045ca5

kskalski (approved):

PR #318, opened 2026-05-17
Fix derive name collisions

chen-robert:
## Summary
- fully qualify paths emitted by derive macros instead of importing them into generated scopes
- avoid generated helper names that collide with user-defined types

## Testing
- cargo test --workspace

### DIFF
--- a/wincode-derive/src/assert_zero_copy.rs
+++ b/wincode-derive/src/assert_zero_copy.rs
@@ -52,28 +52,27 @@ pub(crate) fn assert_zero_copy(args: &SchemaArgs, repr: &StructRepr) -> Result<T
 
     Ok(quote! {
         const _: () = {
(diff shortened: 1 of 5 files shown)


kskalski:
Could you provide motivation for avoiding imports in the generated code blocks? I suppose it collides with user-defined imports, though does it fail or cause wrong types to be used or just emits some warnings?

chen-robert:
@kskalski Yes, this breaks a common case for Anchor - user account structs called `Config`. This fails to compile because `Config` is shadowed by the wincode `Config` struct in the generated code.

```
#[account]
struct Config {
  pub value: u64,
}
```

becomes

```
#[derive(SchemaRead)]
pub struct Config {
  pub value: u64,
}
```

kskalski (approved):

cpubot (approved):

PR #320, opened 2026-05-19
feat: impl SchemaWrite/SchemaRead for Cell/RefCell

### LINKED ISSUE #314: Missing "rc" feature
I use bincode V1 in my project I tried to migrate to this project and there was a little gotcha that kept me from doing that and It's the "rc" feature that serde support which helps with structs like

```rust
#[derive(Serialize, Deserialize, Default, Clone)]
pub struct Rw<T> {
    pub value: Rc<RefCell<T>>,
}
```

It kinda ignores all the Rc<Refcell<Type>> and only serialize it as it's equivalent to 

```rust
#[derive(Serialize, Deserialize, Default, Clone)]
pub struct Rw<T> {
    pub value: T,
}
```

Can we have similar behavior here too?

cpubot:
Closes #314

### DIFF
--- a/wincode/src/schema/impls.rs
+++ b/wincode/src/schema/impls.rs
@@ -2340,7 +2340,7 @@ where
 {
     type Dst = Cell<T::Dst>;
 
-    const TYPE_META: TypeMeta = T::TYPE_META.keep_zero_copy(false);
+    const TYPE_META: TypeMeta = T::TYPE_META;
 
     #[inline]
     fn read(reader: impl Reader<'de>, dst: &mut MaybeUninit<Self::Dst>) -> ReadResult<()> {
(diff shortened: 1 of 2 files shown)


wincode/src/schema/impls.rs
```diff
@@ -2312,3 +2313,78 @@ where
...
+    }
+}
+
+unsafe impl<'de, C: ConfigCore, T> SchemaRead<'de, C> for Cell<T>
+where
+    T: SchemaRead<'de, C>,
+{
+    type Dst = Cell<T::Dst>;
+
+    const TYPE_META: TypeMeta = T::TYPE_META.keep_zero_copy(false);
```
kskalski: Docs say `Cell<T>` has the same memory layout as `T` (https://doc.rust-lang.org/stable/core/cell/struct.Cell.html#memory-layout), I wonder if it would be fine to use that property, at least for deserialization.

cpubot: Yeah, good call. Updated.

kskalski (approved):

PR #325, opened 2026-06-09
feat(derive): support wincode(with) on single field newtype variant

kskalski:
Allows the `with` attribute to be placed directly on a single-field enum variant, as shorthand for annotating the inner field:

```rust
enum Message {
    #[wincode(with = "containers::Vec<u32, UseIntLen<u16>>")]
    Ids(Vec<u32>),
    Empty,
}
```

This is exactly equivalent to `Ids(#[wincode(with = "...")] Vec<u32>)` — it produces identical bytes and applies the same encoding override. Previously the attribute was only accepted on fields; on a variant it errored with `Unknown field: with`.

#### Difference from serde

serde also accepts `#[serde(with = ...)]` on variants, but it means something different and is more permissive:

- **serde** treats it as a *whole-variant override*: a pair of functions operating over *all* the variant's fields collectively (any arity — unit, tuple, struct). It exists precisely to customize multi-field variants.
- **wincode** `with` is a *type* override for a single value, not a function. So on a variant it can only be sugar for the single inner field, and multi-field variants are rejected.

The error message and module docs call this distinction out explicitly so users coming from serde aren't surprised by the single-field restriction.

### DIFF
--- a/wincode-derive/src/common.rs
+++ b/wincode-derive/src/common.rs
@@ -328,6 +328,10 @@ pub(crate) struct Variant {
     pub(crate) fields: Fields<Field>,
     #[darling(default)]
     pub(crate) tag: Option<Expr>,
+
+    /// Variant-level `SchemaRead` and `SchemaWrite` override for single-field (newtype) variants.
+    #[darling(default)]
+    pub(crate) with: Option<Type>,
 }
 
 impl Variant {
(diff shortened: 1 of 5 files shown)


kskalski:
Hm, if the inner type is the only member, this syntax feels natural and obvious to me - it's placed before a name that is mapped 1:1 to a given T. 

The current syntax spooked both me and the reviewer here (https://github.com/anza-xyz/solana-sdk/pull/748#discussion_r3372486352), so it's not an isolated impression.

On the other hand if newtype struct places `wincode(with=)` next to the inner type, then it's more consistent. I would vote for allowing both (this PR), but let's see if this comes up again in the future.

PR #326, opened 2026-06-15
ci: update to fuzz-upload-action v2

yevshev:
Update to https://github.com/asymmetric-research/fuzz-upload-action/releases/tag/v2.0.0 which uses API keys instead of username/password and no longer requires `FUZZ_API_ORIGIN` (we're deprecating username/password auth and simplifying the config)

> [!NOTE]
> This assumes that an org-level API key has been generated in the AR dashboard and that the `FUZZ_API_KEY` gh actions secret is set.

### DIFF
--- a/.github/workflows/fuzz.yml
+++ b/.github/workflows/fuzz.yml
@@ -4,8 +4,8 @@ on:
   push:
     branches: [master]
 
-permissions:                                                                                                                                                                                                                          
-    contents: read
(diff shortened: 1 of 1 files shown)


nbelenkov (approved):
AR are changing the flow atm, we just need to set the API key before merging

cpubot (approved):

PR #327, opened 2026-06-15
feat: add support for bumpalo crate

cpubot:
This adds support for the [`bumpalo`](https://crates.io/crates/bumpalo) crate.

In particular, the implementation leverages wincode's context system (`SchemaReadContext`) to facilitate deserializing `bumpalo`'s collection/boxed types with the given `bumpalo::Bump` allocator. This will make it straightforward to re-use allocations for deserialization work that sits in hot paths.

To make this fully robust, I am going to add context support to the derive macro so that it is trivial to use these types (and other arbitrary context types) in user-defined structs, but I have left that work for a follow up PR.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -67,9 +67,9 @@ checksum = "843867be96c8daad0d758b57df9392b6d8d271134fce549de6ce169ff98a92af"
 
 [[package]]
 name = "bumpalo"
-version = "3.20.2"
+version = "3.20.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5d20789868f4b01b2f2caec9f5c4e0213b41e3e5702a50157d699ae31ced2fcb"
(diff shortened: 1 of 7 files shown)


kskalski (dismissed):
Looks nice. 

I suppose this should be more-or-less transparent to functionality of zero-copy references, right?
 
If we deserialize something as `bumpalo::Vec<'b, &'a Struct>`, we just need a constraint that `'a : 'b`, i.e. references do not outlive neither the allocator, nor the input slice.
If so, we could add such to tests.

A more obvious case (i.e. simply guaranteed by Rust) is deserializing zero-copy refs borrowing the `bumpalo::Vec<'b, u8>`, which should also not outlive the `'b`, but we could also have a test that nothing gets in a way.

kskalski:
That's great, I think it's worth adding such cases to tests.

kskalski (approved):

PR #329, opened 2026-06-17
fuzz: cover all wincode-decoded types in solana_types harness

R9295:
This also bumps rust-toolchain.toml to the same version as agave due to the use of ``std::iter::chain``

### DIFF
--- a/fuzz/Cargo.lock
+++ b/fuzz/Cargo.lock
@@ -3,17 +3,86 @@
 version = 4
 
 [[package]]
+name = "Inflector"
+version = "0.11.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "fe438c63458706e03479442743baae6c88256498e6431708f6dfc520a26515d3"
+dependencies = [
+ "lazy_static",
+ "regex",
+]
+
+[[package]]
 name = "adler2"
 version = "2.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
-name = "agave-feature-set"
-version = "4.1.0-alpha.0"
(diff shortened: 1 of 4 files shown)


nbelenkov (approved):
LGTM, solana-gossip does pull in every dependency known to man, but I don't think we can sadly do anything about that

kskalski:
So far we were preferring to keep the rust-toolchain close to the minimum version, so we can see the real performance on such target. Minimum version is dependent on solana-sdk platform support.

As for the fuzzing, seems like this test is doing subset of what we added in agave with frozen_abi::StableAbi roundtrip tests, so maybe we don't need it here in wincode repo?

R9295:
@kskalski since these are deployed on fuzzcorp (continuous fuzzing), in my opinion, it would be ideal to centralize it in this repository.

Perhaps, what we can do in the CI is to patch the toolchain before deployment. WDYT?

nbelenkov:
I would be keen to have all the round-trip fuzzing for types in this repo, as I think it makes the most sense. Re toolchain I don't have strong preference, whatever works best

kskalski:
I think this could be done as a stop-gap, but essentially I think serialization library should not pull in its user's implementations / types as its dependencies. 

If we want to fuzz-test types defined in agave or solana-sdk when using wincode, we should basically add a testing step out in those repositories or in a separate repository that would track master for all the tested dependencies.

What fits into this repository is possibly a representative duplicate of those types so we can detect regressions as part of wincode CI, though that will create maintenance burden / dilemma whether we should update those types, so above solution would be preferable.

As a side note, there is some work going on to integrate fuzz testing into `solana-frozen-abi`, a crate to test serialization sanity that is already used across the ecosystem. I do wonder if we could converge on running the tests defined here in `fuzz/` through that crate and put that into appropriate repo. I suppose it's a bit independent question - the crucial step would be to migrate the fuzzing CI workflow to where it seems to belong.

R9295:
@nbelenkov  / @kskalski That's fair. I'll redirect the patches to agave & solana-sdk.

PR #341, opened 2026-06-18
Only add trait bounds for type params that are actually used

IVSOP:
Previous behaviour:
```rs
pub trait SomeTrait {
    type SomeAssociatedType;
}

pub struct SomeStruct<T: SomeTrait> {
    pub data: T::SomeAssociatedType
}
```

Despite `SomeStruct` never actually storing a `T`, but only a `T::SomeAssociatedType`, trait bounds were placed on `T` itself. With these changes, trait bounds are only placed in types that are actually used, such as `T::SomeAssociatedType`. If a `T` were stored, bounds would also be placed for T itself.

The original behaviour is very unergonomic for users working with generics. Also, serde has a behaviour similar to the one that was just implemented.

All existing tests are passing, and added new tests specific to this functionality.

As a side note, serde does things a bit differently. `SomeStruct` would compile even if `T::SomeAssociatedType` did not implement serde's traits, but would then error whenever you tried to actually use serde methods on it. However, like in this new approach, it would only require the traits to be placed on types that are actually used in the struct.

### DIFF
--- a/wincode-derive/src/common.rs
+++ b/wincode-derive/src/common.rs
@@ -7,7 +7,7 @@ use {
     quote::{ToTokens as _, quote},
     std::{
         borrow::Cow,
-        collections::VecDeque,
+        collections::{HashSet, VecDeque},
         fmt::{self, Display},
     },
     syn::{
@@ -846,7 +846,7 @@ impl Visit<'_> for HasLifetime {
 }
 
 struct HasTypeParam<'a> {
-    type_params: &'a std::collections::HashSet<&'a Ident>,
+    type_params: &'a HashSet<&'a Ident>,
     found: bool,
 }
 
@@ -864,21 +864,71 @@ pub(crate) struct GenericField {
     pub(crate) ty: Type,
 }
 
+/// Find fields that mention one of the generic type parameters on the derived type.
+///
+/// The returned `GenericField` keeps two types:
+/// - `target`: the type whose `SchemaRead`/`SchemaWrite` impl will be called
+/// - `ty`: the actual type stored in the field
+///
+/// This is specifically based on the _fields_, not just the generic parameter list.
+/// For example, if a type parameter is only used to name an associated type, we return the
+/// associated type that is actually stored:
+///
+/// ```ignore
+/// trait HasValue {
+///     type Value;
+/// }
+///
+/// struct Wrapper<T: HasValue> {
+///     value: T::Value,
+/// }
+///
+/// // yields:
+/// // target = T::Value
+/// // ty     = T::Value
+///
+/// // does not yield T
+/// ```
+///
+/// The returned `GenericField` keeps both the schema type and the underlying field type.
+/// Usually those are the same:
+///
+/// ```ignore
+/// struct Wrapper<T> {
+///     value: T,
+/// }
+///
+/// // target = T
+/// // ty     = T
+/// ```
+///
+/// They differ when a field uses `#[wincode(with = ...)]`. In that case the
+/// generated code calls `SchemaRead`/`SchemaWrite` on the adapter type, but the
+/// adapter's `Src`/`Dst` must match the real field type:
+///
+/// ```ignore
+/// struct Wrapper<T> {
+///     #[wincode(with = "containers::Vec<_, BincodeLen>")]
+///     values: Vec<T>,
+/// }
+///
+/// // target = containers::Vec<T, BincodeLen>
+/// // ty     = Vec<T>
+/// ```
 pub(crate) fn generic_field_types(
     data: &Data<Variant, Field>,
     generics: &Generics,
-) -> std::collections::HashSet<GenericField> {
-    let type_params: std::collections::HashSet<&Ident> =
-        generics.type_params().map(|p| &p.ident).collect();
+) -> HashSet<GenericField> {
+    let type_params: HashSet<&Ident> = generics.type_params().map(|p| &p.ident).collect();
     if type_params.is_empty() {
-        return std::collections::HashSet::new();
+        return HashSet::new();
     }
 
     fn visit<'a>(
         fields: impl Iterator<Item = &'a Field>,
-        type_params: &std::collections::HashSet<&Ident>,
-    ) -> std::collections::HashSet<GenericField> {
-        let mut result = std::collections::HashSet::new();
+        type_params: &HashSet<&Ident>,
+    ) -> HashSet<GenericField> {
+        let mut result = HashSet::new();
         for field in fields {
             if field.skip.is_some() {
                 continue;
--- a/wincode-derive/src/common.rs
+++ b/wincode-derive/src/common.rs
@@ -845,12 +845,12 @@ impl Visit<'_> for HasLifetime {
     }
 }
 
-struct HasTypeParam {
-    type_params: std::collections::HashSet<Ident>,
+struct HasTypeParam<'a> {
+    type_params: &'a std::collections::HashSet<&'a Ident>,
     found: bool,
 }
 
-impl Visit<'_> for HasTypeParam {
+impl Visit<'_> for HasTypeParam<'_> {
     fn visit_ident(&mut self, i: &Ident) {
         if self.type_params.contains(i) {
             self.found = true;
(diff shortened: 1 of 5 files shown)


wincode-derive/src/common.rs
```diff
@@ -845,6 +845,51 @@ impl Visit<'_> for HasLifetime {
...
+    fn visit_ident(&mut self, i: &Ident) {
+        if self.type_params.contains(i) {
+            self.found = true;
+        }
+    }
+}
+
+pub(crate) fn generic_field_types(data: &Data<Variant, Field>, generics: &Generics) -> Vec<Type> {
+    let type_params: std::collections::HashSet<Ident> =
+        generics.type_params().map(|p| p.ident.clone()).collect();
```
cpubot: Unnecessary clone here. Could just return `&p.ident`

wincode-derive/src/common.rs
```diff
@@ -845,6 +845,51 @@ impl Visit<'_> for HasLifetime {
...
+    if type_params.is_empty() {
+        return Vec::new();
+    }
+    let fields: Vec<&Field> = match data {
```
cpubot: We don't need to allocate a `Vec` here.
```rs
fn visit<'a>(
    fields: impl Iterator<Item = &'a Field>,
    type_params: HashSet<&'a Ident>,
) -> Vec<Type> {
    // ...
}

match data {
    Data::Struct(fields) => visit(fields.iter(), type_params),
    Data::Enum(variants) => visit(variants.iter().flat_map(|v| v.fields.iter()), type_params),
}
```

wincode-derive/src/common.rs
```diff
@@ -845,6 +845,51 @@ impl Visit<'_> for HasLifetime {
...
+    };
+    let mut result = Vec::new();
+    let mut seen = std::collections::HashSet::new();
+    for field in &fields {
+        if field.skip.is_some() {
+            continue;
+        }
+        let target = field.target_resolved();
+        let mut visitor = HasTypeParam {
+            type_params: type_params.clone(),
```
cpubot: Don't need to clone here
```rs
struct HasTypeParam<'a> {
    type_params: &'a HashSet<&'a Ident>,
    found: bool,
}
```

wincode-derive/src/common.rs
```diff
@@ -845,6 +845,51 @@ impl Visit<'_> for HasLifetime {
...
+            found: false,
+        };
+        visitor.visit_type(&target);
+        if visitor.found {
+            let key = target.to_token_stream().to_string();
+            if seen.insert(key) {
```
cpubot: I don't see any reason not to just use the `Type` as the key and return the `HashSet` (rather than simultaneously using a `HashSet` to build a `Vec`).

```rs
seen.insert(target);
```

Though to fix the issue with `#[wincode(with)]`, you'll need the `target` and the `type`, which should still be fine as the key.

```rs
#[derive(Hash, PartialEq, Eq)]
pub(crate) struct GenericField {
    pub(crate) target: Type,
    pub(crate) ty: Type,
}
```

wincode-derive/src/schema_read.rs
```diff
@@ -319,15 +319,15 @@ fn append_config(generics: &mut Generics, crate_name: &Path) {
...
 
 fn append_where_clause(generics: &mut Generics, data: &Data<Variant, Field>, crate_name: &Path) {
+    let field_types = generic_field_types(data, generics);
     let mut predicates: Punctuated<WherePredicate, Token![,]> = Punctuated::new();
-    for param in generics.type_params() {
-        let ident = &param.ident;
+    for ty in &field_types {
         let mut bounds = Punctuated::new();
-        bounds.push(parse_quote!(#crate_name::SchemaRead<'de, __WincodeConfig, Dst = #ident>));
+        bounds.push(parse_quote!(#crate_name::SchemaRead<'de, __WincodeConfig, Dst = #ty>));
```
cpubot, line 326: This needs to actually use the `ty`, not the `target` (which `generic_field_types` currently returns).

IVSOP:
Thank you for the detailed feedback! Hopefully [a6b727a](https://github.com/anza-xyz/wincode/pull/341/commits/a6b727a002b382602c2540d687d0334c154f1041) fixes all the issues.
I've also enabled the `extra-traits` feature in `syn`. The code works without it as the feature is pulled by `darling`, but I think being explicit is better

wincode-derive/src/common.rs
```diff
@@ -845,6 +845,66 @@ impl Visit<'_> for HasLifetime {
...
+#[derive(Hash, PartialEq, Eq)]
+pub(crate) struct GenericField {
+    pub(crate) target: Type,
+    pub(crate) ty: Type,
+}
+
+pub(crate) fn generic_field_types(
+    data: &Data<Variant, Field>,
+    generics: &Generics,
+) -> std::collections::HashSet<GenericField> {
```
cpubot: Let's just pull in `HashSet` rather than fully qualifying each time

wincode/src/schema/mod.rs
```diff
@@ -4879,4 +4879,87 @@ mod tests {
...
+        }
+
+        let original = Both::<MyData> {
+            direct: MyData { value: 42 },
+            assoc: "hello".into(),
+        };
+        let serialized = serialize(&original).unwrap();
+        let deserialized: Both<MyData> = deserialize(&serialized).unwrap();
+        assert_eq!(original, deserialized);
+    }
```
cpubot, line 4964: We should also add a regression test, e.g., 
```rs
#[test]
fn test_generic_type_with_container_adapter() {
    #[derive(SchemaWrite, SchemaRead, Debug, PartialEq, Clone)]
    #[wincode(internal)]
    struct Wrapper<T> {
        #[wincode(with = "containers::Vec<_, BincodeLen>")]
        inner: Vec<T>,
    }

    let original = Wrapper::<u8> {
        inner: vec![1, 2, 3],
    };
    let serialized = serialize(&original).unwrap();
    let deserialized: Wrapper<u8> = deserialize(&serialized).unwrap();
    assert_eq!(original, deserialized);
}
```

wincode-derive/src/common.rs
```diff
@@ -845,6 +845,66 @@ impl Visit<'_> for HasLifetime {
...
+    }
+}
+
+#[derive(Hash, PartialEq, Eq)]
+pub(crate) struct GenericField {
+    pub(crate) target: Type,
+    pub(crate) ty: Type,
+}
+
+pub(crate) fn generic_field_types(
```
cpubot, line 886: Let's add a comprehensive comment here too. E.g.,
```rs
/// Find fields that mention one of the generic type parameters on the derived type.
///
/// The returned `GenericField` keeps two types:
/// - `target`: the type whose `SchemaRead`/`SchemaWrite` impl will be called
/// - `ty`: the actual type stored in the field
///
/// This is specifically based on the _fields_, not just the generic parameter list.
/// For example, if a type parameter is only used to name an associated type, we return the
/// associated type that is actually stored:
///
/// ```ignore
/// trait HasValue {
///     type Value;
/// }
///
/// struct Wrapper<T: HasValue> {
///     value: T::Value,
/// }
///
/// // yields:
/// // target = T::Value
/// // ty     = T::Value
///
/// // does not yield T
/// ```
///
/// The returned `GenericField` keeps both the schema type and the underlying field type.
/// Usually those are the same:
///
/// ```ignore
/// struct Wrapper<T> {
///     value: T,
/// }
///
/// // target = T
/// // ty     = T
/// ```
///
/// They differ when a field uses `#[wincode(with = ...)]`. In that case the
/// generated code calls `SchemaRead`/`SchemaWrite` on the adapter type, but the
/// adapter's `Src`/`Dst` must match the real field type:
///
/// ```ignore
/// struct Wrapper<T> {
///     #[wincode(with = "containers::Vec<_, BincodeLen>")]
///     values: Vec<T>,
/// }
///
/// // target = containers::Vec<T, BincodeLen>
/// // ty     = Vec<T>
/// ```
```

IVSOP:
Implemented your suggestions, and added another test that combines both features by having the container adapter over an associated type. I left the comment you suggested as-is but it I think it might be simplified to something like

```rs
/// Find fields whose types mention a generic type parameter of the derived type.
///
/// The returned [`GenericField`] keeps two types:
/// - `target`: the type whose `SchemaRead`/`SchemaWrite` impl will be called
/// - `ty`: the actual type stored in the field
/// 
/// `target` will be the same as `ty`, except for when a field uses `#[wincode(with = ...)]`:
///
/// ```ignore
/// struct Wrapper<T> {
///     #[wincode(with = "containers::Vec<_, BincodeLen>")]
///     values: Vec<T>,
/// }
/// // target = containers::Vec<T, BincodeLen>
/// // ty     = Vec<T>
/// ```
/// 
/// When this happens, the generated code calls `SchemaRead`/`SchemaWrite` on the adapter type,
/// but the adapter's `Src`/`Dst` must match the real field type.
```

IVSOP:
I prefer shorter comments so I've changed it :)

cpubot (approved):
Nice work 👍

PR #345, opened 2026-06-27
feat: add FromInto schema adapter for #[wincode(with = ...)]

kskalski:
#### Problem
There are a few cases observed migrating to wincode, where structs are serialized using intermediate / helper / legacy format structs that are stright-forward convertible to/from available (current) structs. Serialization of that requires boilerplate schemas to be implemented, which isn't too difficult, but could be simplified by just delegating to standard `From` / `Into` traits.

Serde has dedicated macro support for that (`#[serde(from = "String", into = "String")]`) at type level, as well as field level `serde_with::{serde_as, FromInto}`.

#### Summary of Changes
Serializes a target type by converting it to/from the value of an
intermediate wire schema via From/Into. The wire parameter may be a
self-describing type or a schema adapter over a different value type.

### DIFF
--- a/wincode/src/schema/adapter.rs
+++ b/wincode/src/schema/adapter.rs
@@ -1,0 +1,192 @@
+//! Schema adapters that serialize a type by converting it to and from the value
+//! of an intermediate "wire" schema.
+
+use {
+    crate::{
+        TypeMeta,
+        config::ConfigCore,
+        error::{ReadResult, WriteResult},
+        io::{Reader, Writer},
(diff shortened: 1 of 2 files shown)


kskalski:
Motivational example: https://github.com/anza-xyz/agave/pull/13515/changes#diff-4203fa5a6be1766c16ebf8aad752b470a45d6225978ec840e38fafbcaee40782R431-R441

cpubot (approved):
LGTM

PR #346, opened 2026-06-29
feat(containers): add public encode_from_iter

kskalski:
#### Problem
Non-materialized iterators are difficult to use during serialization

#### Summary of Changes
Export the iterator-encoding primitives publicly as the write-side counterparts to `decode_into_slice_t`:
 - `encode_from_iter_prealloc_check` (preferred): writes a length-prefixed sequence and validates the length against the configured preallocation limit, guaranteeing the wire format is accepted on read-back under the same configuration.
- `encode_from_iter`: the same, but skips the preallocation check for callers that must bypass it or have already performed it.
    
Both accept any `IntoIterator` whose iterator is an `ExactSizeIterator`, so a sequence can be encoded from an ad-hoc iterator (range, map chain, ...) without materializing it into a collection.
    
Widen `write_elem_iter_prealloc_check` to accept `Item: Borrow<T::Src>` (instead of `&'a T::Src`) to support such iterators; the existing slice-based caller is unaffected.

### DIFF
--- a/wincode/src/schema/containers.rs
+++ b/wincode/src/schema/containers.rs
@@ -72,7 +72,10 @@ use {
 use {
     crate::{
         context,
-        schema::{SchemaReadContext, size_of_elem_slice, write_elem_slice_prealloc_check},
+        schema::{
+            SchemaReadContext, size_of_elem_slice, write_elem_iter_prealloc_check,
+            write_elem_slice_prealloc_check,
+        },
     },
     alloc::{boxed::Box as AllocBox, collections, rc::Rc as AllocRc, vec},
 };
@@ -577,6 +580,11 @@ impl<T, E, I> CollectResultExt<T, E> for I where I: Iterator<Item = Result<T, E>
 /// container relies on the collection's [`FromIterator`] impl rather than
 /// writing directly into preallocated memory.
 ///
+/// For manual [`SchemaWrite`] implementations that emit a sequence from an
+/// ad-hoc iterator rather than a concrete collection, see
+/// [`encode_from_iter_prealloc_check`] (and its check-skipping counterpart
+/// [`encode_from_iter`]).
+///
 /// # Allocation efficiency
 ///
 /// During deserialization, the iterator passed to [`FromIterator`] has a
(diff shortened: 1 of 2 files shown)


kskalski:
A couple of notes:
* not sure if we should expose the raw `write_elem_iter` without the pre-alloc check - in theory this APIs could be used in context where alloc checks are not relevant, i.e. manual implementation could skip both materializations - on write as well as on read
* we could possibly also add companion `decode_as_iter` API
* alternative approach is to define `IteratorView` schema that wraps iterator and serializes using `FromIntoIterator` - this is basically how the functionality can be implemented right now https://github.com/anza-xyz/agave/pull/13515/changes#diff-019c659463d2220ed6546c357d570f888159d0acff47e7949696a4024b644acfR657-R699 (other than the actual code from `write_elem_iter`) - I'm not convinced that adding `IteratorView` to library is better than exposing free functions, but maybe
* the issue with using `FromIntoIterator` directly on lazy iterators is the `for<'a> &'a Coll: IntoIterator<…, IntoIter: ExactSizeIterator>` bound - it was designed for owned collections, `&iterator` doesn't typically satisfy `IntoIterator`

kskalski:
I will research the `IteratorView` approach a bit more, since now we moved to struct / derive impl for place where lazy iterator is used, so we might actually want to support adapter that stores an iterator and provides (write)schema for it...

wincode/src/schema/containers.rs
```diff
@@ -754,3 +760,70 @@ where
...
+/// encode_from_iter::<u32, Len, C>(buf.by_ref(), (0u32..6).map(|i| i * i)).unwrap();
+///
+/// assert_eq!(wincode::deserialize::<Vec<u32>>(&buf).unwrap(), [0, 1, 4, 9, 16, 25]);
+/// # }
+/// ```
+#[cfg(feature = "alloc")]
+#[inline]
+pub fn encode_from_iter<T, Len, C>(
+    writer: impl Writer,
+    src: impl ExactSizeIterator<Item: Borrow<T::Src>>,
```
brooksprumo: In general I've found it (slightly) easier to work with functions that take their iterator parameter as `IntoIterator` instead of `Iterator`. Would something like that work here as well?

E.g.:

```rust
impl IntoIterator<Item = X, IntoIter: ExactSizeIterator>,
```

(Where `X` would need to be the actual Item type, of course.)

kskalski: Agreed, done

brooksprumo: Changes lgtm, but looks like I cannot mark this comment thread as resolved.

brooksprumo:
I un-requested my review, as I don't really know the details for wincode. The IntoIterator changes lgtm though!

cpubot (approved):

PR #347, opened 2026-06-30
fix: bv::BitVec canonical decoding

cpubot:
- `block_count` must be exactly the number of blocks needed to hold `bit_len`
- The padding region of the final block (`bit_len % Block::nbits() .. Block::nbits()`) must be all zeros.

### DIFF
--- a/wincode/src/schema/external/bv.rs
+++ b/wincode/src/schema/external/bv.rs
@@ -2,13 +2,20 @@ use {
     crate::{
         ReadResult, SchemaRead, SchemaWrite, TypeMeta, WriteResult,
         config::Config,
-        error::invalid_tag_encoding,
+        error::{invalid_tag_encoding, invalid_value},
         io::{Reader, Writer},
         len::SeqLen,
     },
     bv::{BitVec, Bits, BlockType},
     core::mem::MaybeUninit,
 };
+
+#[inline]
+fn expected_block_count<Block: BlockType>(bits_len: u64) -> ReadResult<usize> {
+    let block_bits = Block::nbits() as u64;
+    let block_count = bits_len.div_ceil(block_bits);
+    usize::try_from(block_count).map_err(|_| invalid_value("BitVec block count overflow"))
+}
 
 unsafe impl<C: Config, Block: BlockType + SchemaWrite<C, Src = Block>> SchemaWrite<C>
     for BitVec<Block>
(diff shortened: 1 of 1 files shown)


wincode/src/schema/external/bv.rs
```diff
@@ -2,14 +2,27 @@ use {
...
+    let block_bits = Block::nbits() as u64;
+    let full_blocks = bits_len / block_bits;
+    let needs_partial_block = bits_len % block_bits != 0;
+    // `bits_len` is attacker-controlled, so avoid `bits_len + block_bits - 1`.
+    // This addition cannot overflow: `block_bits >= 8`, so `full_blocks` is at
+    // most `u64::MAX / 8`, and the addend is only 0 or 1.
+    let block_count = full_blocks + u64::from(needs_partial_block);
+    usize::try_from(block_count).map_err(|_| invalid_value("BitVec block count overflow"))
+}
+
```
kskalski, line 19: this should suffice as I think `div_ceil` will round toward max
```suggestion
#[inline]
fn expected_block_count<Block: BlockType>(bits_len: u64) -> ReadResult<usize> {
    let block_bits = Block::nbits() as u64;
    let block_count = bits_len.div_ceil(block_bits);
    usize::try_from(block_count).map_err(|_| invalid_value("BitVec block count overflow"))
}
```

cpubot: Yep, that's better for sure. Updated

wincode/src/schema/external/bv.rs
```diff
@@ -26,7 +33,7 @@ unsafe impl<C: Config, Block: BlockType + SchemaWrite<C, Src = Block>> SchemaWri
             len_size + if let TypeMeta::Static { size, zero_copy: _ } = Block::TYPE_META {
                 n_blocks * size
             } else {
-                (0..n_blocks).map(|i| Block::size_of(&src.get_raw_block(i))).try_fold(0, |acc, r| r.map(|n| acc + n))?
```
kskalski, line 29: is there a specific concern with using raw_block?

cpubot: For a final partial block, `get_raw_block(i)` returns the backing storage verbatim, including bits beyond the logical `BitVec::len()`. Those bits are ignored by logical equality but still serialized. So a normal in-memory value can serialize non-canonically after operations like:


```rust
let mut bv: BitVec<u8> = BitVec::from_bits(vec![true; 8]);
bv.truncate(3);
```

`get_block(i)` returns the block masked to the logical bit length, so final-block padding is serialized as zero. For full blocks it is equivalent to `get_raw_block(i)`

Using `get_block()` makes the writer canonical and keeps it compatible with the strict decoder

kskalski (approved):

PR #359, opened 2026-07-06
feat(schema): add Discard and DiscardVec adapters for ignored fields

kskalski:
#### Problem
As wire format schemas evolve, there are cases where given value stops being populated / used and should be ignored, while preserving backwards compatibility, e.g.
https://github.com/anza-xyz/agave/blob/e7b03044cf8bc1225eb38b3061348a0117459be6/runtime/src/epoch_stakes.rs#L465

#### Summary of Changes
Provide adapters for fields that should no longer be carried, such as a deprecated field kept only for wire compatibility. On read the encoded value is consumed and validated but reset to its Default; on write the held value is ignored and the Default is serialized. This keeps the stream in sync and both directions consistent while treating the field as absent.
    
`DiscardSeq<T, Len>` does the same for any length-prefixed sequence without allocating a backing buffer, where T is the element schema. Since `Vec`, `String`, sets and maps all share the length-prefix + elements wire layout, one adapter covers them all (a `String` via `DiscardSeq<u8, _>`, a map via `DiscardSeq<(K, V), _>`), and it nests: `DiscardSeq<DiscardSeq<u8, L>, L>` discards a `Vec<Vec<u8>>` with no allocation at any level. The yielded `Vec` type is incidental, so deprecating a field is just repointing it at `DiscardSeq`.

### DIFF
--- a/wincode/src/schema/adapter.rs
+++ b/wincode/src/schema/adapter.rs
@@ -1,6 +1,8 @@
 //! Schema adapters that serialize a type by converting it to and from the value
 //! of an intermediate "wire" schema.
 
+#[cfg(feature = "alloc")]
+use crate::len::SeqLen;
 use {
     crate::{
         TypeMeta,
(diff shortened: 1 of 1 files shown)


kskalski:
Rebased and did a couple of updates:
* renamed `DiscardVec` to `DiscardSeq` - it still outputs `Vec`, but it is an universal sequence bytes consumer, so I documented its use for various collections
* added tests for `String` to `DiscardVec<u8>`
* added tests for nesting schema (or other `DiscardVec`) inside `DiscardVec`

wincode/src/schema/adapter.rs
```diff
@@ -222,12 +224,247 @@ where
...
+where
+    C: ConfigCore,
+    Inner: SchemaRead<'de, C>,
+    Inner::Dst: Default,
+{
+    type Dst = Inner::Dst;
+
+    // `zero_copy` is cleared: the value is a fresh `Default`, not a view into the
+    // wire bytes. The (unchanged) size is forwarded from `Inner`.
+    const TYPE_META: TypeMeta = <Inner as SchemaRead<'de, C>>::TYPE_META.keep_zero_copy(false);
```
cpubot: Nit: `Inner` is already constrained as `SchemaRead<'de, C>`, so I believe we can omit the qualification here

```suggestion
    const TYPE_META: TypeMeta = Inner::TYPE_META.keep_zero_copy(false);
```

wincode/src/schema/adapter.rs
```diff
@@ -222,12 +224,247 @@ where
...
+        dst.write(Default::default());
+        Ok(())
+    }
+}
+
+unsafe impl<C, Inner> SchemaWrite<C> for Discard<Inner>
+where
+    C: ConfigCore,
+    Inner: SchemaWrite<C>,
+    Inner::Src: Sized + Default,
```
cpubot: nit: `Default` already requires `Sized`, so we technically don't need the `Sized` bound here explicitly

wincode/src/schema/adapter.rs
```diff
@@ -222,12 +224,247 @@ where
...
+    #[inline]
+    fn read(reader: impl Reader<'de>, dst: &mut MaybeUninit<Self::Dst>) -> ReadResult<()> {
+        // Decode with the inner schema to advance the reader, then drop the value.
+        // No byte-skip fast path (unlike `DiscardSeq`): a single value has no
+        // per-element bounds checks to hoist, and dropping the decoded value lets
+        // the optimizer elide the copy out of the reader.
+        //
+        // `?` propagates before `dst` is written, so `dst` stays uninitialized on
+        // error, upholding the `read` safety contract.
+        <Inner as SchemaRead<'de, C>>::get(reader)?;
```
cpubot: Similar to the above, `Inner` is already qualified as `SchemaRead<'de, C>`, so we don't need to qualify here
```suggestion
        Inner::get(reader)?;
```

wincode/src/schema/adapter.rs
```diff
@@ -222,12 +224,247 @@ where
...
+    // representation doesn't match the wire bytes. The size is forwarded from
+    // `Inner` (its `Default` encodes to the same size).
+    const TYPE_META: TypeMeta = <Inner as SchemaWrite<C>>::TYPE_META.keep_zero_copy(false);
+
+    #[inline]
+    fn size_of(_src: &Self::Src) -> WriteResult<usize> {
+        if let TypeMeta::Static { size, .. } = <Self as SchemaWrite<C>>::TYPE_META {
+            return Ok(size);
+        }
+        <Inner as SchemaWrite<C>>::size_of(&Default::default())
```
cpubot: Similar to the above about qualifying
```suggestion
        Inner::size_of(&Default::default())
```

wincode/src/schema/adapter.rs
```diff
@@ -222,12 +224,247 @@ where
...
+    }
+
+    #[inline]
+    fn write(writer: impl Writer, _src: &Self::Src) -> WriteResult<()> {
+        // Ignore the provided value; serialize `Inner`'s `Default` instead,
+        // matching what `read` reconstructs.
+        <Inner as SchemaWrite<C>>::write(writer, &Default::default())
```
cpubot:
```suggestion
        Inner::write(writer, &Default::default())
```

wincode/src/schema/adapter.rs
```diff
@@ -222,12 +224,247 @@ where
...
+    type Src = Inner::Src;
+
+    // `zero_copy` is cleared: the instance value is ignored, so the in-memory
+    // representation doesn't match the wire bytes. The size is forwarded from
+    // `Inner` (its `Default` encodes to the same size).
+    const TYPE_META: TypeMeta = <Inner as SchemaWrite<C>>::TYPE_META.keep_zero_copy(false);
+
+    #[inline]
+    fn size_of(_src: &Self::Src) -> WriteResult<usize> {
+        if let TypeMeta::Static { size, .. } = <Self as SchemaWrite<C>>::TYPE_META {
```
cpubot:
```suggestion
        if let TypeMeta::Static { size, .. } = Self::TYPE_META {
```

wincode/src/schema/adapter.rs
```diff
@@ -222,12 +224,247 @@ where
...
+    C: ConfigCore,
+    Len: SeqLen<C>,
+    T: SchemaRead<'de, C>,
+{
+    type Dst = alloc::vec::Vec<T::Dst>;
+
+    #[inline]
+    fn read(mut reader: impl Reader<'de>, dst: &mut MaybeUninit<Self::Dst>) -> ReadResult<()> {
+        let len = Len::read_prealloc_check::<T::Dst>(reader.by_ref())?;
+        if let TypeMeta::Static { size, .. } = <T as SchemaRead<'de, C>>::TYPE_META {
```
cpubot:
```suggestion
        if let TypeMeta::Static { size, .. } = T::TYPE_META {
```

wincode/src/schema/adapter.rs
```diff
@@ -222,12 +224,247 @@ where
...
+            // Reserve the whole `len * size`-byte body up front so the per-element
+            // decodes below elide their own bounds checks. `get` still validates
+            // each element, so non-zero-copy types are handled correctly too.
+            //
+            // SAFETY: exactly `len` decodes of `size` bytes run through the trusted
+            // window, consuming `len * size` bytes total — never more.
+            let mut trusted = unsafe { reader.as_trusted_for_seq(len, size)? };
+            for _ in 0..len {
+                <T as SchemaRead<'de, C>>::get(trusted.by_ref())?;
```
cpubot:
```suggestion
                T::get(trusted.by_ref())?;
```

wincode/src/schema/adapter.rs
```diff
@@ -222,12 +224,247 @@ where
...
+            }
+        } else {
+            // Dynamically-sized elements: decode and drop each.
+            for _ in 0..len {
+                <T as SchemaRead<'de, C>>::get(reader.by_ref())?;
```
cpubot:
```suggestion
                T::get(reader.by_ref())?;
```

cpubot (commented):
A few style nits.

Was hoping we'd be able to avoid the hard `Vec` destination for `DiscardSeq`, but seeing as we don't have higher-kinded types in Rust, this is likely the best solution.

wincode/src/schema/adapter.rs
kskalski: done

wincode/src/schema/adapter.rs
kskalski: done

wincode/src/schema/adapter.rs
kskalski: done

wincode/src/schema/adapter.rs
kskalski: done

wincode/src/schema/adapter.rs
kskalski: done

wincode/src/schema/adapter.rs
kskalski: done

wincode/src/schema/adapter.rs
kskalski: done

wincode/src/schema/adapter.rs
kskalski: done

wincode/src/schema/adapter.rs
kskalski: done

kskalski:
Cleanedup and trimmed down a few comments too.

cpubot (approved):

PR #363, opened 2026-07-07
feat(schema): add DefaultOnEmptyRead adapter

kskalski:
Provides EOF-fallback decoding so trailing struct fields appended in a later version still decode from older, shorter encodings, filling the missing fields with their Default. Writing is unchanged.

This adapter was duplicated verbatim in several agave crates (storage-proto, ledger - more are about to come, e.g. https://github.com/anza-xyz/agave/pull/12489 and https://github.com/anza-xyz/agave/blob/faa63fed27999b27e3d03f7bd435864f7c743c0e/runtime/src/serde_snapshot.rs#L91-L96); moving it into the core library lets those sites share one implementation.

### DIFF
--- a/wincode/src/schema/adapter.rs
+++ b/wincode/src/schema/adapter.rs
@@ -35,7 +35,7 @@ use {
 ///
 /// ```
 /// # #[cfg(feature = "derive")] {
-/// # use wincode::FromInto;
+/// # use wincode::adapter::FromInto;
 /// # use wincode_derive::{SchemaWrite, SchemaRead};
 /// // Our in-memory type with a layout we don't want to serialize directly.
 /// #[derive(Debug, PartialEq, Clone, Copy)]
@@ -131,7 +131,7 @@ where
 ///
 /// ```
 /// # #[cfg(feature = "derive")] {
-/// # use wincode::DefaultOnEmptyRead;
+/// # use wincode::adapter::DefaultOnEmptyRead;
 /// # use wincode_derive::{SchemaWrite, SchemaRead};
 /// #[derive(SchemaWrite, SchemaRead, Debug, PartialEq)]
 /// struct Record {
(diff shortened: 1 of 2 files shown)


wincode/src/schema/mod.rs
```diff
@@ -58,7 +58,7 @@ mod adapter;
 mod compile_fail;
 pub mod containers;
 pub mod context;
-pub use adapter::FromInto;
+pub use adapter::{DefaultOnEmptyRead, FromInto};
```
cpubot: What do you think about just exporting `pub use adapter;`? The more we add, the more we're just polluting `wincode::`. That way they live at `wincode::adapter`, akin to `wincode::containers`.

kskalski: right, I was thinking the same after looking at that diff and more adapter coming in https://github.com/anza-xyz/wincode/pull/359
added in [22d4237](https://github.com/anza-xyz/wincode/commit/22d4237b6d5a1d26b0a742fa8fc4db5fe757db3f)

wincode/src/schema/adapter.rs
```diff
@@ -113,9 +113,122 @@ where
...
+///   partially-present value, so a genuinely truncated field is masked instead
+///   of reported and the fields after it are misaligned. It is only safe on a
+///   trailing run of fields where every field from the first
+///   `DefaultOnEmptyRead` onward is itself optional-on-EOF.
+///
+/// Prefer applying it to trailing fields of a **top-level struct** decoded with
+/// [`deserialize_exact`](crate::deserialize_exact), where reaching EOF exactly
+/// at a field boundary is well defined and the end-of-input check is
+/// straightforward and cheap.
+pub struct DefaultOnEmptyRead<T>(PhantomData<T>);
```
cpubot, line 178: Non-blocking style nit: what do you think about `DefaultOnEof`? A little more concise

cpubot (approved):

wincode/src/schema/adapter.rs
kskalski: I think current name is more direct and clear, Eof is well established condition name, but it still raises a question / doubt on what wincode considers as end of file.

PR #370, opened 2026-07-10
fix(derive): compute zero-copy layout against `from` target

kskalski:
#### Problem
`#[wincode(from = "T")]` classifies zero-copy using the schema struct layout instead of the target `T`. A schema that does not cover all of `T` is marked zero-copy with a too-small size, so the slice readers borrow fewer bytes than the `&[T]` they produced, reading/writing past the input buffer.

#### Summary of Changes
* use target type when comparing its size to size of fields
* pass-through info to calculate correct zero-copy flag when generating TYPE_META fields

### DIFF
--- a/wincode-derive/src/common.rs
+++ b/wincode-derive/src/common.rs
@@ -203,6 +203,7 @@ pub(crate) trait FieldsExt {
         trait_impl: TraitImpl,
         repr: &StructRepr,
         crate_name: &Path,
+        dst: &Type,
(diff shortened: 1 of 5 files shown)


kskalski:
Agreed. Given how schemas can target foreign types, it seems this isn't by any means critical functionality. What I could imagine useful is generating mapping types with zero-copy references for fields that support them, but that's a completely different story.

PR #371, opened 2026-07-11
chore(derive)!: remove derive `from` attribute

cpubot:
This was useful before wincode got traction and integration with the solana-sdk, but now we're maintaining (nontrivial) functionality that simply isn't used anywhere inside our suite of projects. Potentially unfortunate for some downstream users (though frankly the use-case is quite niche, so I doubt it's actually used extensively in the wild), but removing eliminates a whole set of maintenance burden for us.

### DIFF
--- a/wincode-derive/src/common.rs
+++ b/wincode-derive/src/common.rs
@@ -470,43 +470,6 @@ impl VariantsExt for &[Variant] {
 
 pub(crate) type ImplBody = Data<Variant, Field>;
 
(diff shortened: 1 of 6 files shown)


kskalski (approved):

steviez:
Adding a fly-by note - looks like we should probably bump the major version for next wincode release

kskalski:
Since we are still in 0.x stage, that basically means a new "minor" number, i.e. wincode-derive gets 0.5.0 version.

PR #372, opened 2026-07-11
feat(schema): impl for `(T,)`

cpubot:


wincode/src/schema/impls.rs
```diff
@@ -489,6 +489,41 @@ unsafe impl<'de, C: ConfigCore> SchemaRead<'de, C> for () {
...
+where
+    T: SchemaRead<'de, C>,
+{
+    type Dst = (T::Dst,);
+
+    const TYPE_META: TypeMeta = T::TYPE_META.keep_zero_copy(false);
+
+    #[inline]
+    fn read(reader: impl Reader<'de>, dst: &mut MaybeUninit<Self::Dst>) -> ReadResult<()> {
+        let slot = unsafe { &mut *(&raw mut (*dst.as_mut_ptr()).0).cast() };
```
kskalski, line 522: something that came up when I was looking into `from=` fixes for field initialization - would it make sense to add a helper that would facilitate getting uninit of a field with a forced type (I think after getting rid of `from=` we will always know the actual type, but maybe the macro can also be invoked with `_`)?
```rust
/// Projects a `&mut MaybeUninit<Outer>` place expression down to `&mut MaybeUninit<Field>`
/// for the given field, e.g. `uninit_field!(dst, 0, T::Dst)` or
/// `uninit_field!(dst, some_field, SomeFieldTy)`.
///
/// SAFETY (internal): sound because `&raw mut (*p).field` is a genuine field projection
/// (correct offset, correctly aligned for `Field`), and `MaybeUninit<Field>` has the same
/// layout as `Field`, so reinterpreting the field pointer as `&mut MaybeUninit<Field>` is
/// valid, PROVIDED `$ty` actually matches the field's real type — the macro can't check
/// this itself; a mismatched `$ty` would be silently unsound (wrong cast size/align).
macro_rules! uninit_field {
    ($dst:expr, $field:tt, $ty:ty) => {
        unsafe { &mut *(&raw mut (*$dst.as_mut_ptr()).$field).cast::<MaybeUninit<$ty>>() }
    };
}

    let slot = uninit_field!(dst, 0, T::Dst);
    T::read(reader, slot)
}
```

not sure if we could also use that in generated code, but it would make a lot of it cleaner

kskalski (approved):

wincode/src/schema/impls.rs
cpubot: Yeah, I agree it would be nice to have a helper that does all of this. I'm apprehensive about the macro route because it hides `unsafe` usage and is easily misused, but I don't think there's really a great type-safe solution

PR #375, opened 2026-07-13
feat(bv): make canonical encoding/decoding opt-in

cpubot:
https://github.com/anza-xyz/wincode/pull/347 introduced canonical encoding / decoding for `bv::Bitvec`. This is against wincode's _compatible with bincode by default_ goal. With strict decoding, it is possible to reject valid representations produced by the serde implementation, as it does not canonicalize in serialization. I think the safer route is to make canonical decoding opt-in and bincode/serde compatibility the default.


wincode/src/schema/external/bv.rs
```diff
@@ -228,10 +284,9 @@ mod tests {
             let bincode_serialized = bincode::serialize(&bv).unwrap();
             let schema_serialized = serialize(&bv).unwrap();
             prop_assert_eq!(serialized_size(&bv).unwrap() as usize, schema_serialized.len());
-            prop_assert_eq!(&bincode_serialized, &schema_serialized);
```
kskalski, line 231: this should still pass for the normalized bv produced from `bits`, right?

wincode/src/schema/external/bv.rs
```diff
@@ -135,88 +171,108 @@ mod tests {
...
+        assert_eq!(bincode_deserialized, bv);
 
-        assert_ne!(canonical, malleable);
+        let bincode_bytes = bincode::serialize(&bv).unwrap();
 
-        let from_canonical: BitVec<u8> = deserialize(&canonical).unwrap();
-        assert_eq!(from_canonical, bv);
-        assert!(deserialize::<BitVec<u8>>(&malleable).is_err());
+        #[cfg(feature = "bv-strict")]
+        assert!(deserialize::<BitVec<u8>>(&bincode_bytes).is_err());
```
kskalski, line 195: we could also assert the we can deserialize wincode bytes

wincode/src/schema/external/bv.rs
```diff
@@ -135,88 +171,108 @@ mod tests {
...
-        assert_eq!(
-            serialized_size(&bv_from_bits).unwrap() as usize,
-            serialized_from_bits.len()
-        );
-        let deserialized_from_bits: BitVec = deserialize(&serialized_from_bits).unwrap();
-        assert_eq!(bv_from_bits, deserialized_from_bits);
-    }
+    fn test_bitvec_padding() {
+        let mut bv = BitVec::<u8>::from(vec![0xff]);
+        bv.truncate(3);
```
kskalski, line 176: we could trigger check on how the strict mode influences serializing the blocks - with varint encoding that differs between strict/non-strict and we should make sure `serialized_size` matches actual encoded `len()` 

```rust
#[test]
fn test_bitvec_varint_dirty_padding() {
    let mut bv = BitVec::<usize>::from(vec![usize::MAX]);
    bv.truncate(3);
    let c = Configuration::default().with_varint_encoding();
    let bincode_c = bincode::DefaultOptions::new().with_varint_encoding();

    let schema_serialized = config::serialize(&bv, c).unwrap();
    assert_eq!(config::serialized_size(&bv, c).unwrap() as usize, schema_serialized.len());
    let bincode_deserialized: BitVec = bincode_c.deserialize(&schema_serialized).unwrap();
    assert_eq!(bincode_deserialized, bv);

    let bincode_serialized = bincode_c.serialize(&bv).unwrap();
    #[cfg(feature = "bv-strict")]
    assert!(config::deserialize::<BitVec, _>(&bincode_serialized, c).is_err());
    #[cfg(not(feature = "bv-strict"))]
    assert_eq!(config::deserialize::<BitVec, _>(&bincode_serialized, c).unwrap(), bv);
}
```

wincode/src/schema/external/bv.rs
cpubot: Yeah, shouldn't have been deleted. Fixed

wincode/src/schema/external/bv.rs
cpubot: Done

wincode/src/schema/external/bv.rs
cpubot: Good call --added

kskalski (approved):

PR #377, opened 2026-07-13
chore(deps): remove solana-short-vec

cpubot:
It's been deprecated -- let's remove it before 0.6.0


kskalski:
> It's been deprecated -- let's remove it before 0.6.0

 Are we going to bump wincode to 0.6? Is that required, I thought all breaking changes were only in derive and I imagined that we can do a patch for wincode while moving wincode-derive to 0.5

wincode/src/schema/mod.rs
```diff
@@ -1635,46 +1635,6 @@ mod tests {
         }
     }
 
-    #[test]
-    #[cfg(feature = "solana-short-vec")]
-    fn test_uninit_builder_with() {
```
kskalski, line 1640: ok - we still do have coverage with `test_uninit_builder_with_container`

kskalski (approved):
Ah, this https://github.com/anza-xyz/wincode/commit/3d63dcfe1027fb9368e88edad916ea3c028f142f change is breaking for `wincode`

PR #381, opened 2026-07-14
feat(io)!: rework `Reader` trait to be unsafe + required method

cpubot:
This PR proposes two changes to the `Reader` trait.

### `Reader` becomes an `unsafe trait`.

The current required method is `copy_into_slice`, which has the following signature:
```rs
fn copy_into_slice(&mut self, dst: &mut [MaybeUninit<u8>]) -> ReadResult<()>;
```

Safe code relies on this method fully initializing the `dst` on success, but this postcondition is never encoded as a safety obligation.

As such, this PR updates the `Reader` trait to be `unsafe` to implement, and specifies that initialization post-condition explicitly under its safety contract.

### `Reader`'s required method, `copy_into_slice`, now accepts `&mut [u8]`.

This turns out to be the better default than `&mut [MaybeUninit<u8>]`. 

Often times, user code will have access to a `&mut [u8]`, but cannot use `copy_into_slice` without using either `from_raw_parts_mut` or `transmute` on it, which is poor ergonomics. Furthermore, this conversion is lossy at the type level -- the given slice may initialized, but downstream code must treat it as uninitialized, which can result in double initialization. For example, consider the std io `ReadAdapter` (abridged):

```rs
impl<R: Read + ?Sized> Reader<'_> for ReadAdapter<R> {
    #[inline(always)]
    fn copy_into_slice(&mut self, dst: &mut [MaybeUninit<u8>]) -> ReadResult<()> {
        // Choice here -- do we assume that `R: Read` doesn't read the contents of `dst`, or 
        // explicitly zero-initialize?
        //
        // Choice 1: assume `R: Read` doesn't read the contents (potentially exposing uninit mem)
        let buf = unsafe { transmute::<&mut [MaybeUninit<u8>], &mut [u8]>(dst) };
        // Choice 2: make no assumptions, zero-initialize the `dst`
        buf.fill(MaybeUninit::new(0));

        if let Err(e) = reader.read_exact(buf) {
            return maybe_eof_to_read_size_limit(e, buf.len());
        };
        Ok(())
    }
}
```

Now consider the calling code:
```rs
let mut reader = ReadAdapter::new(/* */);
let mut buf = [0u8; 32];
// Cannot pass `buf` directly.
let dst = unsafe { transmute::<&mut [u8], &mut [MaybeUninit<u8>]>(buf.as_mut_slice()) };
reader.copy_into_slice(dst);
```

If `ReadAdapter` chooses to zero initialize `dst` (choice 2), the already zeroed `buf` is zeroed again.

`copy_into_slice` accepting `&mut [u8]` resolves this completely.

This PR makes this adjustment, and additionally exposes `copy_into_uninit_slice`, which has the same signature as the current `copy_into_slice`. By default, it zeroes the given buffer and delegates to `copy_into_slice`.

```rs
fn copy_into_uninit_slice(&mut self, dst: &mut [MaybeUninit<u8>]) -> ReadResult<()> {
    dst.fill(MaybeUninit::new(0));
    // SAFETY: Every element is initialized with a valid `u8`, and
    // `MaybeUninit<u8>` has the same layout as `u8`.
    let dst = unsafe { transmute::<&mut [MaybeUninit<u8>], &mut [u8]>(dst) };
    self.copy_into_slice(dst)
}
```

Implementations that can support writing directly into uninitialized storage are updated to do so.

### DIFF
--- a/wincode/src/io/mod.rs
+++ b/wincode/src/io/mod.rs
@@ -70,7 +70,29 @@ pub(super) const fn transpose<const N: usize, T>(
 /// - Callers should prefer [`Reader::copy_into_slice`] or `take_*` methods to remain
 ///   compatible with readers that don't support borrowing, if possible.
 /// - Returns [`ReadError::UnsupportedBorrow`] for readers that do not support borrowing.
-pub trait Reader<'a> {
+///
+/// # Performance
+///
+/// The default [`Reader::copy_into_uninit_slice`] implementation initializes
+/// the destination before delegating to [`Reader::copy_into_slice`].
+/// Implementors that can safely write directly into `MaybeUninit<u8>` storage
+/// should override it to avoid this initialization. Such overrides must still
+/// initialize every destination byte before returning `Ok(())`.
+///
+/// # Safety
+///
+/// Implementors must ensure that:
+///
+/// - If `copy_into_uninit_slice` returns `Ok(())`, every element of the
+///   destination has been initialized.
+/// - All safe methods uphold Rust's validity, aliasing, and lifetime rules for
+///   every safe input. Callers are not responsible for proving that destination
+///   buffers do not overlap internal storage.
+/// - References returned by borrowing methods remain valid for their documented
+///   lifetimes.
+/// - Any unchecked reader returned by `as_trusted_for` obeys that method's
+///   documented bounds contract.
+pub unsafe trait Reader<'a> {
     /// Borrow capabilities of this reader.
     ///
     /// A bitmask of [`BorrowKind`] values indicating which kinds of borrows are supported.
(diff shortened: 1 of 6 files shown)


wincode/src/io/mod.rs
```diff
@@ -260,44 +285,82 @@ pub trait Reader<'a> {
...
-    /// Copy and consume exactly `dst.len()` bytes from the [`Reader`] into `dst`.
+    /// Attempts to copy and consume exactly `dst.len()` bytes.
     ///
-    /// # Safety
+    /// On success, exactly `dst.len()` bytes are consumed and every element of
+    /// `dst` contains the corresponding byte. On error, some bytes may have been
+    /// consumed; `dst` remains initialized, but its contents are unspecified.
+    fn copy_into_slice(&mut self, dst: &mut [u8]) -> ReadResult<()>;
+
+    /// Attempts to copy and consume exactly `dst.len()` bytes into potentially
```
kskalski, line 295: Nit: we could somehow quality that "consume" refers to the input reader is reading (or maybe replace them we "read"). I would also suggest reordering the phrasing: 
```suggestion
    /// Attempts to consume `dst.len()` bytes from input and copy them into potentially
```

kskalski (approved):
Just non-blocking suggestion about comments, which had the same form before - we could apply better wording about "copy and consume" across the file in follow-up

PR #388, opened 2026-07-16
fix assert_zero_copy trait metadata assertion

### LINKED ISSUE #383: `#[wincode(assert_zero_copy)` asymmetry
When deriving implementations for `SchemaWrite` together with the `#[wincode(assert_zero_copy)]` directive, the implementation only considers the `SchemaRead` implementation. This implicitly adds the trait bound for `SchemaRead` for that type. Moreover, it consumes the `TYPE_META` property from the `SchemaRead` implementation - which means the `SchemaWrite::TYPE_META` property is ignored, see PoC.

[Code](https://github.com/anza-xyz/wincode/blob/ce424cbcd1c3cca34ad16b8d462feb444fcbf172/wincode-derive/src/assert_zero_copy.rs#L55-L82)
```rust
const _assert_schema_read_impl: fn() = || {
    // implicitly assumes we have a `SchemaRead` implementation
    fn assert_schema_read_impl<'de, T: #crate_name::SchemaRead<'de, #config_path>>() {}
    assert_schema_read_impl::<#ident>()
};
// ...
// Assert the struct has no padding bytes.
const _assert_no_padding: () = {
    // we only look at SchemaRead implementation
    if let #crate_name::TypeMeta::Static { size, .. } = <#ident as #crate_name::SchemaRead<'_, #config_path>>::TYPE_META {
        if size != core::mem::size_of::<#ident>() {
            panic!("wincode(assert_zero_copy) was applied to a type with padding");
        }
    } else {
        panic!("wincode(assert_zero_copy) was applied to a type with `TypeMeta::Dynamic`");
    }
};
```

### Fix

Generalize `assert_zero_copy` to assume the correct trait depending on where it is called from.

### PoC

```rust
/// A field type with deliberately asymmetric metadata: `SchemaRead::TYPE_META` claims a
/// padding-free `Static` zero-copy layout, while `SchemaWrite::TYPE_META` is `Dynamic`.
struct AsymMeta(u8);

// SAFETY: asserted only to exercise the derive's zero-copy assertions; `AsymMeta` is a
// single byte with no invalid bit patterns.
unsafe impl<C: ConfigCore> config::ZeroCopy<C> for AsymMeta {}

unsafe impl<C: Config> SchemaWrite<C> for AsymMeta {
    type Src = Self;

    // The write side does NOT advertise a static zero-copy layout.
    const TYPE_META: TypeMeta = TypeMeta::Dynamic;

    fn size_of(_src: &Self::Src) -> WriteResult<usize> {
        Ok(1)
    }
    fn write(writer: impl Writer, src: &Self::Src) -> WriteResult<()> {
        <u8 as SchemaWrite<C>>::write(writer, &src.0)
    }
}

unsafe impl<'de, C: Config> SchemaRead<'de, C> for AsymMeta {
    type Dst = Self;

    // The read side advertises a padding-free static zero-copy layout.
    const TYPE_META: TypeMeta = TypeMeta::Static {
        size: 1,
        zero_copy: true,
    };

        dst.write(AsymMeta(<u8 as SchemaRead<'de, C>>::get(reader)?));
        Ok(())
    }
}

/// Derives `SchemaWrite` with `#[wincode(assert_zero_copy)]`, yet its
/// `SchemaWrite::TYPE_META` is `Dynamic` (inherited from the `AsymMeta` field).
///
/// It compiles anyway because `assert_zero_copy`'s padding check
/// (`wincode-derive/src/assert_zero_copy.rs:74-82`) inspects
/// `<Self as SchemaRead>::TYPE_META` (which is `Static`), not the `SchemaWrite`
/// metadata produced by the derive the assertion is attached to.
#[repr(transparent)]
#[wincode(internal, assert_zero_copy)]
struct WrapperWriteDynamic(AsymMeta);

/// Showcases that the `SchemaWrite` derive's `assert_zero_copy` validates the wrong
/// trait's metadata. `WrapperWriteDynamic` passed the zero-copy assertion at compile
/// time (otherwise this test would not build), yet its `SchemaWrite::TYPE_META` — the
/// metadata governing the write path the assertion is attached to — is `Dynamic`.
#[test]
fn test_assert_zero_copy_ignores_schema_write_type_meta() {
    // The write metadata the assertion should have validated is `Dynamic`...
    assert!(matches!(
        <WrapperWriteDynamic as SchemaWrite<DefaultConfig>>::TYPE_META,
        TypeMeta::Dynamic
    ));
    // ...but the assertion only looked at the read metadata, which is `Static`.
    assert!(matches!(
        <WrapperWriteDynamic as SchemaRead<DefaultConfig>>::TYPE_META,
        TypeMeta::Static { .. }
    ));
}
```

koriyoshi2041:
## Summary

Fixes #383.

`#[wincode(assert_zero_copy)]` was shared by the `SchemaRead` and `SchemaWrite` derives, but always asserted the `SchemaRead` implementation and inspected `SchemaRead::TYPE_META`. That let a `SchemaWrite` derive pass the zero-copy assertion when the write-side metadata was `Dynamic` but the read-side metadata was `Static`.

This change makes the assertion helper receive the trait currently being derived, then checks the matching trait bound and `TYPE_META`.

## Tests

- `cargo fmt --check`
- `cargo test -p wincode --doc --features derive`
- `cargo test -p wincode-derive`
- `cargo test -p wincode --features derive`

### DIFF
--- a/wincode-derive/src/common.rs
+++ b/wincode-derive/src/common.rs
@@ -13,8 +13,9 @@ use {
         rc::Rc,
     },
     syn::{
-        DeriveInput, Expr, GenericArgument, Generics, Ident, Lifetime, LitInt, Member, Path, Type,
-        TypeImplTrait, TypeParamBound, TypeReference, TypeTraitObject, Visibility, parse_quote,
+        DeriveInput, Expr, GenericArgument, Generics, Ident, Lifetime, Lit, LitInt, Member, Path,
+        Type, TypeImplTrait, TypeParamBound, TypeReference, TypeTraitObject, Visibility,
+        parse_quote,
         spanned::Spanned,
         visit::{self, Visit},
         visit_mut::{self, VisitMut},
@@ -587,6 +588,7 @@ pub(crate) struct SchemaArgs {
     /// Supports both flag-style and explicit path specification:
     /// - `#[wincode(assert_zero_copy)]` - uses default config
     /// - `#[wincode(assert_zero_copy(MyConfig))]` - uses custom config path
+    /// - `#[wincode(assert_zero_copy(schema = "read"))]` - checks only read metadata
     #[darling(default)]
     pub(crate) assert_zero_copy: Option<AssertZeroCopyConfig>,
     /// Specifies the path to the `wincode` crate.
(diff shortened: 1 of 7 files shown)


kskalski:
We should probably support asserting separately for one (current behaviors) selected (currently impossible) or both (better default) schemas to be zero-copy.

Do you think this could be easily added to this PR?

koriyoshi2041:
Yes. I think the clean behavior is:

- bare `assert_zero_copy` checks both read and write metadata (new default)
- an explicit selector allows read-only or write-only assertions
- the existing custom-config form remains supported

The implementation is small; the only part I do not want to guess is the public attribute spelling because the current positional argument is already a config path. Would you prefer named options such as `assert_zero_copy(schema = "read", config = MyConfig)` (with `schema = "both"` as the default)? I can add the parser, both asymmetric compile-fail controls, and compatibility coverage in this PR once that spelling is settled.

kskalski:
Yeah, I think that would work well:
* `assert_zero_copy!()` translates to `assert_zero_copy!(config = DefaultConfig, schema = "both")`
* `assert_zero_copy!(SomeConfig)` translates to `assert_zero_copy!(config = SomeConfig, schema = "both")`
* schema can be `read`, `write` or `both` (default)

kskalski (commented):
This looks good. I wonder though if we will have to make the default as `read` for now until nest 0.X version bump to preserve strict compatibility. @cpubot what do you think?

koriyoshi2041:
Updated the default to `schema = "read"` in 3c1324a to preserve 0.x compatibility, while keeping explicit `read`, `write`, and `both` selectors available. I also added coverage showing the bare form accepts read-static/write-dynamic metadata and made the existing write-metadata compile-fail case explicitly request `schema = "both"`.\n\nVerified with derive parser tests, the asymmetric metadata integration test, compile-fail doctests, formatting, and `git diff --check`.

kskalski (dismissed):
One possibility would be to emit deprecation warning until next major bump that not specifying any schema type will change its default behavior (and about this bug being a possibility right now), but I think it's not worth it. I suppose this is really a niche case, so we can just make it a hard switch on next breaking release.

wincode-derive/src/common.rs
```diff
@@ -549,25 +551,87 @@ impl SchemaArgs {
...
+                                value.value()
+                            )));
+                        }
+                    };
+                }
+                _ => return Err(darling::Error::unexpected_type("path or named option")),
+            }
         }
+
+        Ok(AssertZeroCopyConfig { config, schema })
```
cpubot: This can be simplified substantially:

impl `FromMeta` for `ZeroCopySchema`:
```rs
impl FromMeta for ZeroCopySchema {
    fn from_string(value: &str) -> Result<Self> {
        match value {
            "read" => Ok(Self::Read),
            "write" => Ok(Self::Write),
            "both" => Ok(Self::Both),
            _ => Err(darling::Error::custom(format!(
                "unknown schema `{value}`; expected `read`, `write`, or `both`"
            ))),
        }
    }
}
```

Then handle duplicates with a single helper
```rs
fn from_list(items: &[NestedMeta]) -> darling::Result<Self> {
    let mut config = None;
    let mut schema = None;

    fn set_once_with<T>(
        target: &mut Option<T>,
        field: &'static str,
        span: &impl Spanned,
        value: impl FnOnce() -> Result<T>,
    ) -> Result<()> {
        if target.is_some() {
            return Err(darling::Error::duplicate_field(field).with_span(span));
        }
        *target = Some(value()?);
        Ok(())
    }

    for item in items {
        match item {
            NestedMeta::Meta(syn::Meta::Path(path)) => {
                set_once_with(&mut config, "config", path, || Ok(path.clone()))?;
            }
            NestedMeta::Meta(syn::Meta::NameValue(value)) if value.path.is_ident("config") => {
                set_once_with(&mut config, "config", value, || {
                    Path::from_expr(&value.value)
                })?;
            }
            NestedMeta::Meta(syn::Meta::NameValue(value)) if value.path.is_ident("schema") => {
                set_once_with(&mut schema, "schema", value, || {
                    ZeroCopySchema::from_expr(&value.value)
                })?;
            }
            _ => {
                return Err(
                    darling::Error::unexpected_type("path or named option").with_span(item)
                );
            }
        }
    }

    Ok(AssertZeroCopyConfig {
        config,
        schema: schema.unwrap_or(ZeroCopySchema::Read),
    })
}
```

wincode/src/lib.rs
```diff
@@ -240,7 +240,7 @@
 //! |Attribute|Type|Default|Description
 //! |---|---|---|---|
 //! |`tag_encoding`|`Type`|`None`|Specifies the encoding/decoding schema to use for the variant discriminant. Only usable on enums.|
-//! |`assert_zero_copy`|`bool`\|`Path`|`false`|Generates compile-time asserts to ensure the type meets zero-copy requirements. Can specify a custom config path, will use the [`DefaultConfig`](config::DefaultConfig) if `bool` form is used.|
+//! |`assert_zero_copy`|`bool`\|`Path`\|options|`false`|Generates compile-time asserts to ensure the type meets zero-copy requirements. A custom config path may be supplied positionally or as `config = MyConfig`; `schema = "read"`, `"write"`, or `"both"` selects which metadata is checked (`"read"` by default for compatibility).|
```
cpubot: `by default for compatibility` -- should drop `for compatibility`.

wincode/src/schema/compile_fail.rs
```diff
@@ -148,3 +148,57 @@ fn uninit_builder_read_forbids_lifetime_launder() {}
...
+///
+///     fn read(reader: impl Reader<'de>, dst: &mut MaybeUninit<Self::Dst>) -> ReadResult<()> {
+///         dst.write(AsymMeta(<u8 as SchemaRead<'de, C>>::get(reader)?));
+///         Ok(())
+///     }
+/// }
+///
+/// #[repr(transparent)]
+/// #[derive(SchemaWrite, SchemaRead)]
+/// #[wincode(internal, assert_zero_copy(schema = "both"))]
```
cpubot: This test currently compile-fails for the wrong reason. Rustdoc builds doctests as separate crates, so `#[wincode(internal)]` makes generated `crate::...` paths resolve against the doctest crate

cpubot (changes requested):

koriyoshi2041:
Addressed the three review points in dddc6c2 and rebased onto current master. The parser now uses FromMeta plus a single duplicate-field helper, the compatibility wording is gone, and the compile-fail case resolves through the public crate path so it reaches the intended metadata assertion. Verified derive tests, the asymmetric integration test, all doctests/compile-fail cases, formatting, and diff checks.

koriyoshi2041:
The requested parser, compatibility, and public-path test changes are all in dddc6c2, and the branch remains mergeable. Could you confirm whether any review item is still outstanding? I can update it promptly if so.

PR #404, opened 2026-07-21
feat(derive): Support `context`

cpubot:
This adds derive macro support for the wincode context system. Syntax looks like the following:
```rs
#[wincode(context = "&'bump Bump")]
struct Foo<'bump> {
    id: u32,
    #[wincode(context)]
    bar: String<'bump>,
    #[wincode(context)]
    baz: Vec<'bump, u64>,
}
```

By specifying a top-level `#[wincode(context = "")]`, the `SchemaRead` macro will implement `SchemaReadContext` for the specified context type.

Fields that specify `#[wincode(context)]` will be treated as `SchemaReadContext` impls with the same context type as the above. Fields without that attribute will use their typical `SchemaRead` impl.


wincode-derive/src/common.rs
```diff
@@ -186,6 +233,64 @@ impl Field {
     pub(crate) fn has_lifetime(&self) -> bool {
         self.ty.has_lifetime()
     }
+
+    /// Generate the trait bound for this field, including the exact destination
+    /// or source type expected by the containing type.
+    ///
+    /// Context lifetimes remain tied to the context; other read lifetimes are
+    /// rewritten to `'de` because they may borrow from the input.
+    pub(crate) fn as_constraint(&self, tr: TraitImpl) -> TokenStream {
```
cpubot, line 242: The following methods `as_constraint`, `qualifier`, and `target_fully_qualified` are conveniences for avoiding manual generation of fairly verbose quoted expressions. It works for both `SchemaRead` and `SchemaWrite`, and eliminates a lot of existing boilerplate, as well as abstracting over `SchemaReadContext` vs `SchemaRead` with appropriate lifetime rewrites and constraints.

wincode-derive/src/common.rs
```diff
@@ -224,21 +329,12 @@ impl FieldsExt for Fields<Field> {
...
-            TraitImpl::SchemaWrite => {
-                let items = self.unskipped_iter().map(|field| {
-                    let target = field.target_resolved();
-                    quote! { <#target as #crate_name::SchemaWrite<__WincodeConfig>>::TYPE_META }
-                });
-                quote! { #(#items),* }
-            }
+        let tuple_expansion = {
+            let items = self.unskipped_iter().map(|field| {
+                let target = field.target_fully_qualified(trait_impl);
```
cpubot, line 334: A nice effect of above convenience methods: no manual branching and quoting of fully qualified types with appropriate lifetime parameters, replacements, and target resolution.

wincode-derive/src/common.rs
```diff
@@ -473,7 +560,7 @@ pub(crate) fn get_crate_name(args: &SchemaArgs) -> Path {
 }
 
 #[derive(FromDeriveInput)]
-#[darling(attributes(wincode), forward_attrs)]
+#[darling(attributes(wincode), forward_attrs, and_then = Self::validate)]
```
cpubot, line 563: This avoids having to call `validate` in every derive macro expansion site, and additionally allows us to enrich fields with cached data, rather than having to thread through all the various parameters required to generate fully qualified quoted expressions

wincode-derive/src/schema_read.rs
```diff
@@ -351,12 +437,13 @@ fn append_where_clause(generics: &mut Generics, data: &Data<Variant, Field>, cra
...
     ) {
-        let ty = &field.ty.with_lifetime("de");
-        let target = field.target_resolved().with_lifetime("de");
+        let constraint = field.as_constraint(TraitImpl::SchemaRead);
+        let target = field
+            .target_resolved()
+            .with_lifetime_excluding("de", &field.context_lifetimes);
         let mut bounds = Punctuated::new();
-        bounds.push(parse_quote!(#crate_name::SchemaRead<'de, __WincodeConfig, Dst = #ty>));
+        bounds.push(parse_quote!(#constraint));
```
kskalski, line 480: since using >=2 context fields treats context as `Copy`, we could add an explicit contraint to the impl if we can count >=2 uses of `context` among the fields such that this condition is more explicit and yields better error messages (otherwise user will just see attempt of using move value somewhere in generated code)

(or we could just generate a const condition assert in the code directly)

wincode/src/schema/external/bumpalo.rs
```diff
@@ -310,4 +310,109 @@ mod tests {
 
         assert_eq!(bytes_ref, b"borrowed");
     }
+
+    #[test]
+    fn context_derive_struct_can_outlive_input_bytes() {
+        #[derive(SchemaRead, SchemaWrite, Debug, PartialEq)]
+        #[wincode(internal, context = "&'bump Bump")]
+        struct Foo<'bump> {
```
kskalski, line 318: could we add a test where Foo has generic parameter passed on to one of the fields and where Foo has other lifetime than utilized by context ?

wincode-derive/src/schema_read.rs
```diff
@@ -68,12 +79,28 @@ fn impl_struct(
...
-                        #crate_name::io::Reader::by_ref(&mut reader),
-                        unsafe { &mut *(&raw mut (*dst_ptr).#ident).cast::<#hint>() }
-                    )?;
-                    #init_count
+                match &field.context {
+                    Some(_) => {
+                        quote! {
+                            #assert_safe_lifetime_shortening
+                            #fully_qualified::read_with_context(
+                                ctx,
```
robertkiel, line 87: `ctx` may not be `Copy`. Structs with >1 fields with non-`Copy` context fail compilation.

```rust
use crate::{context, SchemaRead};

// Enum variant with two context fields, non-Copy context.
#[derive(SchemaRead)]
#[wincode(internal, context = "context::Len")]
enum TwoCtxEnum {
    V {
        #[wincode(context)]
        a: alloc::vec::Vec<u8>,
        #[wincode(context)]
        b: alloc::vec::Vec<u8>,
    },
}
```

Compilation error
```
error[E0382]: use of moved value
   move occurs because value has type `Len`, which does not implement the `Copy` trait
 = note: this error originates in the derive macro `SchemaRead`
```

robertkiel: related https://github.com/anza-xyz/wincode/pull/404#discussion_r3628163091

wincode-derive/src/schema_read.rs
cpubot: Yeah, users will already get a compile error like you mentioned. But probably nicer to get an explicit `type doesn't satisfy Copy` error

wincode/src/schema/external/bumpalo.rs
cpubot: Added https://github.com/anza-xyz/wincode/pull/404/changes/c9f5b5b5f75b0be01aa6d34a72ec83e7faaa6b79

wincode-derive/src/schema_read.rs
cpubot: Added https://github.com/anza-xyz/wincode/pull/404/changes/2713f71dc752304bac3a8ec438fce83e1c8de80c

wincode-derive/src/schema_read.rs
cpubot: Good catch. Will fix this in a follow up PR since the issue predates it

kskalski (approved):

PR #427, opened 2026-08-11
feat(io): Support deserialization limit

cpubot:
This adds a new `Reader` implementation, `LimitReader`, that will enforce a given limit of bytes read from its inner `Reader`.

A corresponding `DESERIALIZATION_SIZE_LIMIT` paramter has been added to `ConfigCore`. If set, all entrypoints in `config` (e.g., `config::deserialize`, `config::deserialize_with_context`, etc) will wrap their given reader in `LimitReader` with the configured limit.

Main caveat to this implementation is that this limit will not be enforced for users calling `SchemaRead::read` or `SchemaRead::get` directly using a compile-time-only `Config` with a deserialization size limit specified -- that is only possible with a more substantial rewrite to the `Reader` system. It is not viable for every `SchemaRead` / `SchemaReadContext` implementation to conditionally construct a `LimitReader` over its given `Reader`. I think this is a reasonable trade-off to provide this functionality in a minimally invasive way.

### DIFF
--- a/wincode/src/config/serde.rs
+++ b/wincode/src/config/serde.rs
@@ -81,15 +81,15 @@ macro_rules! maybe_size_limit {
     };
 
     (@reader slice, $src:ident, $limit:ident) => {{
-        let mut src = $src;
+        let src = $src;
         let len = $limit.min(src.len());
-        src.take_borrowed(len).unwrap()
+        &src[..len]
     }};
 
     (@reader slice_mut, $src:ident, $limit:ident) => {{
-        let mut src = $src;
+        let src = $src;
         let len = $limit.min(src.len());
-        src.take_borrowed_mut(len).unwrap()
+        &mut src[..len]
     }};
 }
 
--- a/wincode/src/config/serde.rs
+++ b/wincode/src/config/serde.rs
@@ -61,12 +61,12 @@ pub trait Serialize<C: Config>: SchemaWrite<C> {
 impl<T, C: Config> Serialize<C> for T where T: SchemaWrite<C> + ?Sized {}
 
 macro_rules! maybe_size_limit {
-    ($config:ty, $src:expr, |$reader:ident| $body:expr $(,)?) => {{
-        let mut src = $src;
+    ($config:ty, $src:expr, $kind:ident, |$reader:ident| $body:expr $(,)?) => {{
+        let src = $src;
 
         match <$config as $crate::config::ConfigCore>::DESERIALIZATION_SIZE_LIMIT {
             Some(limit) => {
-                let $reader = src.as_limited_for(limit);
+                let $reader = maybe_size_limit!(@reader $kind, src, limit);
                 $body
             }
             None => {
@@ -74,6 +74,22 @@ macro_rules! maybe_size_limit {
                 $body
             }
         }
+    }};
+
+    (@reader generic, $src:ident, $limit:ident) => {
+        $crate::io::LimitReader::new($src, $limit)
+    };
+
+    (@reader slice, $src:ident, $limit:ident) => {{
+        let mut src = $src;
+        let len = $limit.min(src.len());
+        src.take_borrowed(len).unwrap()
+    }};
+
+    (@reader slice_mut, $src:ident, $limit:ident) => {{
+        let mut src = $src;
+        let len = $limit.min(src.len());
+        src.take_borrowed_mut(len).unwrap()
     }};
 }
 
@@ -83,7 +99,7 @@ pub trait Deserialize<'de, C: Config>: SchemaRead<'de, C> {
     #[inline(always)]
     #[expect(unused_variables)]
     fn deserialize(src: &'de [u8], config: C) -> ReadResult<Self::Dst> {
-        maybe_size_limit!(C, src, |reader| Self::get(reader))
+        maybe_size_limit!(C, src, slice, |reader| Self::get(reader))
     }
 
     /// Deserialize the input bytes into `dst`.
@@ -94,7 +110,7 @@ pub trait Deserialize<'de, C: Config>: SchemaRead<'de, C> {
         dst: &mut MaybeUninit<Self::Dst>,
         config: C,
     ) -> ReadResult<()> {
-        maybe_size_limit!(C, src, |reader| Self::read(reader, dst))
+        maybe_size_limit!(C, src, slice, |reader| Self::read(reader, dst))
     }
 }
 
@@ -107,7 +123,7 @@ pub trait DeserializeOwned<C: Config>: SchemaReadOwned<C> {
     fn deserialize_from<'de>(
         src: impl Reader<'de>,
     ) -> ReadResult<<Self as SchemaRead<'de, C>>::Dst> {
-        maybe_size_limit!(C, src, |reader| Self::get(reader))
+        maybe_size_limit!(C, src, generic, |reader| Self::get(reader))
     }
 
     /// Deserialize from the given [`Reader`] into `dst`.
@@ -116,7 +132,7 @@ pub trait DeserializeOwned<C: Config>: SchemaReadOwned<C> {
         src: impl Reader<'de>,
         dst: &mut MaybeUninit<<Self as SchemaRead<'de, C>>::Dst>,
     ) -> ReadResult<()> {
-        maybe_size_limit!(C, src, |reader| Self::read(reader, dst))
+        maybe_size_limit!(C, src, generic, |reader| Self::read(reader, dst))
     }
 }
 
@@ -210,7 +226,21 @@ pub fn deserialize_exact<'de, T, C: Config>(mut src: &'de [u8], config: C) -> Re
 where
     T: SchemaRead<'de, C, Dst = T>,
 {
-    let value = maybe_size_limit!(C, src.by_ref(), |reader| T::get(reader))?;
+    let value = match C::DESERIALIZATION_SIZE_LIMIT {
+        Some(limit) => {
+            let mut limited = &src[..limit.min(src.len())];
+            let initial_len = limited.len();
+            let value = T::get(limited.by_ref())?;
+
+            #[expect(clippy::arithmetic_side_effects)]
+            let consumed = initial_len - limited.len();
+            src = &src[consumed..];
+
+            value
+        }
+        None => T::get(src.by_ref())?,
+    };
+
     if src.is_empty() {
         Ok(value)
     } else {
@@ -229,7 +259,7 @@ pub fn deserialize_with_context<'de, Ctx, T, C: Config>(
 where
     T: SchemaReadContext<'de, C, Ctx, Dst = T>,
 {
-    maybe_size_limit!(C, src, |reader| T::get_with_context(ctx, reader))
+    maybe_size_limit!(C, src, slice, |reader| T::get_with_context(ctx, reader))
 }
 
 /// Like [`crate::deserialize_mut`], but allows the caller to provide a custom configuration.
@@ -239,7 +269,7 @@ pub fn deserialize_mut<'de, T, C: Config>(src: &'de mut [u8], config: C) -> Read
 where
     T: SchemaRead<'de, C, Dst = T>,
 {
-    maybe_size_limit!(C, src, |reader| T::get(reader))
+    maybe_size_limit!(C, src, slice_mut, |reader| T::get(reader))
 }
 
 /// Like [`crate::deserialize_from`], but allows the caller to provide a custom configuration.
@@ -271,9 +301,12 @@ pub unsafe trait ZeroCopy<C: ConfigCore>: 'static {
     where
         Self: SchemaRead<'de, C, Dst = Self> + Sized,
     {
-        maybe_size_limit!(C, bytes, |reader| <&Self as SchemaRead<'de, C>>::get(
-            reader
-        ))
+        maybe_size_limit!(
+            C,
+            bytes,
+            slice,
+            |reader| <&Self as SchemaRead<'de, C>>::get(reader)
+        )
     }
 
     /// Like [`crate::ZeroCopy::from_bytes_mut`], but allows the caller to provide a custom configuration.
@@ -283,9 +316,10 @@ pub unsafe trait ZeroCopy<C: ConfigCore>: 'static {
     where
         Self: SchemaRead<'de, C, Dst = Self> + Sized,
     {
-        maybe_size_limit!(C, bytes, |reader| <&mut Self as SchemaRead<'de, C>>::get(
-            reader
-        ))
+        maybe_size_limit!(C, bytes, slice_mut, |reader| <&mut Self as SchemaRead<
+            'de,
+            C,
+        >>::get(reader))
     }
 }
 
@@ -303,7 +337,7 @@ mod tests {
 
         assert!(matches!(
             deserialize::<u64, _>(&bytes, limited),
-            Err(ReadError::Io(IoReadError::ReadSizeLimit(4)))
+            Err(ReadError::Io(IoReadError::ReadSizeLimit(8)))
         ));
         assert!(matches!(
             deserialize_from::<u64, _>(bytes.as_slice(), limited),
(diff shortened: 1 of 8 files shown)


wincode/src/io/limit_reader.rs
```diff
@@ -0,0 +1,207 @@
...
+///
+/// # Examples
+///
+/// ```
+/// # use wincode::io::LimitReader;
+/// # use wincode::io::ReadError;
+/// #
+/// let mut bytes = [0u8; 8];
+/// wincode::serialize_into(&mut bytes[..], &42u64);
+/// let reader = LimitReader::new(&bytes[..], 4);
```
kskalski, line 35: Do you think there could be a way to short-circuit combinations of readers (e.g. `LimitReader<&[u8]>` can actually be mapped into a subslice `&[u8]`)?
If this feature gets adoption, we could consider adding `as_limited_to(self, usize) -> impl Reader` to `Reader`, which would open up optimizing reservations. Until then I guess this wrapper is good enough.

kskalski (dismissed):

wincode/src/io/limit_reader.rs
cpubot: Tested this a bit locally -- definitely viable, and I think the approach is nice improvement. The slice impls can simply call `self.take_borrowed(limit.min(self.len())` to create a shorter slice that will naturally impose the given limit.

cpubot: Follow-up -- one wrinkle with this approach is that the reservation / advancement semantics differ between slice implementations and the `LimitReader`.

- `LimitReader` lazily performs accounting and advances
- Slice impls perform an eager window reservation

Both enforce the same maximum read boundary, but they do not leave the parent reader at the same position afterward. Unsure if this will matter in practice (probably not too much), but worth deciding if we're okay with a method like `as_limited_for` having two different reservation / advancement semantics. I guess it's not that different from `as_trusted_for` specifying that it may not perform an advancement if the underlying implementation cannot support it.

cpubot: Updated -- I think this approach is preferable

wincode/src/io/mod.rs
```diff
@@ -253,6 +256,21 @@ pub unsafe trait Reader<'a> {
...
+    /// Operations through the returned reader that would exceed this limit must
+    /// return an error. Because this method is safe, callers have no obligation to
+    /// independently remain within the limit.
+    ///
+    /// Implementations may reserve up to `size` bytes from the parent immediately,
+    /// even if fewer bytes are subsequently read, or may enforce the limit
+    /// incrementally. Callers must not rely on the parent reader's resulting
+    /// position.
+    #[inline]
+    fn as_limited_for(&mut self, size: usize) -> ReadResult<impl Reader<'a>> {
```
cpubot: This could just return `impl Reader<'a>`, since constructing a `LimitReader` is infallible, and the slice / cursor overrides are logically infallible since they clamp to their internal buffer length. Can't think of any other IO cases where returning an error is useful. Thoughts @kskalski?

kskalski: We will want to keep the contract weak, i.e. it's a limit, not a reservation / guarantee that consumer needs *exactly* `size` bytes, as it's going to be used for validation.

So yes, seems like infallible is better. The only case I can think of is some kind of wrapper type that needs to do fallible operation to update internal state - say Mutex, or RefCell.

wincode/src/io/limit_reader.rs
kskalski: I'm thinking about the use case like this:
* deserialize `(A, B)`
* want to bound bytes consumed specifically by `A`
* use reader limited with that bound to deserialize `A`
* continue deserializing `B` with the same (original) reader, without limit

The slice impl currently forbids that. I wonder though how to best handle that:
* make the API specify both min/max bytes to be consumed?
* let user manually use `LimitReader` for `A` - if we want to avoid confusing like this, I suppose `as_limited_for` should be consuming `self`

cpubot: Good find. You're right, current implementation advances the underlying slice such that the following does not work
```rs
struct Tuple<A, B> {
    a: A,
    b: B,
}

unsafe impl<'de, C: ConfigCore, A, B> SchemaRead<'de, C> for Tuple<A, B>
where
    A: SchemaRead<'de, C>,
    B: SchemaRead<'de, C>,
{
    type Dst = (A::Dst, B::Dst);

    #[inline]
    fn read(mut reader: impl Reader<'de>, dst: &mut MaybeUninit<Self::Dst>) -> ReadResult<()> {
        let a = {
            let reader = reader.as_limited_for(4);
            A::get(reader)
        }?;
        let b = B::get(reader)?;
        dst.write((a, b));
        Ok(())
    }
}

#[test]
fn test_tuple() {
    let tuple = (42u32, 42u64);
    let serialized = crate::serialize(&tuple).unwrap();
    let deserialized = <Tuple<u32, u64>>::deserialize(&serialized).unwrap();
    assert_eq!(deserialized, tuple)
}
```

My sense is that the right policy is just to remove the override for the slice / cursor impls rather than making `as_limited_for` consume `self` and indicate in the docs that the limited reader should remain synchronized with the original reader while not consuming more bytes than are actually read. If we made `as_limited_for` consume `self`, we still have the same problem, e.g.,
```rs
struct Foo {
    tuple: Tuple<u32, u64>,
    // `tuple` calls `as_limited_for` and the original input slice is truncated
    bar: u8
}
```

wincode/src/io/mod.rs
cpubot: Done

wincode/src/io/limit_reader.rs
cpubot: Updated

kskalski: If we can't really optimize `as_limited_for` for the typical / easiest slice case, maybe it's isn't such a good API after all - just using `LimitReader::new` seems to be better than current state of the PR.

The child / parent synchronization could be solved though - a slice reader that advances parent on drop does work correctly (`&mut` will not work though):
```rust
pub struct LimitedSlice<'p, 'a> {
    parent: &'p mut &'a [u8],
    window: &'a [u8],
    /// Window length at construction, used to derive the consumed byte count on drop.
    len: usize,
}
```

so we could consider that. 

Parent ownership issue could be handled a bit better if we apply synchronization on the call-site, e.g. inside `maybe_size_limit!`, since it can better track the exact mut borrowing scopes for child read and parent update.

wincode/src/config/serde.rs
```diff
@@ -60,13 +60,30 @@ pub trait Serialize<C: Config>: SchemaWrite<C> {
...
+    }};
+}
+
 /// Like [`crate::Deserialize`], but allows the caller to provide a custom configuration.
 pub trait Deserialize<'de, C: Config>: SchemaRead<'de, C> {
     /// Deserialize the input bytes into a new `Self::Dst`.
     #[inline(always)]
     #[expect(unused_variables)]
     fn deserialize(src: &'de [u8], config: C) -> ReadResult<Self::Dst> {
-        Self::get(src)
```
kskalski, line 69: the APIs that receive `&[u8]` could definitely be optimized to use sub-slice instead of limited reader / `as_limited_for`

I wish to actually clean some of them to use `impl Reader` before 1.0, but maybe we could then use the `supports_borrow` trick to check if we are dealing with slice and branch off cheaply based on that.

wincode/src/io/limit_reader.rs
```diff
@@ -0,0 +1,203 @@
...
+            inner,
+            remaining: limit,
+        }
+    }
+
+    #[inline]
+    const fn reserve_limit(&mut self, needed: usize) -> ReadResult<()> {
+        if needed > self.remaining {
+            #[expect(clippy::arithmetic_side_effects)]
+            return Err(read_size_limit(needed - self.remaining));
```
kskalski: given 
```rust
    #[error("Attempting to read {0} bytes")]
    ReadSizeLimit(usize),
```
this should just pass `needed`. We could consider using a new error for this, though this gets a bit tricky given how callers rely on exact error variant in https://github.com/anza-xyz/wincode/blob/4ccbb1c7bdf2cb24f3672a96d9285ab9337c1c99/wincode/src/schema/adapter.rs#L198

wincode/src/io/mod.rs
kskalski, line 363: `as_limited_for` should be handled here

kskalski (commented):
Feels like we should reconsider peek_borrowed kind of APIs for reader

wincode/src/io/limit_reader.rs
cpubot: > maybe it's isn't such a good API after all - just using LimitReader::new seems to be better than current state of the PR.

Reasonable assessment. IMO we can just ditch `as_limited_for` for now and revisit later -- coming up with the right design for a trait method can be a separate issue discussion / PR. We can just ship `LimitedReader` and come back to this design

wincode/src/config/serde.rs
cpubot: Done -- handled by the macro explicitly

wincode/src/io/limit_reader.rs
cpubot: Seems unlikely that users are relying on an exact `ReadSizeLimit(n)`, but you're right this is a slight behavioral change in `ReadSizeLimit` accounting. A new error variant is reasonable, but is technically a breaking change. I'll update to pass `needed` so we have symmetry

cpubot: Updated

wincode/src/config/serde.rs
```diff
@@ -60,13 +60,46 @@ pub trait Serialize<C: Config>: SchemaWrite<C> {
...
+    (@reader slice, $src:ident, $limit:ident) => {{
+        let mut src = $src;
+        let len = $limit.min(src.len());
+        src.take_borrowed(len).unwrap()
+    }};
+
+    (@reader slice_mut, $src:ident, $limit:ident) => {{
+        let mut src = $src;
+        let len = $limit.min(src.len());
+        src.take_borrowed_mut(len).unwrap()
```
kskalski: take borrowed_mut here (similarly above) can be actually changed into
```rust
        let len = $limit.min($src.len());
        &mut $src[..len]
```
which seems cleaner and probably more correct w.r.t. lifetimes

wincode/src/schema/mod.rs
```diff
@@ -5192,4 +5192,16 @@ mod tests {
         *bytes.last_mut().unwrap() = 2; // corrupt the second bool
         assert!(deserialize::<Vec<bool>>(&bytes).is_err());
     }
+
+    #[test]
+    fn ref_cell_scratch() {
```
kskalski: doesn't seem to belong to this PR

wincode/src/config/serde.rs
```diff
@@ -264,6 +316,73 @@ pub unsafe trait ZeroCopy<C: ConfigCore>: 'static {
...
+            <u8 as ZeroCopy<_>>::from_bytes(&bytes, limited),
+            Err(ReadError::Io(IoReadError::ReadSizeLimit(1)))
+        ));
+
+        let mut bytes = bytes;
+        assert!(matches!(
+            <u8 as ZeroCopy<_>>::from_bytes_mut(&mut bytes, limited),
+            Err(ReadError::Io(IoReadError::ReadSizeLimit(1)))
+        ));
     }
```
kskalski, line 550: Proposed extra tests:
* Truncation charges bytes actually read; `LimitReader` charges bytes reserved. The two entrypoints must still accept and reject the same inputs at the same limits.
* Enforcement is per-call-site, so every entrypoint needs its own coverage: an entrypoint that forgets `maybe_size_limit!` silently applies no limit.
*  A limit larger than the input is the ordinary configuration (limit = max message size, buffers usually smaller), and the only case that exercises the clamp against input length.

```rust
    #[test]
    fn slice_and_generic_paths_agree_at_every_limit() {
        macro_rules! check {
            ($ty:ty, $val:expr, $($limit:literal),+) => {$({
                let bytes = crate::serialize(&$val).unwrap();
                let cfg = Configuration::default().with_deserialization_size_limit::<$limit>();
                let via_slice = deserialize::<$ty, _>(&bytes, cfg);
                let via_reader = deserialize_from::<$ty, _>(bytes.as_slice(), cfg);
                assert_eq!(
                    via_slice.is_ok(), via_reader.is_ok(),
                    "{} at limit {}: slice={:?} reader={:?}",
                    stringify!($ty), $limit, via_slice.err(), via_reader.err(),
                );
                if let (Err(a), Err(b)) = (via_slice, via_reader) {
                    assert_eq!(
                        format!("{a:?}"), format!("{b:?}"),
                        "{} at limit {}: differing errors", stringify!($ty), $limit,
                    );
                }
            })+};
        }

        // Vec<u32>: 8-byte len prefix + a trusted window of 4 * len.
        check!(Vec<u32>, vec![1u32, 2, 3], 0, 1, 7, 8, 9, 12, 16, 19, 20, 21, 100);
        // Vec<String>: nested length prefixes, no single trusted window.
        check!(Vec<String>, vec!["ab".to_string(), "cde".to_string()], 0, 8, 16, 18, 26, 29, 30, 31);
        // Fixed-size types.
        check!(u64, 42u64, 0, 4, 7, 8, 9);
        check!(core::ops::Range<u32>, 0u32..5, 0, 4, 7, 8, 9);
    }

    #[test]
    fn every_entrypoint_enforces_the_limit() {
        let bytes = 42u64.to_le_bytes();
        let short = Configuration::default().with_deserialization_size_limit::<4>();
        let exact = Configuration::default().with_deserialization_size_limit::<8>();

        let mut dst = MaybeUninit::<u64>::uninit();
        assert!(matches!(
            <u64 as Deserialize<_>>::deserialize_into(&bytes, &mut dst, short),
            Err(ReadError::Io(IoReadError::ReadSizeLimit(8)))
        ));
        assert!(<u64 as Deserialize<_>>::deserialize_into(&bytes, &mut dst, exact).is_ok());

        // `deserialize_from_into` takes no config argument, so bind `C` through a call.
        fn from_into<C: Config>(src: &[u8], _config: C) -> ReadResult<u64> {
            let mut dst = MaybeUninit::uninit();
            <u64 as DeserializeOwned<C>>::deserialize_from_into(src, &mut dst)?;
            // SAFETY: `deserialize_from_into` returned `Ok`, initializing `dst`.
            Ok(unsafe { dst.assume_init() })
        }
        assert!(matches!(
            from_into(bytes.as_slice(), short),
            Err(ReadError::Io(IoReadError::ReadSizeLimit(8)))
        ));
        assert_eq!(from_into(bytes.as_slice(), exact).unwrap(), 42);

        let mut mutable = bytes;
        assert!(matches!(
            deserialize_mut::<u64, _>(&mut mutable, short),
            Err(ReadError::Io(IoReadError::ReadSizeLimit(8)))
        ));
        assert_eq!(deserialize_mut::<u64, _>(&mut mutable, exact).unwrap(), 42);

        // `Vec<u32>` with a caller-supplied length reads 4 * len bytes and no length prefix.
        let elements = crate::serialize(&[1u32, 2, 3]).unwrap();
        assert_eq!(elements.len(), 12);
        assert!(matches!(
            deserialize_with_context::<_, Vec<u32>, _>(
                crate::schema::context::Len(3),
                &elements,
                short,
            ),
            Err(ReadError::Io(IoReadError::ReadSizeLimit(_)))
        ));
        let twelve = Configuration::default().with_deserialization_size_limit::<12>();
        assert_eq!(
            deserialize_with_context::<_, Vec<u32>, _>(
                crate::schema::context::Len(3),
                &elements,
                twelve,
            )
            .unwrap(),
            vec![1, 2, 3],
        );
    }

    #[test]
    fn limit_larger_than_input_is_clamped() {
        let bytes = 42u64.to_le_bytes();
        let generous = Configuration::default().with_deserialization_size_limit::<1024>();

        assert_eq!(deserialize::<u64, _>(&bytes, generous).unwrap(), 42);
        assert_eq!(deserialize_exact::<u64, _>(&bytes, generous).unwrap(), 42);
        assert_eq!(
            deserialize_from::<u64, _>(bytes.as_slice(), generous).unwrap(),
            42
        );

        let mut mutable = bytes;
        assert_eq!(deserialize_mut::<u64, _>(&mut mutable, generous).unwrap(), 42);

        let mut one = [7u8];
        assert_eq!(*<u8 as ZeroCopy<_>>::from_bytes(&one, generous).unwrap(), 7);
        assert_eq!(
            *<u8 as ZeroCopy<_>>::from_bytes_mut(&mut one, generous).unwrap(),
            7
        );

        // Trailing bytes are still detected when the limit exceeds the input.
        let mut trailing = [0u8; 9];
        trailing[..8].copy_from_slice(&bytes);
        assert!(matches!(
            deserialize_exact::<u64, _>(&trailing, generous),
            Err(ReadError::TrailingBytes)
        ));
    }
```

wincode/src/schema/mod.rs
cpubot: Whoops

wincode/src/config/serde.rs
cpubot: done

wincode/src/config/serde.rs
cpubot: Done

kskalski (approved):

PR #435, opened 2026-08-15
feat(collections): support strict map/set decoding

### LINKED ISSUE #429: Support strict map/set decoding
Deserializing a map or set type (e.g., `HashMap`, `HashSet`) ignores duplicate keys during deserialization -- i.e., multiple `insert`s on the same key simply overwrite the previous value. For certain use-cases, it may be preferable to return an error if the sequence of values being deserialized into a map or set contains duplicates, which I'll refer to as _strict decoding_. I do not believe this strict decoding behavior is the right default, as all sequences in wincode use the same encoding format, and it is perfectly reasonable for someone to deserialize a set of values serialized as a `Vec` into a `HashSet` if one wanted to, for example, explicitly de-dupe that sequence.

To support this, we can add a `Config` boolean constant (something like `STRICT_MAP_SET`) to allow users to opt into this behavior.

### LINKED ISSUE #454: Keyed collection schemas in `containers` with a duplicate key policy
Split from #429, which covers strict map/set decoding as a global config option. This tracks the per-field half.

`containers` has `Vec`, `Box`, `VecDeque` etc. for customizing the length encoding of a sequence, but nothing for the keyed collections, so there is no way to control either the length encoding or duplicate key handling for a single `HashMap`/`HashSet` field.

Add `containers::{HashMap, BTreeMap, HashSet, BTreeSet, IndexMap, IndexSet}` taking a length encoding and a duplicate key policy:

```rust
struct Foo {
    #[wincode(with = "containers::HashMap<u32, u64, BincodeLen, RandomState, CheckUniqueKeys>")]
    map: HashMap<u32, u64>,
}
```

The policy should default to today's behavior (a repeated key is left to the collection's own `insert`), with `CheckUniqueKeys` failing the read instead.

surajk-m:
**Why**: deserializing a map or set silently collapses repeated keys. All sequences share one wire format, so nothing distinguishes a `Vec` of unique values from a `HashSet`, which makes permissive decoding the right default -- but leaves no way to reject input where collapsing two entries into one matters.

**What**: add keyed collection schemas under containers (`HashMap`, `BTreeMap`, `HashSet`, `BTreeSet`, `IndexMap`, `IndexSet`) that take a length encoding and a `DuplicateKeyPolicy` for per field control.

```rust
// globally
let config = DefaultConfig::new().enable_strict_map_set();

// or per field
#[wincode(with = "containers::HashMap<u32, u64, BincodeLen, CheckUniqueKeys>")]
map: HashMap<u32, u64>,
```

Closes #454. Part of #429, the per-field `with` adapters. The global config route stays open.


wincode/src/schema/containers.rs
```diff
@@ -50,8 +50,40 @@
...
+//! ```
+//!
+//! To flip the default for every keyed collection at once, see
+//! [`enable_strict_map_set`](crate::config::Configuration::enable_strict_map_set).
 #[cfg(all(feature = "alloc", target_has_atomic = "ptr"))]
 use alloc::sync::Arc as AllocArc;
+#[cfg(feature = "std")]
+use core::hash::{BuildHasher, Hash};
+#[cfg(all(feature = "indexmap", feature = "std"))]
+use indexmap::{IndexMap as ExtIndexMap, IndexSet as ExtIndexSet};
```
kskalski: we shoudl keep all indexmap impl inside `external/indexmap.rs`

wincode/src/config/mod.rs
```diff
@@ -27,13 +27,15 @@ pub const PREALLOCATION_SIZE_LIMIT_DISABLED: usize = usize::MAX;
...
 /// - Tag encoding is [`u32`].
+/// - Strict map/set decoding is disabled.
 pub struct Configuration<
     const ZERO_COPY_ALIGN_CHECK: bool = true,
     const PREALLOCATION_SIZE_LIMIT: usize = DEFAULT_PREALLOCATION_SIZE_LIMIT,
     LengthEncoding = BincodeLen,
     ByteOrder = LittleEndian,
     IntEncoding = FixInt,
     TagEncoding = u32,
+    const STRICT_MAP_SET: bool = false,
```
kskalski: I would vote for not adding a config knob for strict set decoding, it can be served equally well using the adapters and `with=` on specific fields. @cpubot what's your take on this?

wincode/src/schema/mod.rs
```diff
@@ -519,6 +519,114 @@ where
...
+    fn miscounted_iter() -> crate::WriteError {
+        crate::WriteError::Custom("ExactSizeIterator did not yield exactly its reported len()")
+    }
+
+    let len = src.len();
+    macro_rules! write_entries {
+        ($w:expr) => {{
+            Len::write($w.by_ref(), len)?;
+            // Unlike `write_elem_iter`, let the iterator drive the loop -- pulling `len`
+            // times through `next()` instead costs ~40% on hash table iterators. Hold it
```
kskalski: can you elaborate on this? is this something measured by benchmark or profiler? I think "iterator drive the loop" is essentially calling next() and checking for `Some`, the same that 0..len loop would do

if this is some real win, maybe it should also be applied to `write_elem_iter`, but we would need a way to measure / reproduce this

wincode/src/schema/impls.rs
kskalski: we could get rid of several macros by completely moving the implementations for direct types into `map_container!` and `set_container!` in form of simple delegation to the schema types

we need to handle linkedlist by more direct impl without using the macros I suppose

wincode/src/config/mod.rs
cpubot: I think there's value in setting this globally for certain applications. But, I think the two implementations should be separated -- i.e., one PR for the config knob and one for `with` adapters.

wincode/src/schema/containers.rs
surajk-m: Done!

wincode/src/config/mod.rs
surajk-m: Removed. I could open a separate PR for the global knob. Thanks @kskalski @cpubot

wincode/src/schema/mod.rs
surajk-m: Yeah, measured with the existing criterion benches. `HashMap<u64, u64>/wincode/serialize_into/1000` is ~725 ns with the `0..len` + `next()` shape vs ~514 ns iterator driven. You are right that both call `next()`, the difference is codegen, the extra induction variable + Option branch seems to hurt hashbrown's iterator specifically (slice iterator don't care). Haven't dug into the asm.

To reproduce this, we could swap the loop body in `write_kv_iterator` and then run `cargo bench "HashMap<u64, u64>/wincode/serialize"`.

Agree on `write_elem_iter`, probably the same win, but there's no `HashSet` bench to verify with. I could do a follow-up with a bench added.

wincode/src/schema/impls.rs
surajk-m: Done!

wincode/src/error.rs
```diff
@@ -55,6 +55,8 @@ pub enum ReadError {
     InvalidCharLead(u8),
     #[error("Trailing bytes remain after deserialization")]
     TrailingBytes,
+    #[error("Duplicate key encountered while decoding a keyed collection of {0}")]
+    DuplicateKey(&'static str),
```
cpubot: This is technically a breaking change. We could avoid this with `Custom` for now, and fold in `DuplicateKey` with the next breaking release. Thoughts @kskalski?

wincode/src/schema/containers.rs
```diff
@@ -515,6 +547,457 @@ where
...
+
+    macro_rules! read_elems {
+        ($reader:expr) => {{
+            // Reserve capacity, capped for unique keys; iteration still uses `len`.
+            let mut set = make(capacity(len));
+            for _ in 0..len {
+                let present = insert(&mut set, T::get($reader.by_ref())?);
+                if Dup::REJECT_DUPLICATES && present {
+                    return Err(duplicate_key());
+                }
```
kskalski: we could add 
```rust
    /// Fail the read if the entry just decoded collided with an earlier one.
    #[inline(always)]
    fn check(collided: bool) -> ReadResult<()> {
        if Self::REJECT_DUPLICATES && collided {
            return Err(crate::error::duplicate_key());
        }
        Ok(())
    }
```
to the trait to simplify the calls

wincode/src/schema/containers.rs
```diff
@@ -515,6 +547,457 @@ where
     }
 }
 
+/// How a keyed collection schema reacts when the encoded sequence repeats a key.
+///
+/// Default is [`AllowDuplicateKeys`]. See [`HashMap`] for an example.
+pub trait DuplicateKeyPolicy {
```
kskalski: we could actually seal the trait, probably doesn't make sense to have other than two possible modes

wincode/src/schema/external/indexmap.rs
```diff
@@ -1,20 +1,34 @@
...
+    crate::containers::{map_container, set_container},
     core::hash::{BuildHasher, Hash},
-    indexmap::{IndexMap, IndexSet},
+    indexmap::{IndexMap as ExtIndexMap, IndexSet as ExtIndexSet},
 };
 
-impl_seq_kv! { "indexmap", IndexMap<K: Hash | Eq, V, S: BuildHasher | Default>, IndexMap::with_capacity_and_hasher, cap_unique_keys }
-impl_seq_v! { "indexmap", IndexSet<K: Hash | Eq, S: BuildHasher | Default>, IndexSet::with_capacity_and_hasher, insert, cap_unique_keys }
+map_container! {
+    #[cfg(all())]
```
kskalski: cfg(all) doesn't do anything, just remove it

wincode/src/schema/containers.rs
```diff
@@ -515,6 +547,457 @@ where
...
+}
+
+#[cfg(feature = "alloc")]
+pub(crate) use seq_capacity;
+
+/// Read a length-prefixed sequence of key/value pairs into a map-like collection,
+/// where `insert` reports whether an entry for that key already existed.
+#[cfg(feature = "alloc")]
+#[inline]
+pub(crate) fn read_kv_seq<'de, K, V, Len, Dup, C, M>(
```
kskalski: we could replace this impl by just using `let map = $crate::containers::read_elem_seq::<($key, $value), Len, Dup, C, _>(` though maybe worth benchamarking, since deserializing (K,V) vs manually first K, then V might yield sub-optimal code, since wincode currently cannot optimize for tuple's memory layout

wincode/src/schema/containers.rs
surajk-m: Done!

wincode/src/schema/containers.rs
surajk-m: Sealed!

wincode/src/schema/mod.rs
surajk-m: @kskalski Did it here instead of a follow-up `write_elem_iter ` is now iterator driven too, with a `HashSet<u64>` bench added to verify: serialize_into/100 goes ~47.6 ns -> 39.2 ns (-17%), /1000 ~(-3%). Please review this.

wincode/src/schema/containers.rs
```diff
@@ -50,8 +50,34 @@
...
+//! assert_eq!(Lenient::deserialize(&bytes).unwrap(), BTreeSet::from([1, 2]));
+//! assert!(matches!(Strict::deserialize(&bytes), Err(ReadError::Custom(_))));
+//! # }
+//! ```
 #[cfg(all(feature = "alloc", target_has_atomic = "ptr"))]
 use alloc::sync::Arc as AllocArc;
+#[cfg(feature = "std")]
+use core::hash::{BuildHasher, Hash};
+#[cfg(feature = "std")]
+use std::collections::{
```
kskalski: could we collapse imports gated by the same `cfg` into single block?

wincode/src/schema/mod.rs
```diff
@@ -519,6 +522,113 @@ where
...
+    src: impl ExactSizeIterator<Item = (&'a K::Src, &'a V::Src)>,
+) -> WriteResult<()>
+where
+    C: ConfigCore,
+    Len: SeqLen<C>,
+    K: SchemaWrite<C, Src: Sized + 'a>,
+    V: SchemaWrite<C, Src: Sized + 'a>,
+{
+    #[cold]
+    fn miscounted_iter() -> crate::WriteError {
```
kskalski: we could move it to top level, since it's used in two functions

wincode/src/schema/containers.rs
kskalski: were you able to compare with this function replaced by `read_elem_seq`?

wincode/src/schema/mod.rs
```diff
@@ -519,6 +522,113 @@ where
...
+        crate::WriteError::Custom("ExactSizeIterator did not yield exactly its reported len()")
+    }
+
+    let len = src.len();
+    macro_rules! write_entries {
+        ($w:expr) => {{
+            Len::write($w.by_ref(), len)?;
+            // Driving this with `for _ in 0..len` + `next()` is ~40% slower on
```
kskalski: I run the benchmarks on a beefy devbox and the impact of this change is not that large - a few % for `HashMap<u64, u64>`, negative up to 2% gain for `HashMap<[u8; 16], PodStruct>/wincode/serialize_into/`. There is however a non-negligible impact on Vec serialize_into:
```
Vec<UnitEnum>/wincode/serialize_into/100
                        time:   [22.641 ns 22.644 ns 22.649 ns]
                        change: [−20.754% −20.628% −20.501%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 3 outliers among 100 measurements (3.00%)
  1 (1.00%) low mild
  1 (1.00%) high mild
  1 (1.00%) high severe

Vec<UnitEnum>/wincode/serialize_into/1000
                        time:   [230.70 ns 230.72 ns 230.75 ns]
                        thrpt:  [1.6467 GiB/s 1.6469 GiB/s 1.6471 GiB/s]
                 change:
                        time:   [+5.6545% +5.6983% +5.7422%] (p = 0.00 < 0.05)
                        thrpt:  [−5.4304% −5.3911% −5.3518%]
                        Performance has regressed.
Found 2 outliers among 100 measurements (2.00%)

Vec<SameSizedEnum>/wincode/serialize_into/100
                        time:   [42.336 ns 42.342 ns 42.348 ns]
                        change: [+36.538% +37.474% +38.378%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 5 outliers among 100 measurements (5.00%)
  4 (4.00%) high mild
  1 (1.00%) high severe
Vec<SameSizedEnum>/wincode/serialize_into/1000
                        time:   [420.08 ns 420.10 ns 420.12 ns]
                        thrpt:  [4.4514 GiB/s 4.4516 GiB/s 4.4518 GiB/s]
                 change:
                        time:   [+29.404% +31.206% +32.884%] (p = 0.00 < 0.05)
                        thrpt:  [−24.746% −23.784% −22.723%]
                        Performance has regressed.
Found 3 outliers among 100 measurements (3.00%)
  1 (1.00%) high mild
  2 (2.00%) high severe
Vec<SameSizedEnum>/wincode/serialize_into/10000
                        time:   [4.1379 µs 4.1380 µs 4.1382 µs]
                        thrpt:  [4.5029 GiB/s 4.5031 GiB/s 4.5032 GiB/s]
                 change:
                        time:   [+27.645% +29.513% +31.299%] (p = 0.00 < 0.05)
                        thrpt:  [−23.838% −22.788% −21.658%]
                        Performance has regressed.
Found 8 outliers among 100 measurements (8.00%)
  6 (6.00%) high mild
  2 (2.00%) high severe

Vec<MixedSizedEnum>/wincode/serialize_into/100
                        time:   [70.002 ns 70.307 ns 70.633 ns]
                        change: [−7.9648% −7.3772% −6.6605%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 10 outliers among 100 measurements (10.00%)
  6 (6.00%) high mild
  4 (4.00%) high severe

```

which we should understand better and mitigrate negative impact.

I propose to revert this aspect of the change - keep using 0..len, so we can make sure all patterns using this impl are not regressing.

Then we can do a perf change as follow up.

wincode/src/schema/containers.rs
```diff
@@ -515,6 +547,466 @@ where
...
+    macro_rules! read_elems {
+        ($reader:expr) => {{
+            // Reserve capacity, capped for unique keys; iteration still uses `len`.
+            let mut set = make(capacity(len));
+            for _ in 0..len {
+                Dup::check(insert(&mut set, T::get($reader.by_ref())?))?;
+            }
+            set
+        }};
+    }
```
kskalski: can be simplified to
```suggestion
pub(crate) fn read_elem_seq<'de, T, Len, C, S>(
    mut reader: impl Reader<'de>,
    make: impl FnOnce(usize) -> S,
    mut insert: impl FnMut(&mut S, T::Dst) -> ReadResult<()>,
) -> ReadResult<S>
where
    C: ConfigCore,
    Len: SeqLen<C>,
    T: SchemaRead<'de, C>,
{
    let len = Len::read_prealloc_check::<T::Dst>(reader.by_ref())?;

    macro_rules! read_elems {
        ($reader:expr) => {{
            let mut set = make(len);
            for _ in 0..len {
                insert(&mut set, T::get($reader.by_ref())?)?;
            }
            set
        }};
    }
```

which should also make the callsites simpler
```rust
|map, k, v| Dup::check(map.insert(k, v).is_some())   // map_container
|set, k| Dup::check(!set.insert(k))                  // set_container
```
and linked list:
```rust
    let list = containers::read_elem_seq::<T, C::LengthEncoding, C, _>(
        reader,
        |_| LinkedList::new(),
        |list, value| {
            list.push_back(value);
            Ok(())
        },
    )?;
    dst.write(list);
    Ok(())
}
```

wincode/src/schema/containers.rs
surajk-m: yes, ran it. Tuple version is ~9% slower on `HashMap<[u8;16], PodStruct>`/wincode/deserialize, wash on `HashMap<u64,u64>` and `BTreeMap`. Only shows up when `K` and `V` are static size, the tuple read opens a nested trusted window per entry plus the drop guard. 

Kept `read_kv_seq` with a note.

wincode/src/schema/mod.rs
surajk-m: Done!

wincode/src/schema/mod.rs
surajk-m: Agree, reverted!

wincode/src/schema/containers.rs
surajk-m: did the same to `read_kv_seq`. Thanks!

wincode/src/schema/containers.rs
```diff
@@ -515,6 +548,474 @@ where
...
+
+#[cfg(feature = "alloc")]
+pub(crate) use seq_capacity;
+
+/// Read a length-prefixed sequence of key/value pairs into a map-like collection.
+///
+/// `make` receives the decoded length and builds the collection; `insert` places one
+/// entry and may fail the read (see [`DuplicateKeyPolicy::check`]).
+///
+/// Reads the key and value separately rather than as a `(K, V)` tuple: the tuple read
```
kskalski: I would trim down this comment to just mention that manual K, V deserialization is faster due to wincode's inability to optimize for tuple memory layout.
And move it to the impl code near K, V rather than here in function doc-comment

wincode/src/schema/containers.rs
```diff
@@ -515,6 +548,474 @@ where
...
+            type Src = $target<$key::Src, $value::Src $(, $state)?>;
+
+            #[inline]
+            fn size_of(src: &Self::Src) -> $crate::WriteResult<usize> {
+                $crate::schema::size_of_kv_iter::<$key, $value, Len, C>(src.iter())
+            }
+
+            #[inline]
+            fn write(writer: impl $crate::io::Writer, src: &Self::Src) -> $crate::WriteResult<()> {
+                $crate::schema::write_kv_iter_prealloc_check::<$key, $value, Len, C>(writer, src.iter())
```
kskalski: This is actually the only caller of `write_kv*` functions and we are forcing src to be iterated through `.iter()`, which we then push through a function that by operating on `ExactSizedIterator` is obliged to check if `len()` conforms to the number of iterated items...

We should just inline the impl here and operate directly on the collection that we know conforms to the `len()` contract, so we can skip the extra checks. In my tests this gets rid of the regressions.

kskalski (commented):
I found that our benchmarks are quite noisy, some of the regressions I've seen before might not be real, but I did manage to reproduce the one you mentioned from the beginning. And I think it's worth getting rid of - fortunately with current shape of this PR this should be easy - see comment on the code.

You can also try running with a more stable benchmark with https://github.com/anza-xyz/wincode/pull/440

wincode/src/schema/containers.rs
surajk-m: Thanks @kskalski, moved it next to the `K/V` reads, kept just the tuple layout bit.

wincode/src/schema/containers.rs
surajk-m: Done, inlined both impls into the macro and dropped the `write_kv*` helpers.

wincode/src/schema/containers.rs
```diff
@@ -515,6 +548,513 @@ where
...
+    }
+}
+
+mod sealed {
+    pub trait Sealed {}
+    impl Sealed for super::AllowDuplicateKeys {}
+    impl Sealed for super::CheckUniqueKeys {}
+}
+
+/// A repeated key overwrites the entry decoded for it earlier (last one wins).
```
kskalski: technically this is impl dependent, e.g. hashset preserves the previous value, so worth mentioning here

wincode/src/schema/external/indexmap.rs
```diff
@@ -1,20 +1,32 @@
...
+    crate::containers::{map_container, set_container},
     core::hash::{BuildHasher, Hash},
-    indexmap::{IndexMap, IndexSet},
+    indexmap::{IndexMap as ExtIndexMap, IndexSet as ExtIndexSet},
 };
 
-impl_seq_kv! { "indexmap", IndexMap<K: Hash | Eq, V, S: BuildHasher | Default>, IndexMap::with_capacity_and_hasher, cap_unique_keys }
-impl_seq_v! { "indexmap", IndexSet<K: Hash | Eq, S: BuildHasher | Default>, IndexSet::with_capacity_and_hasher, insert, cap_unique_keys }
+map_container! {
+    /// Like [`HashMap`](crate::containers::HashMap), for [`IndexMap`](indexmap::IndexMap).
```
kskalski: technically this references std-only type, so doc comment builds will be polluted by warnings in no-std context, might be best to gate the comment

kskalski (commented):
I think this looks good now, just a few nits, but I would like @cpubot to also take a look

wincode/src/schema/external/indexmap.rs
surajk-m: gated!

wincode/src/schema/containers.rs
```diff
@@ -50,6 +50,26 @@
 //! assert_eq!(my_struct, wincode::deserialize(&bytes).unwrap());
 //! # }
 //! ```
+//!
+//! # Keyed collections
+//!
+//! Map and set containers also take a [`DuplicateKeyPolicy`]. By default a repeated
+//! key overwrites the previous entry, like the plain `HashMap`/`BTreeMap` schemas.
```
kskalski: here we also need an update to distingush map from set, or to keep it short, we could use
```
Default is [`AllowDuplicateKeys`]; [`CheckUniqueKeys`] rejects a repeat instead.
```

wincode/src/schema/containers.rs
```diff
@@ -515,6 +548,523 @@ where
...
+        $(#[cfg($cfg:meta)])?
+        $(#[doc = $doc:expr])*
+        $(#[cfg_attr($ca:meta, doc = $cadoc:expr)])*
+        $name:ident => $target:ident<$key:ident : $($constraint:path)|*
+            $(, $state:ident : $($state_constraint:path)|* = $state_default:ty)?>,
+        $with_capacity:expr
+        $(, $cap_unique_keys:ident)?
+    ) => {
+        $crate::containers::map_container! {
+            @struct $(#[cfg($cfg)])? $(#[doc = $doc])* $(#[cfg_attr($ca, doc = $cadoc)])* $name<$key> $(, $state = $state_default)?
```
kskalski: I think handling cfg_attr for the docs here is adding too much complexity, let's just phrase the comments in such a way that they refer to the trait or available types at the same `cfg` visibility. Could mention Hashmap for canonical example without linking to it too.

wincode/src/schema/containers.rs
```diff
@@ -515,6 +548,523 @@ where
...
+        pub struct $name<
+            $($generic,)*
+            Len,
+            Dup = $crate::containers::AllowDuplicateKeys,
+            $state = $state_default,
+        >(core::marker::PhantomData<($($generic,)* Len, Dup, $state)>);
+        $(#[doc = $doc])*
+        $(#[cfg_attr($ca, doc = $cadoc)])*
+        #[cfg(all($($cfg,)? not(feature = "std")))]
+        pub struct $name<$($generic,)* Len, Dup, $state>(
```
kskalski: let's keep Dup as last type param, so it matches other structs, it's more likely going to be used with default than customized state - I think it only affects IndexMap, which doesn't have `std` feature

wincode/src/schema/containers.rs
surajk-m: Thanks!

wincode/src/schema/containers.rs
surajk-m: Agree!

wincode/src/schema/containers.rs
surajk-m: tried it, but it doesn't work. The impls bind positionally as <Len, Dup, S>, so a <Len, S, Dup> struct in `no-std` ends up with no `SchemaRead` at all. Would need to cfg-split the impl headers too, same complexity as the cfg_attr thing.

kept <Len, Dup, S> in both.

kskalski: I think it should be solvable:
```
  unsafe impl<'de, C: ConfigCore, $key, $value, Len, Dup $(, $state)?>   // binder — order irrelevant
      SchemaRead<'de, C> for $name<$key, $value, Len, Dup $(, $state)?>  // self type — positional
```
  The binder is just a declaration of names; leaving it as Len, Dup, $state is fine. The for $name<…> list is what binds to the struct's parameters, so it needs the same swap the struct got:
```
  -            $crate::SchemaRead<'de, C> for $name<$key, $value, Len, Dup $(, $state)?>
  +            $crate::SchemaRead<'de, C> for $name<$key, $value, Len $(, $state)?, Dup>
```
  Four of those (map/set × SchemaWrite/SchemaRead), plus the six delegation bodies that spell the type out:
 ```
  -                <$name<$key, $value, C::LengthEncoding, AllowDuplicateKeys $(, $state)?> as …>
  +                <$name<$key, $value, C::LengthEncoding $(, $state)?, AllowDuplicateKeys> as …>
 ```
  Crucially the self type is the same in both configs — <K, V, Len, S, Dup> under std and without it. Only the defaults differ, and defaults live on the struct definition, not the impl. So the impls stay cfg-free; the existing two-arm `struct` split absorbs the whole difference. Nothing like the cfg_attr plumbing.

surajk-m: Got it. Thanks @kskalski, earlier i only swapped the struct and left the impl bindings in the old order, that's why it had no `SchemaRead`. Did it properly now.

kskalski (dismissed):

wincode/src/schema/containers.rs
```diff
@@ -515,6 +547,517 @@ where
     }
 }
 
+/// How a keyed collection schema reacts when the encoded sequence repeats a key.
+///
+/// Default is [`AllowDuplicateKeys`]. See `HashMap` for an example.
+pub trait DuplicateKeyPolicy: sealed::Sealed {
+    /// Whether a repeated key aborts the read with
+    /// [`ReadError::Custom`](crate::error::ReadError::Custom).
+    const REJECT_DUPLICATES: bool;
```
cpubot: This `const REJECT_DUPLICATES` isn't really providing much. It's used exclusively for a default implementation of `check` and isn't referenced elsewhere. Why not just implement `check` for both implementations?

wincode/src/schema/containers.rs
```diff
@@ -515,6 +547,517 @@ where
...
+#[cfg(feature = "alloc")]
+pub(crate) use seq_capacity;
+
+/// Read a length-prefixed sequence of key/value pairs into a map-like collection.
+///
+/// `make` receives the decoded length and builds the collection; `insert` places one
+/// entry and may fail the read (see [`DuplicateKeyPolicy::check`]).
+#[cfg(feature = "alloc")]
+#[inline]
+pub(crate) fn read_kv_seq<'de, K, V, Len, C, M>(
```
cpubot, line 615: We can have this accept a `dst: &mut MaybeUninit<M>`. This will avoid returning an intermediate `Result<M>`, which creates extra stack copies (even after inlining) relative to the generated assembly on master, and will plug nicely into the `read` implementations.

wincode/src/schema/containers.rs
```diff
@@ -515,6 +547,517 @@ where
...
+        read_entries!(reader)
+    };
+
+    Ok(map)
+}
+
+/// Variant of [`read_kv_seq`] for collections of standalone elements rather than pairs.
+#[cfg(feature = "alloc")]
+#[inline]
+pub(crate) fn read_elem_seq<'de, T, Len, C, S>(
```
cpubot, line 669: Same comment as the above with respect to accepting a `dst: &mut MaybeUninit<M>`.

wincode/src/schema/containers.rs
surajk-m: done, both helpers take `dst` now and check is implemented per policy.

wincode/src/schema/containers.rs
```diff
@@ -515,6 +547,518 @@ where
     }
 }
 
+/// How a keyed collection schema reacts when the encoded sequence repeats a key.
+///
+/// Default is [`AllowDuplicateKeys`]. See `HashMap` for an example.
+pub trait DuplicateKeyPolicy: sealed::Sealed {
+    /// Fail the read if the entry just decoded collided with an earlier one.
```
cpubot: It's up to the implementation to decide to error on collision. Perhaps something like `Apply the policy to a detected key collision`

surajk-m: Updated!

cpubot (approved):

surajk-m:
Thanks, @cpubot! updated and opened #454.

kskalski:
I edited the description a bit, since it was still referring to STRICT_MAP option, which is not added here.

kskalski (approved):

PR #449, opened 2026-09-01
perf(derive): reserve a trusted window for a struct's static prefix

kskalski:
#### Problem
- a struct is Dynamic as soon as one field is, and the derive then reserves no trusted window at all
- so every static field pays its own bounds check, even where a run of them has a fixed total size and could share one
- header-then-payload, payload-then-trailer and anything alternating all pay for it: those fields go one at a time, cmp/je per field, where the same bytes in an all-static struct merge into a few moves

#### Summary of Changes
- plan every declaration as opening a window, sitting in one, or going direct, and cover each window's run with a single bounds check
  - add `TypeMeta::field_plan` and `FieldPlan` to compute that plan at compile time
  - leave a lone static field direct: a window checks the parent, builds the child and advances the parent again, so a single field will not benefit
  - enter skipped fields as zero-sized statics, so they stay inside the window they were declared in, which keeps initialization in declaration order for the drop guard with no separate bookkeeping
- drop the Static/Dynamic split: an all-static struct is one full-length window and a struct whose static fields never adjoin uses all-direct plan, so the two old arms are the extreme variants of general plan mechanism
- emit one chain function per field, tail calling the next while its window continues, so a field's read/write is written once rather than once per run it could fall in; inlining still pulls a field into every chain ahead of its own, but the const guards cut each chain at its window boundary, leaving one linear stripe per window instead of the whole tail
  - guard the dispatch and the hand-off with `if const`, so the arms a config does not take are dropped before anything is specialized for them
  - hold the plan in an associated const on `WriteFieldPlan` / `ReadFieldPlan`: it depends on the config and the struct's generics (a `const` item couldn't name them and an inline `const {}` would recompute per guard)
  - declare those two traits in the library rather than per expansion, and emit
      the impl beside the schema impl, where a non-local one would be linted
- normalize the schema impls' generics, moving each parameter's inline bounds into the where clause, so every chain function declares them the one way and no bound is split across both places  
- add the wire_layout bench: five arms of 32 wire bytes, differing only in how their static fields group into runs

#### Testing
Numbers, ns per call, two runs each, one benchmark per process at 64-byte alignment:
```
                             master          after
  header/serialize_into      0.866  0.865    0.873  0.865
  header/deserialize         0.876  0.863    0.869  0.865
  prefixed/serialize_into    1.529  1.538    1.188  1.204
  prefixed/deserialize       2.099  2.084    1.333  1.317
  suffixed/serialize_into    1.544  1.563    1.204  1.222
  suffixed/deserialize       1.761  1.637    1.195  1.200
  interleaved/serialize_into 1.998  1.970    1.755  1.753
  interleaved/deserialize    2.221  2.200    1.897  1.895
  alternating/serialize_into 2.456  2.461    2.466  2.438
  alternating/deserialize    2.994  2.980    2.999  2.990
```
header is all static and unmoved, so collapsing the old arms costs the common case nothing. The middle three hold one window per side of a dynamic field and gain in proportion to how much of the struct those windows cover. alternating matches master because no two of its static fields adjoin, so it wins no window either way.


kskalski:
Yeah, this works and also has a benefit of simplifying how we handle `skip` fields (they fall into the same framework as 0 sized run / covered). 

One disadvantage is the explosion of generated-code with noticeable compile-time regression, e.g. compiling struct with 24 fields adds over 1s compilation time. It's quadratic over number of fields, so the impact on small structs is reasonable though.

kskalski:
Ok, I got it into a nice shape and compilation times seem to got under control. This version introduces a few tricks, I updated the PR description to summarize them.

